PR #8, opened 2023-06-30
plotst:0.1.0

Pegacraft:
I am submitting
- [x] a new package
- [ ] an update for a package

**Description:**
Plotst is a package people can use to create a few types of different plots and graphs. It can be useful to display data in a straight forward way.

It supports:
- Scatter plots
- Graph charts
- Histograms
- Bar charts
- Pie charts
- Overlaying plots/charts


packages/preview/plotst-0.1.0/README.md
```diff
@@ -0,0 +1,225 @@
...
+If you need some example code, check out `main.typ` in the `examples` folder. It also includes a compiled version of the current `main.typ`
+
+## Examples:
+
+All these images were created using the `main.typ`
+
+### Scatter plots
+
+```js
+    // Plot 1:
```
laurmaedje, line 36: There seems to be some superfluous indent here. Would be great if the examples were indented consistently with 2 or 4 spaces. Ideally 2 as that is the "official" style.

packages/preview/plotst-0.1.0/example/main.typ
```diff
@@ -0,0 +1,212 @@
+#import "/lib.typ": *
```
laurmaedje, line 1: I'm a bit on the fence with this one. To make this example copy-pasteable it should use the `@preview` import. That makes it difficult to test locally though, of course.

laurmaedje:
There is one more thing I am unsure about: This package contains a lot of auxiliary files (examples, docs, etc.) When somebody imports this packages, these will all be downloaded as a tar.gz and be made available to the compiler. The archive for this package would be 451KB even though the actual source files are probably a few KB. There is the question whether the images and/or docs files should only be part of the external repository and linked to.

Pegacraft:
> There is one more thing I am unsure about: This package contains a lot of auxiliary files (examples, docs, etc.) When somebody imports this packages, these will all be downloaded as a tar.gz and be made available to the compiler. The archive for this package would be 451KB even though the actual source files are probably a few KB. There is the question whether the images and/or docs files should only be part of the external repository and linked to.

You are absolutely right. I can remove all the auxiliary stuff and just link to my repo. That would leave the plain code at 9 files, if I counted correctly ^^

packages/preview/plotst-0.1.0/example/main.typ
Pegacraft: I could put both in and comment one with "for local testing" or something. On my repo that is. I removed the file here

packages/preview/plotst-0.1.0/README.md
Pegacraft: Changed the readme to have consistent indentation. Now it should be fine

packages/preview/plotst-0.1.0/example/main.typ
laurmaedje: Up to you then!

laurmaedje:
Thank you!

PR #16, opened 2023-07-03
fontawesome:0.1.0

duskmoon314:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package provides functions to use icons of fontawesome conveniently by leveraging the desktop fonts. One can simply add icons via their name once the font is installed.


packages/preview/fontawesome-0.1.0/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### Customization
+
+The `fa-icon` function passes args to `text`, so you can customize the icon by passing parameters to it:
+
+`#fa-icon("chess-queen", fill: blue)`
+
+## Example
+
+See the `example.typ` file for a complete example. (TODO: add a gallery for all icons)
```
istudyatuni, line 61: Maybe put a link to the list of icons here? https://fontawesome.com/search?o=r&m=free

packages/preview/fontawesome-0.1.0/lib.typ
```diff
@@ -0,0 +1,2491 @@
...
+
+typst-fontawesome
+
+https://github.com/duskmoon314/typst-fontawesome
+
+generated by typst-fontawesome helper python script
+
+*/
+
+#let FA_VERSION = state("fa-version", "Font Awesome 6")
```
laurmaedje, line 11: why does this use a state? wrapping everything in locate below can incur significant overhead.

duskmoon314: My goal is to provide a function to set the whole icon set. Since we cannot use `set rule` on non-element functions, I find out that `state` can achieve the goal. Thus, one can use `FA_SET.update("Brands")` to change the default set to `Brands` and use `fa-icon("github")`, which cannot be directly used in the `Free` set.

If there is a better way to achieve this, I would like to change this ASAP.

laurmaedje: I see. In that case it should be documented how to use that in the README.

packages/preview/fontawesome-0.1.0/README.md
duskmoon314: Good advice.

packages/preview/fontawesome-0.1.0/lib.typ
duskmoon314: Oh, right. I put it in the example of the original repo and forget the README.

laurmaedje:
Thank you!

PR #14, opened 2023-07-02
ansi-render:0.1.0

8LWXpg:
I am submitting
- [x] a new package
- [ ] an update for a package

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name in conformance with the guidelines
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] ensured that my submission does not infringe upon the rights of a third party
- [x] tested my package locally on my system and it worked
- [x] named this PR as `name:version` of the submitted package
- [x] agree that my package will not be removed without good reason

Description: provides a simple way to render text with ANSI escape sequences in typst

PR #18, opened 2023-07-04
fix typo in ansi-render README

8LWXpg:
[referenced PR](https://github.com/typst/packages/pull/14#discussion_r1251432191)


laurmaedje:
In accordance with the updated submission guidelines minor documentation fixes that cannot break existing users of the package are now allowed.

PR #24, opened 2023-07-06
cetz:0.0.1

johannes-wolf:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: CeTZ (formerly `typst-canvas`) is a package providing a canvas environment for drawing with Typst. Its API is heavily inspired by TikZ and Processing.

URL: https://github.com/johannes-wolf/typst-canvas


laurmaedje:
I understand the desire for versioning things together, but at the same time it seems like a big waste to download a big package bundle from which only a few KB are actually used, especially in the web app.

laurmaedje:
Do you also value these files being part of the tar.gz bundle? If not, we could consider adding an `exclude` key to the manifest with globs of files that should not be part of the bundle.

johannes-wolf:
> Do you also value these files being part of the tar.gz bundle? If not, we could consider adding an `exclude` key to the manifest with globs of files that should not be part of the bundle.

No, they are only useful for the readme. An `exclude` key would be perfect.

laurmaedje:
I implemented an `exclude` key that takes an array of globs.

johannes-wolf:
Thank you, I have added all the images and the pdf to the `exclude` list.
This PR is ready for review.

packages/preview/cetz-0.0.1/typst.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.0.1"
+repository = "https://github.com/johannes-wolf/typst-canvas"
+entrypoint = "lib.typ"
+authors = [
+  "Johannes Wolf <https://github.com/johannes-wolf>",
+  "fenjalien <https://github.com/fenjalien>"
+]
+license = "Apache-2.0"
+description = "Drawing with Typst made easy, providing an API inspired by TikZ and Processing."
+exclude = [ "*.png", "manual.pdf", "manual.typ" ]
```
laurmaedje, line 12: is it intentional that `gallery` isn't excluded completely? the three Typst files aren't big, but maybe a bit pointless.

johannes-wolf: Done.

laurmaedje:
Thank you!

PR #26, opened 2023-07-06
ouset:0.1.0

ludwig-austermann:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I made some tweaks to the existing package with the same version:
- changed the layout mechanism with minor tweaks, which can be regarded as bug fixing
- added a new function on top of the other functions, for easier usage
- old functions gained a named argument, which gives control over alignment functionality

I have not yet announced this package anywhere, so I guess this fix can still be included in 0.1.0. However, I can also package this as a newer version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,10 @@ Required for submissions to this repository:
 
 Optional:
 - `repository`: A link to the repository where this package is developed.
+- `exclude`: An array of globs specifying files that should not be part of the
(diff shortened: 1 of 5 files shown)


laurmaedje:
As there are actual behavioural changes and not just docs fixes, this would require a new version of the package, either 0.1.1 or 0.2.0 depending on whether it contains breaking changes.

PR #31, opened 2023-07-07
idwtet

ludwig-austermann:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package lets developers write better demonstration code, as it adds special code blocks `typst-ex`, which highlights and displays itself as typst code blocks, but also evaluates itself. By this, examples should never be out of date and gain additionally unified design.


laurmaedje:
Just wanted to note that it's possible to evaluate non-code by doing `eval("[" + markup + "]")`.

ludwig-austermann:
> Just wanted to note that it's possible to evaluate non-code by doing `eval("[" + markup + "]")`.

I don't quite understand what you mean by that. If I understand correctly, this can not be applied in the package, or can something be simplified? Maybe, to clarify the use case:

ouset_and_funarray

laurmaedje:
> To express the locality (and because it is still impossible to evaluate non code), each code block starts in code mode and is also thus displayed as such. (A `#{` and a`}` are added accordingly.)

I was commenting on this, specifically on the "and because it is still impossible to evaluate non code" part. Maybe I misunderstood what you mean with it.

laurmaedje:
Also, it's actually not intentional that package imports work in `eval`. Normal imports don't. I need to think a bit about whether there is a problem with allowing it (maybe not, but my instinct and Python's eval tell me be careful).

ludwig-austermann:
Ah yes that makes sense, thanks. I just rewrote some parts and will improve on the design.

ludwig-austermann:
In my package, one would have to define all imports outside then, with no checking, if the imports are indeed correct in the examples.

laurmaedje:
I don't follow. If the imports are outside, then you couldn't access anything inside the `eval` as it doesn't have access to the surrounding scope.

ludwig-austermann:
Oh, I see, I never actually tested that. Well, the package, for sure, wouldn't be as useful. Maybe one should be able to specify which variables `eval` can access from outside? (Python does this, but python is not famous for its best and well thought through language design)

Anyway, I will change the README accordingly.

laurmaedje:
I'll mark this as draft until the import situation is cleared up.

PR #32, opened 2023-07-07
acrostiche:0.2.0

Grisely:
I am submitting
- [ ] a new package
- [x] an update for a package

This update solves a major bug that returned the acronym without the definition on the first call.
It also fixes a lot of typos in the readme file to avoid confusion.
Finally, it introduces the option for the user to choose or remove the heading of the acronym index.

This update is a breaking change in the sense that the default value for the acronym index changes from `Acronyms Index:` to `Acronyms Index` (removed the trailing colon that has nothing to do here).

### DIFF
--- a/packages/preview/acrostiche-0.2.0/acrostiche.typ
+++ b/packages/preview/acrostiche-0.2.0/acrostiche.typ
@@ -1,0 +1,126 @@
+// Acrostiche package for Typst
+// Author: Grizzly
+
+
+#let acros = state("acronyms",none)
+#let init-acronyms(acronyms) = {
+  acros.update(acronyms)
+}
+
+#let reset-acronym(term) = { 
+  // Reset a specific acronym. It will be expanded on next use.
+    state("acronym-state-" + term, false).update(false)
+}
+
+#let reset-all-acronyms() = { 
+  // Reset all acronyms. They will all be expanded on the next use.
+  state("acronyms",none).display(acronyms=>
+    for acr in acronyms.keys() {
+      state("acronym-state-" + acr, false).update(false)
+  })
+}
+
+#let display-def(plural: false, acr) = {
+  //
+  if plural{
+    state("acronyms",none).display(acronyms=>{
+      if acr in acronyms{
+        let defs = acronyms.at(acr)
+        if type(defs) == "string"{ // If user forgot the trailing comma the type is string
+          defs
+        }else if type(defs)== "array"{
+          if acronyms.at(acr).len() == 0{panic("No definitions found for acronym "+acr+". Make sure it is defined in the dictionary passed to #init-acronyms(dict)")
+          }else if acronyms.at(acr).len() == 1{
+            acronyms.at(acr).at(0)+"s"
+          }else{
+            acronyms.at(acr).at(1)
+          }
+        }else{
+          panic("Definitions should be arrays of one or two strings. Definition of "+acr+ " is: "+type(defs))
+        }
+      }else{
+        panic(acr+" is not a key in the acronyms dictionary.")
+      }
+    })
+  }else{
+    state("acronyms",none).display(acronyms=>{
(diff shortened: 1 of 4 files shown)


packages/preview/acrostiche-0.2.0/acrostiche.typ
```diff
@@ -0,0 +1,126 @@
...
+      acr-list = acr-list.sorted().rev()
+    }
+  
+    // print the acronyms
+    for acr in acr-list{
+      table(
+        columns: (20%,80%),
+        stroke:none,
+        inset: 0pt,
+        [*#acr:*], [#acronyms.at(acr).at(0)\ ]
```
JKRhb, line 120: Would it be possible to remove the colon here or make its inclusion configurable?

```suggestion
        [*#acr*], [#acronyms.at(acr).at(0)\ ]
```

packages/preview/acrostiche-0.2.0/acrostiche.typ
```diff
@@ -0,0 +1,126 @@
...
+  //   level: level of the heading. Default to 1.
+  //   outlined: make the index section outlined. Default to false
+  //   sorted: define if and how to sort the acronyms: "up" for alphabetical order, "down" for reverse alphabetical order, "" for no sort (print in the order they are defined). If anything else, sort as "up". Default to ""
+  //   title: set the title of the heading. Default to "Acronyms Index". Passing an empty string will result in removing the heading.
+
+  // assert on input values to avoid cryptic error messages
+  assert(sorted in ("","up","down"), message:"Sorted must be a string either \"\", \"up\" or \"down\"")
+
+  if title != ""{
+    heading(level: level, outlined: outlined)[#title]
```
JKRhb, line 99: Could also expose other heading parameters like `numbering` to be able to adjust the styling of the index's heading?

JKRhb (commented):
Hi @Grisely! Thank you for providing a new version of the package :) Trying it out, I just noticed two potential points for improvement. Since there is no repository to open issues in, I took the freedom of commenting in this PR, hope that this is fine for you.

packages/preview/acrostiche-0.2.0/acrostiche.typ
Grisely: Will do. Probably make a parameter for the string to append after the acronym, defaulting to ":" for backward compatibility.

packages/preview/acrostiche-0.2.0/acrostiche.typ
Grisely: My idea was that if you want a special heading, remove it from the return by passing `title: ""`and then make your own heading exactly as you want. Do that not work for you?

JKRhb: Ah, sure, that also works :) Thank you!

packages/preview/acrostiche-0.2.0/acrostiche.typ
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  state("acronyms",none).display(acronyms=>{
+    
+    // Build acronym list
+    let acr-list = acronyms.keys()
```
JKRhb, line 105: One more thing: I think at the moment, all defined acronyms are being printed, regardless of whether they are being used in the document or not, right? Would it be possible to filter out those that are actually being used?

Grisely: Yes I know about that. The initial version (before packaging) was able to print only the used acronyms. However, because the package is a different file and does not have the same access to the document, the time-traveling features of `state` don't work the same and I could not implement the smart selection of acronyms in the package. I'll try again with future versions of typst.

JKRhb: Ahh, I see, thank you for the explanation!

PR #36, opened 2023-07-09
bone-document:0.1.0

zrr1999:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: some basic template(resume, document, slides).


istudyatuni:
> Note: Please do not submit templates as packages just yet

laurmaedje:
I'll mark this as draft until we have infrastructure in-place for template packages.

laurmaedje:
We now accept templates into the repository. The updated README details what changes need to be done for a package to become a template. You can try your template package locally with the new `typst init` command by compiling the latest Typst compiler from the main branch. Alternatively, you can wait a bit until the next release drops. If there are any questions or something is unclear with the new process, please point this out. Thank you for your patience!

laurmaedje:
Thanks for updating the template! There are a few problem with it that would need to be addressed before merging:

- The `set heading(level: 2)` breaks headings (as seen in the screenshot). I'm not sure what the purpose of it is, but I think it can just be removed.

- The emphasis should rule breaks toggling of nested emphasis. Since the font specification in it is redundant with the fonts already specified via `set text`, it can probably be fully removed to fix it.

- It's okay that the template changes the default language and fonts, but then it should specify that in the README and note the target audience and the fonts that users will need. The description in `typst.toml` is also too vague, it should describe what specifically this template is about.

- The README has copy-paste errors from the `ieee` template ("exports the `ieee` function" and the details about the author specification).

- The template doesn't really fit with the `paper` category. You can keep just `report` or add `office`.

- It's not forbidden to say "Powered by BoneDocument" in the footer, but I think it will reduce the usefulness of the template to a general audience that doesn't want this in their document. One compromise would be to allow users to enable or disable this with an argument to the `bone-document` function.

laurmaedje:
I'll close this due to inactivity. Feel free to reopen if you want to continue with it!

PR #45, opened 2023-07-19
codelst:0.0.3

jneug:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: **codelst** adds useful features for rendering source code. It can add line numbers, load and show source code from files, highlight lines or add labels to specific lines to make references from other places.

As of now it lacks a comprehensive manual, but the examples should make the usage clear enough.


packages/preview/codelst-0.0.1/README.md
```diff
@@ -0,0 +1,40 @@
...
+```typst
+#import "codelst.typ": sourcecode
+```
+
+After importing the package, simple wrap any fenced code block in a call to `#sourcecode()`:
+
+````typst
+#import "@preview/codelst:0.0.1": sourcecode
+
+#sourcecode[```typc
```
laurmaedje, line 24:
```suggestion
#sourcecode[```typ
```

laurmaedje (commented):
The `sourcefile` function doesn't work unfortunately because the `read` call happens within the package. There is currently no way for a package to read files outside of itself. I also noticed that highlighting for some languages (Typst itself in particular) is suboptimal because each line is highlighted individually (might be a deliberate trade-off, but just wanted to mention it in case you are not aware).

jneug:
> The `sourcefile` function doesn't work unfortunately because the `read` call happens within the package. There is currently no way for a package to read files outside of itself. 

Thats unfortunate but understandable. I didn't test that properly. 

I will remove that function.

> I also noticed that highlighting for some languages (Typst itself in particular) is suboptimal because each line is highlighted individually (might be a deliberate trade-off, but just wanted to mention it in case you are not aware). 

I'm aware of that problem and hope to fix it in future releases. I experimented with different solutions, but they weren't as robust and broke for things like page breaks or long lines.

jneug:
I already did some minor fixes for 0.0.2 and added a manual. With the changes to `sourcefile` I bumped the version to 0.0.3 and updated the pull request accordingly.

`sourcefile` now takes the file content instead of the filename and expects the user to call `read` from the main document:
```js
#sourcefile(read("filename.typ"), lang:"typ")
```

laurmaedje:
Thank you!

PR #52, opened 2023-07-26
minitoc:0.1.0

RosiePuddles:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This provides a function as similar to `outline` as I could, but only for sections under the heading above it

```typst
#import "@preview/minitoc:0.1.0": *
#set heading(numbering: "1.1")

#outline()

= Heading 1
#minitoc()

== Heading 1.1
#lorem(20)

=== Heading 1.1.1
#lorem(30)

== Heading 1.2
#lorem(10)

= Heading 2
```

[Compiled PDF](https://gitlab.com/human_person/typst-local-outline/-/raw/main/example/example.pdf?ref_type=heads)

Here `local_outline` would give an `outline`-like output but only for headings under "Heading 1". For a LaTeX equivalent (why I wrote this int eh first place) have a look at minitoc.


laurmaedje:
I might be wrong, but I _think_ it should be possible to write this as an abstraction on top of `outline` so that you don't need to reimplement numbering, indent etc. You'd basically query for previous and following headings, and then use the outline's `target` parameter to restrict the headings with `before` and `after` (which can take locations). 

Apart from that, I'd suggest using the "more canonical" kebab-case style (i.e. `local-outline`), but that's up to you. :)

RosiePuddles:
It's been changed to an abstraction on top of `outline` now which both makes the code better and works better. One issue I had was that I can't change the heading level for `outline` with `#show outline: set heading(level: 3)` for example which would make `local-outline` with a title more useful because at the moment if it's numbered it's at level 1. But I may be able to find a workaround later

reknih:
Hey, thanks for your package submission! Could you add a bit of "pazazz" to the package name? As-is, it's merely descriptive, which could deceive users into assuming this is an officially endorsed package. Sorry for the late decision!

laurmaedje:
Just a heads up: The package directory layout changed. It's now `packages/preview/package/version` instead of `packages/preview/package-version`.

RosiePuddles:
would minitoc be okay? taking straight from the LaTeX equivalent

laurmaedje:
minitoc is fine! I think renaming the function to `minitoc` too might be clearer, but the function names are completely up to you, you can also keep it as is.

RosiePuddles:
should all be fixed and okay now

packages/preview/minitoc/0.1.0/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+- **title**: The title for the local outline. This is the same as for [`outline.title`](https://typst.app/docs/reference/meta/outline/#parameters-title).
+- **target**: What should be included. This is the same as for [`outline.target`](https://typst.app/docs/reference/meta/outline/#parameters-target)
+- **depth**: The maximum depth different to include. For example, if depth was 1 in the example, "Heading 1.1.1" would not be included
+- **indent**: How the entries should be indented. Takes the same types as for [`outline.indent`](https://typst.app/docs/reference/meta/outline/#parameters-indent) and is passed directly to it
+- **fill**: Content to put between the numbering and title, and the page number. Same types as for [`outline.fill`](https://typst.app/docs/reference/meta/outline/#parameters-fill)
+
+## Unintended consequences
+
+Because `minitoc` uses `outline`, if you apply numbering to the title of outline with `#show outline: set heading(numbering: "1.")` or similar, any title in `local-outline` will be numbered and be a level 1 heading. This cannot be changed with `#show outline: set heading(level: 3)` or similar unfortunately.
```
laurmaedje, line 60:
```suggestion
Because `minitoc` uses `outline`, if you apply numbering to the title of outline with `#show outline: set heading(numbering: "1.")` or similar, any title in `minitoc` will be numbered and be a level 1 heading. This cannot be changed with `#show outline: set heading(level: 3)` or similar unfortunately.
```

laurmaedje:
Thank you!

RosiePuddles:
thanks a lot

PR #107, opened 2023-08-30
rose-pine:0.1.0

oplik0:
I am submitting
- [x] a new package
- [ ] an update for a package

A simple implementation of [Rose Pine theme](https://rosepinetheme.com/) that has some minimal base theming (used with `#show: apply()`) and also simply exposes all colors for all variants of Rose Pine, which is useful in a similar manner to [splash](https://github.com/kaarmu/typst-palettes).

~~Note that this is a draft Pull Request as I'm waiting until I can transfer the repository under rose-pine organization on GitHub (https://github.com/rose-pine/rose-pine-theme/issues/120) in order to update the repository metadata~~ done


laurmaedje (commented):
Could you change the names to LICENSE and README.md in uppercase so that it is uniform with all other packages?

Moreover, I wanted to note that the `template.typ` which is cloned into the three themes could also be expressed via Typst code instead of the `$` replacement. You could define the `theme` function just once and let it take a dictionary. This could probably also work alongside @rose-pine build, such that the generated files only have the dictionary.

packages/preview/rose-pine/0.1.0/src/apply.typ
```diff
@@ -0,0 +1,35 @@
...
+    #it
+  ]
+  #show ref: it => [
+    #set text(fill: theme-data.foam)
+    #it
+  ]
+  #show footnote: it => [
+    #set text(fill: theme-data.pine)
+    #it
+  ]
```
laurmaedje, line 23: sorry for not noticing this before, but I think this would be an improvement
```suggestion
  #show link: set text(fill: theme-data.iris)
  #show ref: set text(fill: theme-data.foam)
  #show footnote: set text(fill: theme-data.pine)
```
for instance for `link` it will prevent extra spaces around the link which might be unintended

oplik0: It's certainly better (I think originally I was thinking of adding more styling which is why I did it that way, but the end result is worse), but I actually don't see the extra spaces in rendered links? The spacing in PDFs generated with both variations in my quick testing appears to be the exact same number of pixels

laurmaedje: I meant that when you have a link directly after text with no space then this will keep it that way while your previous code will add an extra space (because there are spaces in the content block). Alternatively a code block instead of a content block would also have prevented the issue.

PR #140, opened 2023-09-21
Fix typo in nth/typst.toml

fnoaman:
Fixed a typo.


laurmaedje:
Merging this would require approval from the original submitter of the package @extua

extua:
Thanks for the correction! I'm happy for this to be merged.

PR #164, opened 2023-09-29
hidden-bib:0.1.0

pklaschka:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A Typst package to create hidden bibliographies (especially useful when using a Chicago Notes citation style in a letter where the footnote is enough) and/or bibliographies with unmentioned (hidden) citations (the equivalent of `\nocite{}` in TeX).

[GitHub Repository including Examples](https://github.com/pklaschka/typst-hidden-bib)

### DIFF
--- a/packages/preview/hidden-bib/0.1.0/README.md
+++ b/packages/preview/hidden-bib/0.1.0/README.md
@@ -10,6 +10,8 @@ A Typst package to create hidden bibliographies or bibliographies with unmention
 
 In some documents, such as a letter, you may want to cite a reference without printing a bibliography.
 
+This can easily be achieved by wrapping your `bibliography(...)` with the `hidden-bibliography` function after importing the `hidden-bib` package.
+
 The code then looks like this:
 
 ```typ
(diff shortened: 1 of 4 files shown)


packages/preview/hidden-bib/0.1.0/README.md
```diff
@@ -0,0 +1,83 @@
...
+#hidden-bibliography("/refs.yml")
+```
+
+*Note that this automatically sets the `style` option to `"chicago-notes"` unless you specify a different style.*
+
+### Relative Paths in Hidden Bibliographies
+
+Due to the way Typst's `bibliography` function works, you can't use relative paths for the bibliography file with the new function. This is because the `bibliography` function is evaluated relative to the file it is called from, which in this case is the package file that's stored in a different location.
+
+We therefore recommend using an absolute path (from the Typst root directory) like `"/refs.bib"`.
```
laurmaedje, line 30: Absolute paths will unfortunately also not work if it's a package because within a package they are relative to the package root and not the project root. (https://github.com/typst/typst/issues/971)

pklaschka: Ah, damn it! 😅

But thank you for your review and the info. In that case, I'll convert the PR to draft (if it's alright with you to keep it open for a few days) while I ponder if there's some elegant way to do this. I feel that the package should not consist only of the `truly-hidden` function—that would be too abstract and would not provide a great user experience by itself.

One approach I have in mind would be to at least `#set` the `style` parameter in the `truly-hidden` environment so that then, authors will only have to write

```typ
#truly-hidden(
  bibliography("refs.yml")
)
```

and it automatically uses – unless overwritten – the `chicago-notes` style (like the current `hidden-bibliography` function).

I will give this some more thought and re-request review once I have a decision.

laurmaedje: > I'll convert the PR to draft (if it's alright with you to keep it open for a few days)

Sure, no problem at all. The current path situation is unfortunate. :/ It will be fixed at some point.

pklaschka: (Hopefully) resolved by making `hidden-bibliography` a wrapper function for `bibliography` so that the path handling gets performed in the calling file. I think this is the most elegant solution at the moment.

pklaschka (commented):
I've now implemented what, I think, is the most elegant solution to this, so the PR is ready for review again.

Thanks in advance!

PR #190, opened 2023-10-20
Prooftree package

remih23:
I am submitting
- [x] a new package
- [ ] an update for a package

This package allows to easily typeset trees of inference rules, proof trees, type rules, ... 

- [x] agree that my package will not be removed without good reaso

### DIFF
--- a/packages/preview/prooftree/0.1.0/README.md
+++ b/packages/preview/prooftree/0.1.0/README.md
@@ -1,0 +1,66 @@
+# Prooftree
+
+A Typst package for typesetting proof trees. You can import this package with :
+
+```typst
+#import "@preview/prooftree:0.1.0"
+```
+
+## Examples
+
+Here is a basic usage example :
+
+```typst
+#let r = prooftree.rule(name: "Rule name", "Conclusion", "Premise 1", "Premise 2", "Premise 3")
+#prooftree.proof-tree(r)
+```
+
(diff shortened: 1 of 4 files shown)


laurmaedje:
There is already a package with a very similar name called "prooftrees". Could you please choose a name that is more distinct and also not merely descriptive (as described in the README). The name of the other package is merely descriptive as well, which was an oversight, but we can't change that anymore.

packages/preview/prooftree/0.1.0/README.md
```diff
@@ -0,0 +1,75 @@
...
+Result:
+
+<img src="https://github.com/typst/packages/assets/148579668/e349f3c9-d461-48ef-a4f4-d57ef4b940c9" width="480px">
+
+Here is a more complete example, to typeset a proof tree in [natural deduction](https://en.wikipedia.org/wiki/Natural_deduction). First, define the set of rules :
+
+```typst
+#let ax(ccl) = prooftree.rule(name: "ax", ccl)
+
+#let and_i(ccl, p1, p2) = prooftree.rule(name: $and_i$, ccl, p1, p2)
```
laurmaedje, line 27: Maybe these also also use kebab-case to be consistent with the rest?

Thumuss:
I'd like to have an update on this package because the rule at the right of the line is something missing to `prooftrees`. As a student in computer sciences, it could be a real addition for logicians in general.

pauladam94:
I will update today this package. Same base code but with some fix and a new name for the package.

PR #209, opened 2023-10-31
marginalia:0.2.0

jmml97:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: This Typst package provides functions to add (optionally numbered) margin notes to your documents.

This package is based on the `drafting` package but specializes on numbered margin notes, as opposed to *drafting* margin notes. As such, some functions have been removed and the defaults updated. Adds features for numbering the notes and controlling the numbering style, as well as the note text style.

- [X] selected a name in conformance with the guidelines
- [X] ensured that my submission does not infringe upon the rights of a third party
- [X] named this PR as `name:version` of the submitted package
- [X] agree that my package will not be removed without good reason


JamesxX:
This would probably be better as a pull request to the original package surely, rather than copying over their code?

jmml97:
> This would probably be better as a pull request to the original package surely, rather than copying over their code?

I initially did that, but soon realized that the intent of the original package and of my needs for this package were different. I had to change the default settings because my purpose was not to make draft notes but rather to make permanent margin notes with numbers. This package has some functionality added. 

My thoughts were that they could coexist as two different packages.

laurmaedje:
In the end it's up to you, but did you discuss it with the author of the other package? Maybe they'd like your extension to the package.

pklaschka:
I'm not sure, but I think what might be a bit confusing with the PR right now is that it says it's based on the `margin-notes` package while (as per the `README.md`), this seems to be a typo and it is based on the `drafting` package.

And this, then, makes the distinction (and reason for the distinction) much clearer (if the source package was scoped as `margin-notes`, I would agree with @JamesxX, but with the name `drafting`, I agree that the scope of this package feels different and I wouldn't search for this feature in a package called `drafting`).

jmml97:
> I'm not sure, but I think what might be a bit confusing with the PR right now is that it says it's based on the `margin-notes` package while (as per the `README.md`), this seems to be a typo and it is based on the `drafting` package.
> 
> And this, then, makes the distinction (and reason for the distinction) much clearer (if the source package was scoped as `margin-notes`, I would agree with @JamesxX, but with the name `drafting`, I agree that the scope of this package feels different and I wouldn't search for this feature in a package called `drafting`).

Whoops! My mistake! Yes, it is based on the `drafting` package. The default style of that package is one for commenting drafts (hence the name of that package). It does not include numbered margin notes and has some other features that are not needed in a margin notes package. 

> In the end it's up to you, but did you discuss it with the author of the other package? Maybe they'd like your extension to the package.

I could discuss it with him, though I am not sure he would agree to changing the defaults for the `margin-note` function...

reknih:
Thank you for your package submission. We have to ask you to change the name of this package since it is [merely descriptive](https://en.wikipedia.org/wiki/Marginalia). Once the name is changed, the merging process can continue!

ntjess:
👋 Author of the `typst-drafting` package here, I wasn't tagged so I didn't see this until now. I think there are a few ways that, if you want your own package, you can use `#import @preview/drafting:XXX` and build on top of it rather than copy-pasting code -- if you're interested, I'm happy to work with you if changes are needed in `typst-drafting`.

My thought is, if bugs are fixed here or in my package, it will be much easier to reap the benefits if they rely on a common library. I'm even happy to split mine into two packages if the community prefers. It also came up in discord (https://discord.com/channels/1054443721975922748/1148077628960542771/1148188972489654292) but I initially declined since the code base is quite small so far.

Looking at the examples, I think the feature set is a great addition and I welcome a PR if you would consider adding it

ntjess:
> I wouldn't search for this feature in a package called `drafting`

@pklaschka That's a fair point as well -- I added it as a `keyword` and part of the description so that if you use the text filter on https://typst.app/docs/packages, it does show up

laurmaedje:
@jmml97 based on @ntjess's comments, how should we proceed here?

jmml97:
> @jmml97 based on @ntjess's comments, how should we proceed here?

Hello! Sorry, it's been a few busy weeks. I will look at how the features could be integrated into `drafting`. My main concerns about it are two:

* That the `margin-note` command defaults include the line and box decoration, making it more suitable for the drafting focus of that package rather than a more general margin note purpose.
* That because of that drafting focus these functions to provide margin notes are not easily found.

laurmaedje:
Should this be kept open or closed?

jmml97:
> Should this be kept open or closed?

I've been very busy recently and have not been able to work on this. Maybe it is best to close it and then reopen it once I have decided how to do the changes. Thanks for your time.

PR #210, opened 2023-11-01
Added atmosphere to unit list

Jarrod-Angove:
Added atmosphere to unit list 

Commonly used in a lot of my thermodynamics homework so I thought it might be nice to have here ¯\\_(ツ)_/¯

Also removed the quote in the definition for coulomb that was preventing it from working.


laurmaedje:
Please propose changes at the package's repository, not here: https://github.com/ChHecker/unify

PR #215, opened 2023-11-03
secret:0.1.0

GeorgeMuscat:
I am submitting
- [ :white_check_mark: ] a new package

Description: 
This package currently provides a very simple method to safely redact information from rendered documents.

NOTE: Information will not be redacted in the source `.typ` files.

The `redact` function replaces text with a rectangle box. The text is not rendered underneath the box, due to the use of the builtin `hide` function.

- [ :white_check_mark: ] selected a name in conformance with the guidelines
- [ :white_check_mark: ] added a `typst.toml` file with all required keys
- [ :white_check_mark: ] added a `README.md` with documentation for my package
- [ :white_check_mark: ] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [ :white_check_mark: ] ensured that my submission does not infringe upon the rights of a third party
- [ :white_check_mark: ] tested my package locally on my system and it worked
- [ :white_check_mark: ] named this PR as `name:version` of the submitted package
- [ :white_check_mark: ] agree that my package will not be removed without good reason


packages/preview/secret/0.1.0/lib.typ
```diff
@@ -0,0 +1,9 @@
+A simple package to keep your secrets.
+
+#let redact(text, fill: black, height: 1em) = {
+  box(rect(fill: fill, height: height)[#hide(text)])
```
andreasKroepelin, line 4: This can simply be `box(fill: fill, height: height, hide(text))`.

andreasKroepelin (commented):
Not really sure if a one-liner justifies a whole package...

packages/preview/secret/0.1.0/lib.typ
GeorgeMuscat: Good point, will update my code for it, thanks :)

GeorgeMuscat:
> Not really sure if a one-liner justifies a whole package...

Fair point, but I didn't see any particular restriction on the size of a package and I feel like a redaction feature like this could be useful for others. 

I think it would be especially useful for people with a lesser technical background to be able to use a simple function, instead of a more naive and poorly implemented solution that may not redact the information properly.   

The more accessible we make Typst, the better imo.

istudyatuni:
Package name suggestion: `redacted`

GeorgeMuscat:
> Package name suggestion: `redacted`

I am happy to change it to this if others prefer this name :)

laurmaedje:
I think "redacted" wouldn't comply with the "not merely descriptive" naming policy. Not sure about "secret". Background: The idea is that another person can submit a different package with a similar purpose in the future, without one of them being clearly favored through the naming.

Robotechnic:
If your package implements a way to show or redact information on demands with a show rule or something like that, it would be great.
That, way we could create a nonredacted document and a redacted document with the same typst code.

reknih:
Thank you for your contribution! We have to ask you to change the name of this package since both `secret` and `redacted` could be the natural name for packages which redact secrets (see merely descriptive naming rule). By taking up this name, the potential for other packages to improve upon this one is reduced. We can continue merging this once the name is changed!

sitandr:
I don't believe it would be good if the one-liner package would be merged. That situation reminds me of this [article](https://www.davidhaney.io/npm-left-pad-have-we-forgotten-how-to-program/). I understand that this situation is a somewhat different, but I'm sure that if someone knows how to import packages and will find that package among many others, they will probably know about `box` and `hide`.

I don't like polluting packages like this, that will make it harder for newcomers to find a library that suits them. I believe this is much more like a snippet than a package, kind of snippets that are produced in giant amounts at Discord.

I added this snippet to my unofficial [Typst Examples Book](https://sitandr.github.io/typst-examples-book/book/snippets/hidding.html) (still WIP, unfortunately, but there are lots of things already). I think this would be a better place to publish the code.

andreasKroepelin:
> I don't believe it would be good if the one-liner package would be merged. That situation reminds me of this [article](https://www.davidhaney.io/npm-left-pad-have-we-forgotten-how-to-program/).

I was thinking of leftpad (and possibly this article) as well when I wrote my initial comment on this PR.

laurmaedje:
We don't have a clear policy on package size yet, but the policies are also still in-flux (as stated in the README). Based on the discussion and community comments here, I would reject this as too small to justify a package.

PR #228, opened 2023-11-07
wavy:0.1.0

Enter-tainer:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package use jogs to run wavedrom in typst. [wavedrom](https://wavedrom.com/) is a popular js library to draw waveforms. With this people can use wavedrom inside typst.


packages/preview/wavy/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "wavy"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Wenzhuo Liu"]
+license = "MIT"
+description = "Draw digital timing diagram in Typst using Wavedrom."
+
```
laurmaedje, line 8: could you exclude `wavg.svg`? The README itself should be included, but the example SVG isn't necessary (it's also not that small actually at 112KB).

laurmaedje:
Would it makes sense to minify the JS file since it's basically an executable file anyway?

Enter-tainer:
The js is minified and its size reduced to 220K now

laurmaedje:
Great, thanks!

laurmaedje:
I didn't think of it earlier, but I believe we are lacking attribution for `wavy.js`.

Enter-tainer:
> I didn't think of it earlier, but I believe we are lacking attribution for `wavy.js`.

do you mean the license thing? i can take a look

laurmaedje:
What I typically do is create a NOTICE file, it's a bit more lightweight and should fulfill the same purpose. typst also has one.

laurmaedje:
Since in this case it's both MIT maybe it would even suffice to add their copyright statement to the license here. I guess the implications on the scope of the license would be a bit different though.

Enter-tainer:
> What I typically do is create a NOTICE file, it's a bit more lightweight and should fulfill the same purpose. typst also has one.

ok i will do it. btw this may apply to other plugin packages. svg-bob graphviz layout-rs... you may update the package pr template to emph this.

laurmaedje:
yeah, it's true. I need to read up on attribution for binaries like wasm, but you're right that it should be the same...

PR #243, opened 2023-11-17
diagraph 0.2.0

Robotechnic:
I am submitting
- [ ] a new package
- [x] an update for a package

This new version use typst to render labels instead of graphviz. It allows to add any typst element in the node's label.

Currently, it only supports nodes's labels and we will try to support other label types but it may wont be possible.


packages/preview/diagraph/0.2.0/lib.typ
```diff
@@ -0,0 +1,14 @@
+#import "internals.typ": render
```
istudyatuni, line 1: There is no internals.typ

Robotechnic: I forgot to add it to my deploy script, sorry.

istudyatuni:
There is also an extra 0.1.2/README.MD

Robotechnic:
OK, I removed it

laurmaedje:
There are still changes to the 0.1.2 README in the diff. You could submit them as a separate minor fix PR, but a bit unnecessary since people will read the newer README anyway.

Robotechnic:
I revert the README to the hold one by mistake. It's fixed

laurmaedje:
There may be some strangeness going on because I deleted the duplicate README already a while ago: https://github.com/typst/packages/commit/ae8031bb5a38acf9b5363f34c8bc717ec7ad84a4. Git with case-sensitive / case-insensitive file systems is a bit weird.

PR #251, opened 2023-11-20
fletcher:0.1.0

Jollywatt:
I am submitting
- [x] a new package
- [ ] an update for a package

This package is for drawing “diagrams with arrows”, mainly commutative diagrams, finite state machines, and similar things, inspired by `tikz-cd`.

**It is very similar in scope to the `commute` package**, which already exists. This originally started as a fork of that package, but this became a full rewrite. It also depends on `cetz` (unlike `commute`).

Parent repository: ~https://github.com/Jollywatt/arrow-diagrams~ https://github.com/Jollywatt/typst-fletcher

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name in conformance with the guidelines
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
     (A manual is provided with proper documentation.)
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] ensured that my submission does not infringe upon the rights of a third party
- [x] tested my package locally on my system and it worked
- [x] named this PR as `name:version` of the submitted package
- [x] agree that my package will not be removed without good reason

PR #255, opened 2023-11-23
`fletcher:0.1.1`

Jollywatt:
I am submitting
- [ ] a new package
- [x] an update for a package (sort of)

PR #251 added the new package `fletcher` (previously `arrow-diagrams`) and was merged. However, the package contained an error preventing it from being loaded. (I imported some dependencies from `@local` instead of `@preview` in some files.)

This PR amends that and nothing else. ~Importantly, **the version number is still 0.1.0**.~ The version has been incremented to 0.1.1.


laurmaedje:
You should submit this under a new version. Packages must be immutable functionality-wise. If this remained at 0.1.0, people that have already loaded the package once couldn't receive the fix because it might be cached indefinitely.

PR #258, opened 2023-11-24
Fix typo and add typo checking to submission guidelines

MultisampledNight:
(deleted template as listed in it, since this is not a package submission)

Fixes a typo in `README.md` I ran across while looking up on how to install local packages.

It might also be useful to add running [`typos`](https://github.com/crate-ci/typos) (can also be used as an installed binary without GH actions) to the submission guidelines since there are quite a few typos in the packages, too, on the other hand I'm not sure if you're fine with that, so I left it out.


laurmaedje:
We can add it to the guidelines

laurmaedje:
Yeah, thanks!

PR #279, opened 2023-12-05
use caption body to avoid "Agorithm 1: Algorithm 1"

Oyveloper:
I am submitting
- [ ] a new package
- [ x ] an update for a package

Description: Since the caption by default also uses the supplement, the current code shows double "Algorithm n". Using only the caption body fixes the issue

### DIFF
--- a/packages/preview/lovelace/0.1.0/lib.typ
+++ b/packages/preview/lovelace/0.1.0/lib.typ
@@ -115,7 +115,7 @@
               counter(figure.where(kind: "lovelace")).display(it.numbering)
               [: ]
             })
-            it.caption
+            it.caption.body
           }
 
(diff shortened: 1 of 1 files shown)


CGMossa:
I'd this your package?

Oyveloper:
@CGMossa not my package, but the issue bugged me. ended up using the altered version in my thesis, but figured I'd contribute back for others to use

Oyveloper:
Oh, I thought this was that repo, sorry. I'll find the original repo instead

Oyveloper:
Turns out this change is actually implemented in the source repo, but not updated here

laurmaedje:
okay, I'm gonna close this then

PR #283, opened 2023-12-06
tada:0.1.0

ntjess:
I am submitting
- [x] a new package
- [ ] an update for a package

[`tada`](https://github.com/ntjess/typst-tada) is built on top of `tablex`, provide composable functions for aggregating, slicing, and displaying tabular data.
Check out its API and usages in the [manual](https://github.com/ntjess/typst-tada/blob/main/docs/manual.pdf).


JamesxX:
Awesome stuff!

packages/preview/tada/0.1.0/lib.typ
```diff
@@ -0,0 +1,6 @@
+#import "src/ops.typ": *
+#import "src/tabledata.typ": *
+#import "src/display.typ": *
+#import "src/ops.typ"
+#import "src/tabledata.typ"
+#import "src/display.typ"
```
laurmaedje, line 6: are these intentionally star imports (which could potentially leak a lot of internals)? should it maybe be a more selected list of things?

emilyyyylime:
Awesome package! It's clear you spent a lot of time perfecting this.

If I may, I would like to point out tiny things that bothered me looking through the example doc:

1. The Euro sign €, along with many other currency symbols, is traditionally located on the right of the amount. I think this should be configurable with states

2. And this is much more of a nitpick but I personally prefer the 'indices' spelling, and generally it is the more common one among programmers. Consider, maybe, changing the argument name or accepting both? (If this is the current behaviour consider me satisfied)

ntjess:
Thanks for the feedback!

> The Euro sign €, along with many other currency symbols, is traditionally located on the right of the amount. I think this should be configurable with states

You're right, after looking into this, several more rules are locale-dependent than just currency symbol, hundreds, and decimal separators. So, I am holding off entirely on native currencies for this release, until perhaps typst itself carries more locale support.

>  I personally prefer the 'indices' spelling

Most frameworks I've encountered use `index` to also refer to an index array (e.g. pandas/polars), or `indexes` when array-like syntax is less supported (Qt). If you could provide some examples where "indices" is used in the user-facing API, it would be helpful to inform whether indeed the convention should change. Personally, I wanted to ensure searches for "index" still show these methods, which would not be the case for "indices".

laurmaedje:
Seems like there is still some discussion. Please tell me once/if this is ready to merge.

packages/preview/tada/0.1.0/src/ops.typ
```diff
@@ -0,0 +1,182 @@
...
+#let collect(td) = {
+  let (rows, field-info) = (td.rows, td.field-info)
+  let computed-fields = field-info.keys().filter(
+    key => "expression" in field-info.at(key)
+  )
+  let processed-rows = ()
+  for row in rows {
+    for key in computed-fields {
+      let scope = row
+      // Populate unspeicified fields with default values
```
kwfn, line 147:
```suggestion
      // Populate unspecified fields with default values
```

ntjess:
@laurmaedje I've converted the PR into a draft since after a longer test drive, I've considered a few additional scenarios that are worth resolving before the initial release. I'll ping you and reconvert the PR when it's ready for review again. Thanks!

ntjess:
It turns out I spent more time this evening fixing the package than I expected to, so the turnaround is a bit faster 😅 

I'm going to open it back up since my biggest concerns were addressed:
- [x] No longer need to `collect` expressions
- [x] Use `to-tablex` method instead of a `table` dictionary key; it stays more up-to-date as data changes
- [x] Better consolidation of `field-info` from user, type, and default arguments

My recommendation is to wait 1-2 days to see if any other community members have feedback, but I would consider it ready-to-merge otherwise.

ntjess:
@laurmaedje I'm playing with a few datasets to see if any more core features are needed, and realize it would be convenient to allow raw strings (` `` `)as well as str objects (`""`) for `eval`'d fields. Give me just a bit more time to incorporate that feature since it drastically simplifies escaping quotes. Sorry for all the noise! Everything should be ready in the next few days, but for now you can hold off a little longer on merging 😅

laurmaedje:
No worries. :)

laurmaedje:
My final remark would be that it might be useful to have something copy-pastable, including the import statement, in the README to get folks started. The SVG is nice for demonstration, but not great for just starting quickly.

ntjess:
With the latest change, the readme is auto-generated from pandoc, then pngs of each example output are added

packages/preview/tada/0.1.0/typst.toml
```diff
@@ -0,0 +1,29 @@
...
+    "dataframe",
+    "table",
+    "tabular",
+    "agg",
+    "aggregate",
+    "filter",
+    "pandas",
+    "polars",
+]
+
```
laurmaedje, line 20: Now that the `manual.pdf` is part of the submission here, it should be excluded (`exclude = ["manual.pdf"]`) from the package itself.

ntjess:
Done ✅

laurmaedje:
Great, thanks for the submission!

PR #285, opened 2023-12-07
boxaroo:0.1.0

Dherse:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Adds colorful boxes to your document with a few nice presets.


laurmaedje (commented):
might make sense to organize the move the example images into `examples/` and the assets into another folder. as-is, it was kind of hard to see whether the exclude list is actually correct.

laurmaedje:
The whole state business is only necessary to work around the lack of custom settable elements, correct?

PR #297, opened 2023-12-13
moremath:0.1.0

dccsillag:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Additional notation, constants and utilities for doing math in Typst, ranging from conveniences to things that might be frequently present in preambles.

- [?] selected a name that isn't the most obvious or canonical name for what the package does

### DIFF
--- a/packages/preview/moremath/0.1.0/moremath.typ
+++ b/packages/preview/moremath/0.1.0/moremath.typ
@@ -1,0 +1,72 @@
+// === GENERIC ===
+
+// A handy thing to slightly increase parentheses size. E.g.:
+// `big((x - y) - (y - z))` will make the external parentheses a bit bigger
+#let big(x) = math.lr(x, size: 150%)
+#let bigp(x) = big($(#x)$)
+
+// A wrapper to number a single equation
+#let numbered(x) = {
+  set math.equation(numbering: "(1)")
+  x
+}
+
+// TODO: not implies, not implied by, not iff; generic "not" (like `\not` in LaTeX)?
+
+// Handy cursive letters (only uppercase for now)
+#let aa = $cal(A)$
+#let bb = $cal(B)$
+#let cc = $cal(C)$
+#let dd = $cal(D)$
+#let ee = $cal(E)$
+#let ff = $cal(F)$
+#let gg = $cal(G)$
+#let hh = $cal(H)$
+#let ii = $cal(I)$
+#let jj = $cal(J)$
+#let kk = $cal(K)$
+#let ll = $cal(L)$
+#let mm = $cal(M)$
+#let nn = $cal(N)$
+// #let oo = $cal(O)$
+#let pp = $cal(P)$
+#let qq = $cal(Q)$
+#let rr = $cal(R)$
+#let ss = $cal(S)$
+#let tt = $cal(T)$
+#let uu = $cal(U)$
+#let vv = $cal(V)$
+#let ww = $cal(W)$
+#let xx = $cal(X)$
+#let yy = $cal(Y)$
+#let zz = $cal(Z)$
+
+// === PROBABILITY THEORY ===
+
+#let indep = $perp #h(-1em) perp$ // Independence relation
+#let nindep = $cancel(indep)$ // Non-independence relation FIXME
+#let Pr = math.op("Pr") // Alternative notation for probability
+#let Ex = math.op("Ex") // Alternative notation for expectation
+#let Var = math.op("Var") // Variance
+#let Cov = math.op("Cov") // Covariance
+#let ind = math.bb($1$) // Indicator
+#let iid = math.upright("iid")
+
+// === MISCELANEOUS IDENTITIES ===
+
+#let sign = math.op("sign")
+#let argmin = math.op("arg min", limits: true)
+#let argmax = math.op("arg max", limits: true)
+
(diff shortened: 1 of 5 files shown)


dccsillag:
I admit I'm a bit concerned that the name is "too obvious", and am looking for a nice name to replace it. But, if you feel it's sufficiently unique then I'd say this is good to go.

laurmaedje:
I'm also on the fence with the name, so if you'd find a more unique name, that'd be great.

laurmaedje:
I'm also wondering, since this is something you were in favor of in the past, whether it would make more sense to figure out which of those things belong in std and which belong in field-specific packages. Otherwise, we might end up in a LaTeX-like situation where everybody loads three moremath packages to get set up.

dccsillag:
Sure, I'll open a discussion for it in the Discord forge.

dccsillag:
- Bracket sizing

  I've been finding my `bigp` and `big` functions extremely useful. Sure, they are very simple (essentially just `lr(#x, size: 150%)`), but it's very practical to just go from "oh, maybe these parens should be a bit larger", add `bigp` in front of it, and be done with it.

  I feel like this (or something like it) could be added straight into the math namespace.

- Numbered equations

  Once we get the new equation numbering&labelling stuff, the `numbered` function in `moremath` will become a moot point, so I don't particularly care for this getting into std. It's just a temporary hack.

- Handy caligraphic letters

  Given that we have the handy blackboard letters, I kinda wish we had handy caligraphic letters as well (especially since they are, IMO, more common than blackboard letters, ignoring the few really common blackboard letters [e.g., `RR`]). Again, I think this could go directly on std.

  I picked double lowercase letters for it on a whim, and suffered from a conflict on the case of `oo` (with infinity). I think `oo` as infinity is great though, so I opted here to just not make `oo = cal(O)` (you could get that anyways with the `Oh` alias also provided on `moremath`).
  Perhaps there is a better way of doing these shorthands, though.

- Additional notation

  Some of these I believe can be added to std, and possibly directly to the math namespace:

  - `argmin` and `argmax`
  - `sign`

  - _Perhaps_, `ind` (since you can't do `11` for `bb(1)` :P)
  - _Perhaps_, `dist` (which is just an upright `d`)

  I also think that `indep` (along with `indep`, hopefully as something like `indep.not`) should somehow enter the list of symbols in Typst... problem is, it's not in unicode. I remember some people pushed back on Discord a while ago against adding it precisely because of this, so we'd have to take a look at that again.

  The remaining ones (as well as a couple that I really should have added) may work better in dedicated packages. For example, a package for probability theory could include:

  - `Pr`, `Ex`, `Var`, `Cov`
  - `iid`
  - `KL`, `IPM` (not in `moremath` yet)

 One for analysis could include: 

 - Landau notation (maybe even unnecessary! Most people would probably just type the letters directly...), i.e., `oh`, `Oh`, `ohmega`, `Ohmega`, `Thetah`
 - General derivative operator `deriv`
 - ... other stuff I haven't added yet :)

max397574:
> Once we get the new equation numbering&labelling stuff, the numbered function in moremath will become a moot point, so I don't particularly care for this getting into std. It's just a temporary hack.

Perhaps a bit of a controversial idea. I don't even know myself if I agree with it because I haven't rly thought through it.
But perhaps if we know that something will be added in std soon we perhaps shouldn't create packages with hacks for it until then. Because then people will start using the packages where we want them to use std and it will be quite hard to get rid of the "package version" of the thing again.

laurmaedje:
The discussion around std upstreaming while probably take a while. I'm not sure how to proceed here, but it is not my intention to block this package submission forever.

dccsillag:
Yep, it does seem like it will take a while. Perhaps:

1. I rename the package, and
2. I add to the very top of its README that there is ongoing work to upstream at least some parts of it into Typst itself, so to look forward to that

That way, people (_cof cof_ me and collaborators, as well as other interested people) can use the conveniences provided here while we iron things out on the std side (a bit like how things are going with tablex).

laurmaedje:
Sounds good.

CGMossa:
I'd recommend the "mini" prefix, because your intention is for the user to specify the least amount of stuff. You got minienvs, already, and minimath would be a bit better name..

Then at some point it can be a miniverse, where the namespace is so oversaturated with goodies, it would be impossible and redudant to define anything yourself.

I like this stuff, exactly like the package `commath` from latex world.

packages/preview/moremath/0.1.0/moremath.typ
```diff
@@ -0,0 +1,72 @@
+// === GENERIC ===
+
+// A handy thing to slightly increase parentheses size. E.g.:
+// `big((x - y) - (y - z))` will make the external parentheses a bit bigger
+#let big(x) = math.lr(x, size: 150%)
```
Enter-tainer, line 5: it is possible to add more variants like $\bigl( \Bigl( \biggl( \Biggl($ `\bigl( \Bigl( \biggl( \Biggl(` ?

dccsillag: Certainly! I'll take a look into it.

laurmaedje:
Friendly reminder that this PR is still open.

laurmaedje:
I'll close this due to inactivity. Feel free to reopen if you want to continue with it!

PR #305, opened 2023-12-21
codly: Fix icon name in example

domoritz:


Leedehai:
Hi - wondering if you could modify the PR title so it relates to the package? :)

domoritz:
Done.

laurmaedje:
This would require @Dherse's approval.

Dherse:
I would suggest opening a pr on my repo so that I don't need to sync it with the packages repo

domoritz:
Done in https://github.com/Dherse/codly/pull/6

laurmaedje:
As I understood things, this should be closed and the fix will just be delivered through a normal update with the next codly version. If that isn't what was intended, just tell me.

PR #308, opened 2023-12-22
polytonoi:0.1.0

dei-layborer:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Creates a function that allows using Roman letters (plus some symbols) to create polytonic Greek text.  Polytonic Greek is what is used in the study of pre-modern Greek, which is more-or-less any Greek text from the Classical era (500 BC or earlier) through the fall of Constantinople in 1453.  Because it uses different conventions and diacritical marks than Modern Greek, any input system designed for the latter isn't fit for purpose. 


packages/preview/polytonoi/0.1.0/typst.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "polytonoi"
+version = "0.1.0"
+entrypoint = "polytonoi.typ"
+authors = ["Dei Layborer"]
+license = "GPL v. 3.0"
```
laurmaedje, line 6: CI failed. This should be `"GPL-3.0-only"` or `"GPL-3.0-or-later"`.

dei-layborer: Ah ok, thank you.  I saw that it was something related to the license, but my brain decided that it was not up for investigating last night :)

laurmaedje:
Thanks!

PR #310, opened 2023-12-23
add fasvg 0.1.0

ivaquero:
I am submitting

- [x] a new package

`fasvg` is an offline `fontawesome` binding for `typst`. Unlike the `fontawesome` package which uses unicode, `fasvg` uses the SVG icon files, which means

- it can be used in online and offline Typst projects without being affected by the unicode support problems.
- it can be more easily tweaked and customized (e.g. height, width, color).


laurmaedje:
This many files in a package is a bit of uncharted territory. From the submission guidelines:

> Packages should not contain large files or a large number of files. This will be judged on a case-by-case basis, but if it needs more than ten files, it should be well-motivated.

ivaquero:
Yes, it has more than 2000 svg files. I am not sure if there is any work around. It's okay to close the PR.

laurmaedje:
I would close this then.

istudyatuni:
I think, since all icons are the same files except viewBox and path, it's possible to store those data in dict, and construct svgs from this:

```typ
#let data = (
    brand-amazon-pay: (view: "0 0 640 512", path: "M14 325.3c2.3-4.2 5.2-4.9 ..."),
)

#let i = data.brand-amazon-pay
#image.decode("<svg xmlns= ... viewBox\"" + i.view + "\" ... <path d=\"" + i.path + "\"/></svg>", format: "svg")
```

That's still a lot of data, though.

PR #329, opened 2024-01-07
Fail on unknown manifest keys + Add `[tool]` section for 3rd-party tools

tingerrr:
This PR denies unknown fields in both the package manifest and package section of the manifest, ensuring typos are caught by CI. To still allow external configuration, 3rd-party tools can use an entry in the `[tool]` section of the document.

This change is backwards compatible with all existing packages, one notable packages came up during testing, `drafting:0.1.2`, which already used `[tool]` without my prior knowledge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Optional:
   otherwise unnecessarily increase the bundle size. Don't exclude the README or
   the LICENSE.
 
-Thrid-party tools can add their own entry under the `[tool]` section to attach
+Third-party tools can add their own entry under the `[tool]` section to attach
 their own typst specific config to the manifest.
 
 ```toml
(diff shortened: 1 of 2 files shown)


tingerrr:
I switched to an empty `Tool` struct, so we retain `Eq` for `PackageManifest`, keeping the deletions of the diff at a minimum.

README.md
```diff
@@ -46,6 +46,17 @@ Optional:
   otherwise unnecessarily increase the bundle size. Don't exclude the README or
   the LICENSE.
 
+Thrid-party tools can add their own entry under the `[tool]` section to attach
```
laurmaedje, line 49:
```suggestion
Third-party tools can add their own entry under the `[tool]` section to attach
```

README.md
```diff
@@ -46,6 +46,17 @@ Optional:
...
+their own typst specific config to the manifest.
```
laurmaedje, line 50:
```suggestion
their own Typst-specific configuration to the manifest.
```

laurmaedje:
Thank you!

PR #338, opened 2024-01-11
pintorita:0.1.0

taylorh140:
I am submitting pintorita a diagramming package based on pintora.
- [x] a new package
- [ ] an update for a package

Description: it generates UML diagram and gantt charts and other diagrams.


taylorh140:
Some concerns with this package:
1. It can be slow to render images. after quite a few timing test I think this is largely attributed to wasmi. 
  - This might change in the future without any changes to this package. 
3. the js is 2mb which is bigish.
  - It does do quite a lot.

packages/preview/pintora/0.1.0/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 mgt
```
laurmaedje, line 3: Likely a copy-paste error.

packages/preview/pintora/0.1.0/pintora.js
laurmaedje, line 1: There should be some form of attribution for this file. See other packages that wrap JS libraries for examples.

laurmaedje (commented):
The name would need to be changed because it's the most obvious and canonical name for bindings to pintora.

Overall, I'm not sure how I feel about these JS-in-wasm-in-Typst libraries, but that's not specific to this package since there are already some.

packages/preview/pintora/0.1.0/LICENSE
taylorh140: Got it.

packages/preview/pintora/0.1.0/pintora.js
taylorh140: Added a notice file although it endded up being nearly as big as a package. so i excluded it from the included files in the toml.

taylorh140:
This should be better. Thanks for looking it over.

packages/preview/pintora/0.1.0/pintora.js
laurmaedje: Hmm. The NOTICE should be part of the distributed assets but maybe it's okay to not include every single npm package, but rather have a NOTICE for the whole library. Similar to how it's done for `mino`: https://github.com/typst/packages/blob/2b0429a0c97a6187395c4b2b1232e45864ea1ae7/packages/preview/mino/0.1.0/NOTICE

taylorh140: Ahh, that should be doable. I think ill also take the opportunity to target the new jogs for better error indication.
It will take a minute though.

laurmaedje:
Thanks!

JKRhb:
@taylorh140 Thank you for providing this package! :) Since there have been a couple of updates to Pintora in the meantime, would it be possible for you to create an updated version of this package as well?

taylorh140:
@JKRhb 
Wow! I'm so glad that people are using it. 
I could probably get it up to date.
Ill move it up on my priority list.

JKRhb:
> @JKRhb Wow! I'm so glad that people are using it. I could probably get it up to date. Ill move it up on my priority list.

That would be great :) Although it can get a bit slow if you have many diagrams, it is very useful for shorter documents and to create very solid first drafts of more complex ones. Having the package is a really valuable addition, I think :)

taylorh140:
Yes I’m very aware of the slowness. I almost didn’t release it. But the
good news is that there is a chance it can be a lot faster. I’m thinking
it’s the wasmi wasm interpreter causing quite a lot of slowdown. But it is
possible that it will be replaced with a much faster backend.

PR #355, opened 2024-01-24
wordometer:0.1.0

Jollywatt:
I am submitting
- [x] a new package
- [ ] an update for a package

Provides a simple interface for basic word counts and document stats. Usage example:
```typ
#import "@preview/wordometer:0.1.0": word-count, total-words

#show: word-count

In this document, there are #total-words words all up.

#word-count(total => [
	The number of words in this block is #total.words
	and there are #total.characters letters.
])
```

Features:
- Ability to exclude content by element function name or by a label.
- Ability to use custom counting functions (e.g., if you wanted to count vowels, or use a different definition of “word”, etc).

It works by recursively traversing content elements until hitting `text`/`raw` nodes.

Repo: https://github.com/Jollywatt/typst-wordometer


Jollywatt:
Actually, please hold off on accepting this. Found some bugs when testing on larger documents.

laurmaedje:
Is there a particular reason the exclusions are strings compared via `repr(it.func())` instead of directly comparing with the functions?

Jollywatt:
Originally it was because not all element functions are in global the namespace (e.g., `space`, `update`, and `display` elements), so I didn’t know how to test for them without `repr` — is there a better way?

But I agree, it would be better for the user to be able to provide the element functions themselves. I’ll make that change soon.

Jollywatt:
Pausing this yet again, because I think the word count can be made much more efficient.

I did a (very naive) benchmark comparing these two methods:
A. Walk the content tree, count the words at each leaf, then sum the results
B. Convert the content tree to one giant string, then count the words
Currently, method A. is implemented.

On a 4k-word technical paper, cold-compile times are 2.52s for A, 1.74s for B, and 1.63s control (no word count). So it’s probably better to use that one by default!

Jollywatt:
Ok, this is well and truly ready! 😅

CGMossa:
Amazing work, and incredibly useful.

ntjess:
Neat stuff! Would you consider adding a user function to get the text itself rather than just the count? It seems like most of the logic is already present and this would be quite useful

If it would take awhile, it can always be a PR/addition for the next package version instead of this one :)

Jollywatt:
@ntjess Added! I’m curious about the use cases you have in mind, though. The function I added produces pretty ugly output (e.g., text is interspersed with extra spaces to make sure consecutive elements aren’t joined), which is fine for word-counting, but might not be for other things.

ntjess:
Sure! One use case I'm interested in is a [word cloud](https://en.wikipedia.org/wiki/Tag_cloud) of the document instead of just a word count. So if you knew all the words, you can size each one based on its number of occurrences :)

ntjess:
Another interesting idea! Down the line, `wordometer` seems like a great platform to support readability metrics like those shown on [wordcounter.net](https://www.wordcounter.net):
image

PR #357, opened 2024-01-26
wrap-it:0.1.0

ntjess:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Stand-in for wrapping figures around text while we await official support


laurmaedje (commented):
Really cool stuff.

In this image

there is this bit of space after "corpore". Is that because a paragraph ends there in Typst's view even though it's technically not the end of the paragraph? If yes, perhaps adding a `linebreak(justify: true)` there would help.

ntjess:
> perhaps adding a linebreak(justify: true) there would help

Thanks for the suggestion. This does work, but I don't want to enable it unconditionally since text may not be justified. I was under the impression that `par(linebreaks: auto)` would handle this circumstance, but that fix doesn't work here. Is there a current method to know whether the environment is justified?

If not, in your opinion is it worth adding a `justify` argument to inform `wrap-content` whether things are justified?

laurmaedje:
> Is there a current method to know whether the environment is justified?

No, I don't think that's possible at the moment. Well, with extreme hackery there might be some way to observe a layout difference on some well-known content with measure...

> If not, in your opinion is it worth adding a justify argument to inform wrap-content whether things are justified?

I think it _might_ be useful, but that's totally up to you.

ntjess:
> observe a layout difference on some well-known content

I was playing around with this exact idea -- place a label at the end of a text run and observe the average ending position. But there's no way to get an answer without placing the content in the document, which isn't ideal. So I think I'll call this a fundamental limitation for now, and propose manual user linebreaks if this is their situation. If enough people complain about the need, I'll add it in 0.1.1

I have a similar rational for a `inset` parameter, rather than requiring the user to `box` their own figures.

laurmaedje:
Sounds good. Thanks for the cool package!

PR #361, opened 2024-01-27
chromo:0.1.0

Julien-cpsn:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: Easily generate printer tests by calling functions. Very usefull for large scale printing, like exam papers. Only CMYK for now. 

For example: 
image

(theses are all made with Typst and not basic images as they can easily be degraded by tweaking the document size)

- - -


laurmaedje:
As is, the name is too obvious for the package's functionality. It would be great if you could pick a more unique name. Please also exclude the examples from the package by adding an `exclude = ["examples"]` line in `typst.toml`.

Julien-cpsn:
Thanks for your quick feedback
Just changed the things mentionned!

In case you wonder,
**Chromo** comes from *"Chromo"* in French, which means "image with bad color taste".

laurmaedje:
Thanks!

PR #380, opened 2024-02-16
anatomy:0.1.0

E8D08F:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A package to visualise the metrics information (horizontal lines) of a font. Useful if user wants to confirm the position of `top-edge` and `bottom-edge` when calling `text()`.


packages/preview/anatomy/0.1.0/lib.typ
```diff
@@ -0,0 +1,77 @@
...
+        bottom-edge: if above-baseline { "baseline" } else { edge }
+      )[.], styles)
+
+      state(edge).update(
+        (if above-baseline { 1 } else { -1 } * frame.height)
+      )
+    }
+
+    for edge in top-edges { set-metric(edge) }
+    for edge in bottom-edges { set-metric(edge, above-baseline: false) }
```
laurmaedje, line 28: It's not really any of my business, but why does this need `state`? It seems like the measurement could just happen in a `locate(loc => style(styles => { ...` without any need for state.

E8D08F: Thank you for your advice. I was thinking about the values of metrics (`ascender` height, etc.) can be used somewhere else in future versions, and exposed to users. I wonder whether there is a simpler way to store global values without using `state`.

packages/preview/anatomy/0.1.0/lib.typ
```diff
@@ -35,7 +35,10 @@
     let stroke_width = 0.5pt
     let description-size = (1 + 2 / 3) / 10 * 1em
 
-    block(width: 100%, above: 0pt, {
+    block(
+      width: 100%,
+      above: description-size,
+      below: description-size, {
```
E8D08F, line 41: Add extra spacing at the top and the end of the `metrics`, so users can have the preferable spacing without surrounding `metrics` by another `block`.

packages/preview/anatomy/0.1.0/lib.typ
laurmaedje: A function like locate/style/.. that takes a closure and provides the measured metrics would be more idiomatic I'd say.

E8D08F: I have removed `state`s (using `style(styles => locate(loc => { ... })`. Does the order matter?).

I used closure for additional typesetting. I’m afraid that I couldn’t find documentation of how to pass a closure as parameter, so it might not be the standard way.

Thank you for your time.

laurmaedje: The order doesn't really matter. It also wasn't my intention to nitpick your package implementation, I was just curious.

laurmaedje:
Thanks!

PR #387, opened 2024-02-21
easytable:0.1.0

monaqa:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: a simple package for writing tables in Typst.
This aims for clear, straightforward markup and some flexibility.

```typst
#easytable({
  th[Header 1 ][Header 2][Header 3  ]
  td[How      ][I       ][want      ]
  td[a        ][drink,  ][alcoholic ]
  td[of       ][course, ][after     ]
  td[the      ][heavy   ][lectures  ]
  td[involving][quantum ][mechanics.]
})
```

image

    - Is "easytable" not obvious?


packages/preview/easytable/0.1.0/README.md
```diff
@@ -0,0 +1,114 @@
+# Typst Easytable Package
+
+A Typst library for writing simple tables.
+
+## Usage
+
+**WARNING: This library is not in `preview` namespace.**
```
laurmaedje, line 7: This warning should be removed before submission to the preview namespace.

laurmaedje (commented):
The name is fine. :)

emilyyyylime:
and wouldn't those be better described by `tr`?

monaqa:
Thank you for your comments, I removed the warning and added some explanation for elements (`td`, `th`, and so on).

laurmaedje:
I think what @emilyyyylime wanted to say is that `td` (table data) in the code should be `tr` (table row) because it describes a row rather than a cell.

monaqa:
Sorry, I misunderstood.

I wondered whether to use `th` & `td` pair or `th` & `tr` pair too, and I chose `th` & `td` in terms of the contrast between Header Cells and Data Cells.
If the combination of `th` and `tr` is comfortable for many people, I think it could be a pair of `th` and `tr`. What do you think?

laurmaedje:
I think by using the HTML terms there is an expectation that it works like HTML, so I would expect `tr` for a row.

monaqa:
Okay, renamed: `td` -> `tr`.

laurmaedje:
Thanks!

PR #389, opened 2024-02-22
Fix version for g-exam:0.1.0

andres-gimenez:
The image was looking for them in the directory where the package is, now it is looking for it in the directory where the .typ of the document is.

I am submitting
- [ ] a new package
- [x] an update for a package


laurmaedje:
Functional changes and bug fixes must be submitted with a new version. Except for minor docs fixes submitted packages are immutable. Otherwise people that already have a cached version can't get the fixes.

PR #394, opened 2024-02-24
Add template bundling

reknih:
Add template bundling support and the Typst IEEE paper template. This adds the following:

The `packages.templates` array in the `typst.toml`: Packages can now declare one or more template starting points. These are directories that are pasted into a new user project / directory when using the web app or `typst init`. This array contains of maps with the following keys:

- `name`: A user-facing name for the entrypoint. Required.
- `description`: A description of what this entrypoint is intended for and how it differs from the others.
- `path`: A path relative to the package root for the template directory. Required.
- `entrypoint`: A path relative to the _template root defined in `path`_ for the main Typst file of the project. Required.
- `thumbnail`: A path relative to the _template root defined in `path`_ for a PNG or WebP image showing a page of the initialized template. Required.

`name`, `path`, and `entrypoint` are required by the compiler.

This still remains to do:
- [x] Add category key
- [x] Add all other templates from typst/templates
- [x] Add docs
- [x] Check that all the templates work


packages/preview/unequivocal-ams/0.1.0/template/main.typ
```diff
@@ -0,0 +1,93 @@
+#import "@preview/unequivocal-ams:0.1.0": ams-article
```
istudyatuni, line 1:
```suggestion
#import "@preview/unequivocal-ams:0.1.0": ams-article, theorem, proof
```

reknih: Thank you!

reknih:
I also wanted to check whether there is just one author affiliation (in square brackets) but `gviz:0.1.0` would be in violation here with `authors = ["Sekoia <https://codeberg.org/Sekoia> <https://github.com/SekoiaTree>"]`. I can add an exception if the check is desired.

CATEGORIES.md
```diff
@@ -0,0 +1,79 @@
...
+  charts, and research paper formats.
+- **`economics`:** Economic modeling tools, financial charting packages, and
+  market analysis templates.
+- **`engineering`:** Packages and templates for all engineering disciplines,
+  including but not limited to Electrial, Civil, and Mechanical engineering.
+- **`computer-science`:** Packages for algorithm visualization, code formatting,
+  and templates for Computer Science periodicals.
+- **`law`:** Legal document templates, case law formatting, and citation tools
+  for legal writing.
+- **`arts`:** Templates for the Humanities.
```
frozolotl, line 78: I think `arts` should be `humanities` instead. Arts seems like a much more specific term.

CATEGORIES.md
```diff
@@ -0,0 +1,79 @@
...
+- **`music`:** Packages for musical notation and composition aids.
```
frozolotl, line 79: Additional ideas:
- `social-science`, or to be more precise---similar to the natural sciences:
  - `geography`
  - `linguistics`
  - `psychology`
  - `sociology`
  - `political`
- `journalism`
- `business`
- `medicine`
- `administration`

Alternatively, do something in the style of MIME types: e.g. `humanities/history`, or `applied/law`. You can use the list at https://en.wikipedia.org/wiki/Outline_of_academic_disciplines as inspiration.

README.md
```diff
@@ -45,9 +47,81 @@ Optional:
...
+in their `typst.toml` file. A template package's `typst.toml` could look like
+this:
+
+```toml
+[package]
+name = "charged-ieee"
+version = "0.1.0"
+entrypoint = "template.typ"
+authors = ["Typst GmbH <https://typst.app>"]
+license = "Unlicense"
```
frozolotl, line 80: Not very important, but I think we should mention a different license here because the Unlicense is [not a good license](https://softwareengineering.stackexchange.com/questions/147111/what-is-wrong-with-the-unlicense).
Maybe use `MIT OR Apache-2.0` instead? Or `EUPLv1.2`, though that one isn't as well known. I must admit that there isn't really one obvious choice for packages and templates.

packages/preview/charged-ieee/0.1.0/journal-paper/main.typ
```diff
@@ -22,7 +22,7 @@
     ),
   ),
   index-terms: ("Scientific writing", "Typesetting", "Document creation", "Syntax"),
-  bibliography-file: "refs.bib",
+  bibliography-file: bibliography("refs.bib"),
```
frozolotl, line 25: Maybe rename it to `bibliography` instead of `bibliography-file` now?

packages/preview/unequivocal-ams/0.1.0/template.typ
```diff
@@ -0,0 +1,250 @@
...
+  numbering: if numbered { "1" },
+)
+
+// And a function for a proof.
+#let proof(body) = block(spacing: 11.5pt, {
+  emph[Proof.]
+  [ ] + body
+  h(1fr)
+  box(scale(160%, origin: bottom + right, sym.square.stroked))
+})
```
Andrew15-5, line 250: Files should always end with LF character if it doesn't break some workflow (rare cases). (Some text editors can automatically add the last LF, but others can't.)

packages/preview/appreciated-letter/0.1.0/letter/main.typ
```diff
@@ -0,0 +1,22 @@
...
+  date: [Morristown, June 9th, 2023],
+  subject: [Revision of our Producrement Contract],
+  name: [Jane Smith \ Regional Director],
+)
+
+Dear Joe,
+
+#lorem(99)
+
+Best,
```
Andrew15-5, line 22: Missing LF character here as well. So a63e44395fda2a08f9d3cc9cbd9949fe83b7fac3 only contains 2 files with missing LF.

README.md
CGMossa: typst is already Apache-2.0, so that works?

laurmaedje: Switched to MIT for package and MIT-0 for templates. See the updated README for more details.

frozolotl: Sounds good to me

PR #404, opened 2024-03-03
Fix GitHub handle parsing and use better email address validation

frozolotl:
This PR switches out the email address validation especially in order to support SMTPUTF8. It uses a small crate for that because it is implemented as described in the RFCs.


laurmaedje:
Thanks! I had briefly checked for a crate, but didn't find a good one with few deps. Have missed this one.

PR #408, opened 2024-03-07
typst-ifsc:0.1.0

gabrielluizep:
I am submitting
- [x] a new package
- [ ] an update for a package

Adding a Typst templates for [Federal Institute of Santa Catarina](https://www.ifsc.edu.br/).


gabrielluizep:
Should I was checking the other templates added to the repo, and I noticed that every template has its own package.

I think I misunterstood the idea, will create a new pull request with the correct way, one template per package.

reknih:
Thanks for being the first person to submit a template!

We had a quick look and have the following tips:
- Generally, you are right that this should be multiple template packages rather than a big one
- The two presentation templates can remain a single package, you should be able to switch between the themes with a show rule parameter
- The template's entrypoint file should define a show rule as you did in your examples. Basically, the templates _are_ the examples.
- The assignment seems to be empty
- The package names need to be adjusted. They must not contain the string `typst` and need to be more unique in case another member of IFSC wants to publish their own, see the README of this repo.

gabrielluizep:
Thanks! Will fix these points and submit a new PR!

gabrielluizep:
Created new PR #409

PR #420, opened 2024-03-15
delegis:0.1.0

pklaschka:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A package and template for creating legislative documents (such as bylaws for non-profits) using Typst. While the template is focused on German documents (such as a _Satzung_ and _Vereinsordnungen_ for non-profits, which is also the basis of this template: the Satzung of WüSpace e. V.), any strings used in the template can be customized, making it possible to use this in other languages, as well.

Note that the package is maintained by WüSpace e. V. (https://github.com/wuespace), and not necessarily myself individually.


packages/preview/delegis/0.1.0/template/main.typ
```diff
@@ -0,0 +1,56 @@
+#import "@preview/delegis:0.1.0": *
+
+#show: it => delegis(
+  // Metadata
+  title: "Vereinsordnung zu ABCDEF",
+  abbrev: "ABCDEFVO",
+  resolution: "3. Beschluss des Vorstands vom 24.01.2024",
+  inEffect: "24.01.2024",
```
laurmaedje, line 8: while not strictly required, I would suggest using `kebab-case` for variable and argument names as it is the idiomatic Typst case convention

pklaschka: Ah, true. I'll fix that (I worked too much in other languages this week, still working on getting back into the Typst way of things).

pklaschka:
(I've adjusted it to kebab-case and added a bit more documentation, so this should be ready for review again. Thanks!)

packages/preview/delegis/0.1.0/template/main.typ
```diff
@@ -0,0 +1,60 @@
+#import "@preview/delegis:0.1.0": *
+
+#show: it => delegis(
+  // Metadata
+  title: "Vereinsordnung zu ABCDEF",
+  abbrev: "ABCDEFVO",
```
laurmaedje, line 6:
```suggestion
  abbreviation: "ABCDEFVO",
```

pklaschka: Good catch – thanks! Sorry, that one slipped through the cracks.

laurmaedje:
Thanks!

PR #429, opened 2024-03-16
scrutinize:0.2.0

SillyFreak:
I am submitting
- [ ] a new package
- [x] an update for a package

Scrutinize 0.2.0 updates it to Typst 0.11.0, using `context` to simplify the API and `--input` to more easily specify if a sample solution is to be generated. Some documentation and metadata errors in the 0.1.0 submission were also corrected.


packages/preview/scrutinize/0.2.0/typst.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.2.0"
+entrypoint = "src/lib.typ"
+authors = [
+  "Clemens Koza <https://github.com/SillyFreak/>"
+]
+license = "MIT"
+description = "A library for building exams, tests, etc. with Typst"
+repository = "https://github.com/SillyFreak/typst-packages/tree/main/scrutinize"
+keywords = ["exam", "test", "school", "teaching"]
+compiler = "0.11.0"
```
laurmaedje, line 12: You might want to use the opportunitity to add [`categories`](https://github.com/typst/packages/blob/main/CATEGORIES.md) and possibly also [`disciplines`](https://github.com/typst/packages/blob/main/DISCIPLINES.md).

packages/preview/scrutinize/0.2.0/typst.toml
```diff
@@ -0,0 +1,19 @@
...
+entrypoint = "src/lib.typ"
+authors = [
+  "Clemens Koza <https://github.com/SillyFreak/>"
+]
+license = "MIT"
+description = "A library for building exams, tests, etc. with Typst"
+repository = "https://github.com/SillyFreak/typst-packages/tree/main/scrutinize"
+categories = ["model", "scripting", "office"]
+disciplines = ["education"]
+keywords = ["exam", "test", "school", "teaching"]
```
SillyFreak, line 13: @laurmaedje  I chose both model and scripting because the grading aspect of the library adds semantic information to the document, and then provides scripting tools to process that information. Also "office" because for a teacher, exams are similarly "day-to-day" as an invoice.

Do these match your notions of the categories? I have no problem changing them again, I just thought it's more efficient to simply put what I thought best there, in case I got it right.

laurmaedje: Sounds good.

PR #430, opened 2024-03-17
slydst:0.1.1

glambrechts:
I am submitting
- [ ] a new package
- [x] an update for a package

A bug where headers and footers were contaminated by the styles of headings was fixed.

PS: This was due to the use of `show` rules like `show heading.where(level: 1): x => pagebreak() + x.body + pagebreak()`, causing the headings to span over several pages.


packages/preview/slydst/0.1.0/.gitignore
laurmaedje, line 1: the old version of the package shouldn't be changed

glambrechts: Ok no problem, I just deleted `.gitignore` files.

PR #383, opened 2024-02-17
g-exam:0.1.0

andres-gimenez:
I am submitting
- [x] a new package
- [ ] an update for a package

Create exams with student information, grade chart, score control, questions, and sub-questions.

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name that isn't the most obvious or canonical name for what the package does
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] `exclude`d PDFs or README images, if any, but not the LICENSE

PR #433, opened 2024-03-17
g-exam:0.2.0

andres-gimenez:
I am submitting
- [ ] a new package
- [x] an update for a package

It fixes margin issues and includes the ability to include solutions to questions.


laurmaedje:
As mentioned [here](https://github.com/typst/packages/pull/383#issuecomment-1980705156), I think this package should actually be a template (the `#show: g-exam.with(author: .., ..)` is a strong indicator of that), so you might consider [converting it into a template](https://github.com/typst/packages?tab=readme-ov-file#templates).

andres-gimenez:
> As mentioned [here](https://github.com/typst/packages/pull/383#issuecomment-1980705156), I think this package should actually be a template (the `#show: g-exam.with(author: .., ..)` is a strong indicator of that), so you might consider [converting it into a template](https://github.com/typst/packages?tab=readme-ov-file#templates).

I'm working on the templates. Can  be created multiple entrypoints?

laurmaedje:
No, there is only one entrypoint. We plan to make that entrypoint configurable in the future (so that users could click UI or answer questions to set up their project), but that doesn't exist yet.

packages/preview/g-exam/0.2.0/template/main.typ
```diff
@@ -0,0 +1,36 @@
+#import "@preview/g-exam:0.2.0": g-exam, g-question, g-subquestion
+
+#show: g-exam.with(
+  school: (
+    name: "My School",
+    logo: read("./logo.png", encoding: none),
```
laurmaedje, line 6: I would use `image("logo.png")` rathen than `read` here. To affects it sizing, you can then do the following in your template:
```
set image(width: 2.5cm, fit: "contain")
logo
```

andres-gimenez: I was considering using image, but I thought it was better to put as little dimension as possible in the exam document. I would like this to be used by unenthusiastic teachers of markup languages. You have to make it very easy to use, that's why I'm interested in sugar syntax.

I'm not going to use the image set, because it changes the size of all the images, I've changed read to image in the home document.

I don't know what you think, but in the next version, I'm going to put several examples inside the screen, as an example. The way I work is to create exams in the same folder. It is not the same to create a thesis, in which you work on it over a long period of time, than to work on creating exams, in which you create several in a month.

laurmaedje: > Can I set a maximum width for a piece of content?

You can use a `block` to force a specific width. Maximum width is trickier, though probably possible with `measure`.

packages/preview/g-exam/0.2.0/typst.toml
```diff
@@ -0,0 +1,18 @@
...
+name = "g-exam"
+version = "0.2.0"
+compiler = "0.11.0"
+entrypoint = "g-exam.typ"
+authors = ["Andrés Giménez Muñoz"]
+license = "MIT"
+description = "Create exams with student information, grade chart, score control, questions, and sub-questions."
+keywords = ["exam", "test", "school", "question", "subquestion", "solutions", "teaching"]
+repository = "https://github.com/MatheSchool/typst-g-exam"
+categories = ["visualization", "paper", "office"]
```
laurmaedje, line 11: I didn't notice this the first time around, but I think only `office` makes sense from these. `visualization` is for plotting and stuff, while `paper` is for academic papers submitted to conferences.

andres-gimenez: The template is oriented to the academic world. If you want to separate it from the research, you could create a new category, which has a lot of potential users. By exmaple ```school``` or ```academic``` can refer to academic world.

laurmaedje: That's what `disciplines = ["education"]` is for, which the template already contains.

PR #439, opened 2024-03-18
curryst:0.2.0

pauladam94:
I am submitting an update for a package.

## Changes :
- label on the right or on the left
- better typesetting of very long horizontal rule
- refactor of the code
- lot more example
- lot more images of those examples


MDLC01:
`README.md` contains a HTML `details` element. I'm not sure whether this will be rendered properly on Typst: Universe.

laurmaedje:
The details element should be okay. I'm not sure how it will be styled, but it shouldn't be a problem in itself.

MDLC01:
Oh and I just realized it uses a HTML definition list too!

Typst: Universe doesn't look like it has any styling for details.

PR #442, opened 2024-03-19
lucky-icml:0.2.1 and bloated-neurips:0.2.1

daskol:
The origin for templates is [daskol/typst-templates][1] repo.

- Add ICML 2024 template.
- Add NeurIPS 2023 template.

---

[1]: https://github.com/daskol/typst-templates/tree/main

I am submitting
- [x] a new package
- [ ] an update for a package

Description: These two packages provide a templates for ones of the largest conferences in machine learning community. Also, there is clear request from community to contribute these templates to Typst packages (see issue https://github.com/daskol/typst-templates/issues/3). Specifically, ICML template was used to prepare submission at the conference (arXiv [preprint](https://arxiv.org/abs/2402.01376)).


laurmaedje:
Thanks for the great template submissions! The only thing that needs to be addressed before merging is the naming. Template names must have at least one unique, creative component so that multiple people can create a template for the same instutition/venue without one having an unfair advantage. For organization-specific templates, we recommend the naming scheme `<adjective>-icml`. See the [submission guidelines](https://github.com/typst/packages?tab=readme-ov-file#submission-guidelines) for details. You can also take a look [at the existing templates](https://typst.app/universe/search?kind=templates) for inspiration.

daskol:
Now I see why `charged-ieee` is named so. Fixed.

If I were a regular Typst user, I would prefer something like `@preview/<user-or-org>/<template>`, `@<github-user-or-org>/<package>`, or event `@github.com/...` (golang-style packages).

laurmaedje:
We considered various approaches back when we started the package manager and settled on the creative naming idea for packages. It's much easier to tell people to "use cetz" rather than to say "use Johannes Wolf's canvas package" (the cetz package used to be called typst-canvas before the package manager). 

I know that it's a bit unconventional, but I think so far it worked out quite well. Some of the best Typst packages have some of the most memorable names. With templates, we felt there was more of a need for some descriptive component, so we recommend some combination.

Another benefit of the approach is that it's simple for a package to switch maintainers. And we don't need to manage a secondary namespace of user names.

laurmaedje:
I noticed two problems while testing the templates locally:
- `#import "@preview/bloated-neurips": *` is missing the version
- `lucky-icml` has a problem with the bibliography-file. The path handling with packages in Typst is a bit problematic. I recommend to put the bibliography call itself into the main file. You can look at `charged-ieee` to see how it's done there.

daskol:
Fixed. I bumped patch version in order to keep template code in consistence with upstream repo.

daskol:
> I know that it's a bit unconventional, but I think so far it worked out quite well. 

Totally agreed. It is the most important point in my option. It should just work first. :-)

laurmaedje:
This symlink is the best way I know of and also what I use to test.

packages/preview/lucky-icml/0.2.1/icml2024.typ
```diff
@@ -0,0 +1,701 @@
...
+  v(0.2in)
+
+  columns(2, gutter: 0.25in, {
+    set text(size: font.normal)
+    show par: set block(spacing: 11pt)
+    // Render abstract.
+    // ICML instruction tels that font size of `Abstract` must equal to 11 but
+    // it does not like so.
+    text(size: font.large)[
+      align(center)[*Abstract*]
```
laurmaedje, line 544: Unfortunately, I found another bug. There's a hash missing here. So it literally says `align(center)[Abstract]`.

daskol: Thanks. Fixed.

laurmaedje:
Thanks!

PR #452, opened 2024-03-19
modern-unito-thesis:0.1.0

eduardz1:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Template for the thesis for my city's university - University of Turin


packages/preview/unito-thesis/0.1.0/lib.typ
```diff
@@ -0,0 +1,319 @@
...
+  // Set to "it" for the italian template
+  lang: "en",
+
+  // The thesis' bibliography, should either be passed as a Hayagriva
+  // .yml file or a BibTeX .bib file or `none` if you don't need
+  // to include a bibliography
+  bibliography-path: none,
+
+  // The university's logo, should be a path to the logo file
+  logo: "imgs/logo.svg",
```
laurmaedje, line 50: Supplying a project-relative path here will unfortunately not work properly (it will be relative to the package). If the logo is always the same, I would just put it into the package itself rather than the template directory.

Otherwise, I suggest having `image("imgs/logo.svg")` here and users could then have their own project-relative image call. If you need to adjust image parameters in your template, you can write:
```
set image(..)
logo
```

packages/preview/unito-thesis/0.1.0/lib.typ
```diff
@@ -0,0 +1,319 @@
...
+  // keyword, `school` keyword and a `degree` keyword
+  affiliation: (),
+
+  // Set to "it" for the italian template
+  lang: "en",
+
+  // The thesis' bibliography, should either be passed as a Hayagriva
+  // .yml file or a BibTeX .bib file or `none` if you don't need
+  // to include a bibliography
+  bibliography-path: none,
```
laurmaedje, line 47: Bibliography has the same problem as images, it should also be a `bibliography("...")` call. You can also take a look at the charged-ieee template to see how it's done there. The whole path handling will be improved in the future, but in the meantime, packages unfortunately have to work around it.

laurmaedje (commented):
Thanks for the great template submission! I left a few comments because the project currently does not work correctly when checked out with `typst init`.

One other thing that needs to be addressed before merging is the naming. Template names must have at least one unique, creative component so that multiple people can create a template for the same instutition without one having an unfair advantage. For organization-specific templates, we recommend the naming scheme `<adjective>-unito-thesis`. See the [submission guidelines](https://github.com/typst/packages?tab=readme-ov-file#submission-guidelines) for details. You can also take a look [at the existing templates](https://typst.app/universe/search?kind=templates) for inspiration.

packages/preview/unito-thesis/0.1.0/lib.typ
eduardz1: is it fine to set the default value to `none` in lib.typ and in the example leave the current path "imgs/logo.svg"? I want to leave the option to the users to use some of the variations to the logo that currently exist but still provide a default one

eduardz1:
Would `modern-unito-thesis` be considered a sufficiently creative? It replaces the old LaTeX one both with a more modern language and the updated logo :)

Otherwise I was thinking `opinionated-unito-thesis`

laurmaedje:
`modern-unito-thesis` is permissible

eduardz1:
I should've addressed everything

packages/preview/modern-unito-thesis/0.1.0/lib.typ
```diff
@@ -0,0 +1,326 @@
...
+      level: 2,
+      numbering: none,
+      outlined: false,
+      if lang == "en" {
+        "Acknowledgments"
+      } else {
+        "Ringraziamenti"
+      }
+    )
+    include "acknowledgments.typ"
```
laurmaedje, line 245: This here and the abstract below has the same problem where it's searching in the package rather than the project directory. You can reproduce the problem by storing the package as a [local package](https://github.com/typst/packages?tab=readme-ov-file#local-packages) and then scaffolding it with `typst init`.

If those things differ by user, you would need to pass them as a parameter (and a user can then write `acknowledgments: include "acknowledgments.typ"`.

If they don't differ, you can move them from the template directory into the package itself.

eduardz1: Sorry, I didn't notice, I'll probably delete them then and take their content as parameter, I'll check if there are no other similar problems

eduardz1:
should be ok now

laurmaedje:
Unfortunately, the package still doesn't compile because of a spelling mistake in a variable name, leading to the error `unexpected argument: acknoledgments`. 

The problem is that you are using three different spellings of the word "acknowledgments":
1. acknoledgments (wrong)
2. acknowledgments (correct, American English)
3. acknowledgements (correct, British English)

eduardz1:
Of course I fixed a typo on that terrible word before pushing and forgot to check that it still compiled, I double checked and it should be fixed for real now, I'm really sorry for all those issues

laurmaedje:
Don't worry about it. Thanks for the submission!

PR #460, opened 2024-03-21
bamdone-aiaa:0.1.0

isaacew:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a template for the AIAA conferences. With nearly 30,000 individual members from 91 countries, and 95 corporate members, AIAA is the world’s largest technical society dedicated to the global aerospace profession.


packages/preview/bamdone-aiaa/0.1.0/template/main.typ
```diff
@@ -0,0 +1,229 @@
...
+// in any manner whatsoever without further attribution or notice
+// to the creator. The author is not responsible for any liability 
+// from the usage or dissemination of this code.
+//
+// Author: Isaac Weintraub, Alexander Von Moll
+// Date: 06 NOV 2023
+// BAMDONE!
+//***************************************************************
+
+#import "@preview/bamdone-aiaa:0.1.0": aiaa
```
laurmaedje, line 16: There are multiple missing imports here, e.g. `nomenclature` and `dropcap`.

packages/preview/bamdone-aiaa/0.1.0/template/main.typ
```diff
@@ -0,0 +1,229 @@
...
+
+== Equations, Numbers, Symbols, and Abbreviations
+
+$ integral_0^r_2 &F(r,phi) upright(d) r upright(d) phi = [sigma r_2 \/ 2(mu_0)] \ dot&  integral_0^infinity  exp(-lambda | z_j - z_i | ) lambda^-1 J_1 (lambda r_2) J_0 (lambda e_i) upright(d) lambda $ <equation1>
+
+Be sure that the symbols in your equation are defined before the equation appears, or immediately following. Italicize symbols (T might refer to temperature, but T is the unit tesla). Refer to “Eq. (1),” not “(1)” or “equation (1)” except at the beginning of a sentence: “Equation (1) is…” Equations can be labeled other than “Eq.” should they represent inequalities, matrices, or boundary conditions. If what is represented is really more than one equation, the abbreviation “Eqs.” can be used. 
+
+An example of using Equations can be used as follows: It can be referred to as @equation1. Note that this will work when it's an in-sentence equation. However, when a sentence begins with Equation, you will need to use the following to make it work:
+
+#bEquation([@equation1]) when it starts at the beginning of a sentence.
```
laurmaedje, line 202: I recommend kebab-case as the most idiomatic Typst case, but it's not required.

packages/preview/bamdone-aiaa/0.1.0/template/main.typ
```diff
@@ -0,0 +1,229 @@
...
+// Author: Isaac Weintraub, Alexander Von Moll
+// Date: 06 NOV 2023
+// BAMDONE!
+//***************************************************************
+
+#import "@preview/bamdone-aiaa:0.1.0": aiaa
+
+#show: aiaa.with(
+    title: "Preparation of Papers for AIAA Technical Conferences",
+    bibliography-file: "refs.bib",
```
laurmaedje, line 20: This should use a `bibliography("...")` call, otherwise the file path will be resolved relative to the package directory, leading to a failure. You can take a look at the charged-ieee template to see how it's done there. The whole path handling will be improved in the future, but in the meantime, packages unfortunately have to work around it.

isaacew:
I wish I knew a way so as to make the nomenclature function and the b-Equation function part of the aiaa template rather than as separated imports. If there's a way to do this; I would like to know your thoughts about the right way to do that. If there isn't a way to do that; it's ok. I was just looking for a 'cleaner' way to import the aiaa package where users can just say import "bamdone-aiaa" and have all the functions they need.

Thank you for your help in this process. I am not an experienced git user.

packages/preview/bamdone-aiaa/0.1.0/template/main.typ
isaacew: It's now `b-equation()` rather than `bEquation` to conform to your suggestion. Thank you.

packages/preview/bamdone-aiaa/0.1.0/template/main.typ
isaacew: I have added the other imports as suggested. Thank you.

packages/preview/bamdone-aiaa/0.1.0/template/main.typ
isaacew: I was able to follow in the example from `charged-ieee`. Thank you for the suggestion!

laurmaedje:
> I wish I knew a way so as to make the nomenclature function and the b-Equation function part of the aiaa template rather than as separated imports. If there's a way to do this; I would like to know your thoughts about the right way to do that.

Aside from the `*`, there isn't another way.

laurmaedje:
Thanks!

PR #466, opened 2024-03-23
invoice-maker:1.1.0

ad-si:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:
Generate beautiful invoices from simple Typst or YAML data records.
Greatly reduces the necessary amount of work to create an invoice.


packages/preview/invoice-maker/1.0.0/lib.typ
```diff
@@ -0,0 +1,448 @@
...
+    font: "Arial",
+    fontsize: 11pt,
+    margin: (
+      top: 20mm,
+      right: 25mm,
+      bottom: 20mm,
+      left: 25mm
+    ),
+  ),
+  items: [],
```
laurmaedje, line 151:
```suggestion
  items: (),
```

packages/preview/invoice-maker/1.0.0/lib.typ
```diff
@@ -0,0 +1,448 @@
...
+      owner: "Inhaber",
+      iban: "IBAN",
+    ),
+  )
+
+#let invoice(
+  language: "en",
+  country: none,
+  title: none,
+  banner_image: none,
```
laurmaedje, line 131: I would recommend kebab-case as the canonical Typst case.

packages/preview/invoice-maker/1.0.0/template/main.typ
```diff
@@ -0,0 +1,75 @@
+#import "@preview/invoice-maker:1.0.0": *
+
+#show: invoice.with(
```
laurmaedje, line 3: Just a remark: If the invoice isn't really intended to have a body, you don't _have_ to do a `show: invoice.with(` rule. It could also just be `invoice(`. But perhaps it's also good to have the flexibility.

packages/preview/invoice-maker/1.0.0/lib.typ
```diff
@@ -0,0 +1,448 @@
...
+  )
+
+  v(1em)
+
+  if cancellation_id == none {
+    let due_date = if due_date != none { due_date }
+          else {
+            TODO
+            // TODO: Reactivate after Typst supports adding dates
+            // datetime.today().add(days: 14).
```
laurmaedje, line 409: you can add datetimes and durations: `datetime.today() + duration(days: 14)`

packages/preview/invoice-maker/1.0.0/lib.typ
```diff
@@ -0,0 +1,448 @@
...
+  #v(8mm)
+]
+
+#let signature_line = line(length: 5cm, stroke: 0.4pt)
+
+#let endnote(num, contents) = [
+  #stack(dir: ltr, spacing: 3pt, super[#num], contents)
+]
+
+#let languages = (
```
laurmaedje, line 62: Is there some way to customize those as a user or add a new language? I think that would be really useful.

packages/preview/invoice-maker/1.0.0/lib.typ
ad-si: Ok, will change it everywhere!
But honestly, I think it's not an ideal choice. Kebab-case isn't selectable with a double-click (at least not per default), which makes working with more cumbersome. Also interesting choice since you're coming from Rust. 😛

packages/preview/invoice-maker/1.0.0/template/main.typ
ad-si: Yeah, I think it's good to provide the template like this,
so that people can just type away if they need to add additional infos.

ad-si:
Thanks for your good ideas! All of them should be included now!

packages/preview/invoice-maker/1.0.0/lib.typ
laurmaedje: There is no great case choice for Typst unfortunately. All have downsides. Snake case in particular has overlap with emphasis syntax and math mode (though kebab case also struggles with the latter). Kebab-case imo looks nicer for CSS-style set rules and some consider it more friendly to beginners. There were a bunch of discussion about this on the Discord at various times.

packages/preview/invoice-maker/1.1.0/template/main.pdf
laurmaedje, line 1: This file should be removed.

packages/preview/invoice-maker/1.0.0/lib.typ
ad-si: Ok, I see.
Also, I realized double click behavior can be customized: https://stackoverflow.com/questions/34781471/vscode-is-there-a-way-to-customise-double-click-select-behaviour

Didn't have time to try it out yet, but if somebody figures out a good configuration for Typst, that would be a good addition to the docs!

laurmaedje:
Thanks!

PR #475, opened 2024-03-25
scholarly-tauthesis:0.4.0

SeSodesa:
I am submitting
- [x] a new package
- [ ] an update for a package

This template package defines the structure and look of Tampere University bachelor's and master's theses.


SeSodesa:
Ah, it looks like the tests have failed. Does the build process not currently support Git submodules?

SeSodesa:
Ah no, it seems the [package] section entrypoint needs to point to `template/main,typ` and not just `main.typ`.

SeSodesa:
Ok, so even after fixing the `package.entrypoint` a submodule version of the files is not accepted. Let my try one more time with a simple copy-pasta of the files into this repository.

SeSodesa:
Ok, now the tests pass.

packages/preview/tauthesis/0.3.2/typst.toml
```diff
@@ -0,0 +1,18 @@
...
+name = "tauthesis"
+version = "0.3.2"
+entrypoint = "template/main.typ"
+authors = ["Santtu Söderholm <santtu.soderholm@tuni.fi>"]
+license = "MIT"
+description = "A template for writing Tampere University theses."
+homepage = "https://gitlab.com/tuni-official/thesis-templates/tau-typst-thesis-template"
+repository = "https://gitlab.com/tuni-official/thesis-templates/tau-typst-thesis-template"
+keywords = ["Tampere University", "thesis template"]
+categories = ["thesis", "layout"]
```
reknih, line 11: I think that the `layout` category should be dropped

packages/preview/tauthesis/0.3.2/typst.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "tauthesis"
```
reknih, line 2: You'll need to change the name since template names must have at least one unique, creative component so that multiple people can create a template for the same instutition without one having an unfair advantage. For organization-specific templates, we recommend the naming scheme <adjective>-tau-thesis. See the [submission guidelines](https://github.com/typst/packages?tab=readme-ov-file#submission-guidelines) for details. You can also take a look [at the existing templates](https://typst.app/universe/search?kind=templates) for inspiration.

packages/preview/tauthesis/0.3.2/template/main.typ
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * to perform the compilation. If you are writing a multi-file
+ * project, this file is where you need to include your content
+ * files.
+ *
+***/
+
+//// Initialize document type.
+
+#import "tauthesis.typ"
```
reknih, line 15: Templates should import their styles through the package manager (i.e. `#import "@preview/tauthesis:0.3.2"`). As-is, the `tauthesis.typ` file is pasted into the user's project directory when they run `typst init` even though it explicitly warns the user to not edit it. Instead, you should move it to the package's root directory and make it the entrypoint of the package in the `typst.toml`. Imports in other files need to be changed, too.

reknih (commented):
This looks very nice! I added a few notices that need to be addressed before merging.

SeSodesa:
Ok, I'll have a look at these soon.

packages/preview/tauthesis/0.3.2/typst.toml
SeSodesa: May I leave the hyphen out from between `tau` and `thesis` and simply come up with a unique-ish prefix?

reknih:
Yes!

On 6. Apr 2024, at 07:31, Santtu Söderholm ***@***.***> wrote:

﻿

@SeSodesa commented on this pull request.

________________________________

In packages/preview/tauthesis/0.3.2/typst.toml<https://github.com/typst/packages/pull/475#discussion_r1554528233>:

> @@ -0,0 +1,18 @@
+[package]
+name = "tauthesis"


May I leave the hyphen out from between tau and thesis and simply come up with a unique-ish prefix?

packages/preview/tauthesis/0.3.2/template/main.typ
SeSodesa: Is there a way to test whether the template works after these changes locally? Is there a specific directory where I could put the `tauthesis` file on my system, where `typst` picks the packages from, or are they somehow baked into the binary?

SeSodesa: Or do I just push the changes here and the CI takes care of that?

SeSodesa: Ah, this seems to be covered [here][https://typst.app/blog/2023/package-manager/]:

> When the command line compiler encounters a package import, it searches for the directory `typst/packages/{namespace}/{name}-{version}` within the OS-dependant [data](https://docs.rs/dirs/latest/dirs/fn.data_dir.html) and [cache](https://docs.rs/dirs/latest/dirs/fn.cache_dir.html) directories.

SeSodesa: And here are the system-specific data directories: <https://docs.rs/dirs/latest/dirs/fn.data_dir.html>.

SeSodesa: Hmm, making a symbolic link `tauthesis-0.3.2 -> $HOME/tauthesis-repo/` in `$HOME/.local/share/typst/packages/preview/` results in the error
```sh
downloading @preview/tauthesis:0.3.2
error: package not found (searched for @preview/tauthesis:0.3.2)
   ┌─ template/main.typ:15:8
   │
15 │ #import "@preview/tauthesis:0.3.2"
```

SeSodesa: And this is after moving the `tauthesis.typ` file to project root and making it the entrypoint in `typst.toml`.

packages/preview/tauthesis/0.3.2/typst.toml
SeSodesa: This was resolved by 73ec8f60b994513da64f4fd1102c3d7d0e0361dd.

packages/preview/tauthesis/0.3.2/template/main.typ
SeSodesa: Ok, this was resolved by having a folder `$HOME/.local/share/typst/packages/preview/tauthesis/0.3.2`. So the hyphen between the package name and version is outdated information.

packages/preview/tauthesis/0.3.2/typst.toml
SeSodesa: The package has been renamed as `scholarly-tauthesis`.

SeSodesa:
I believe this might be ready now. The local version works and the packaging tests pass here as well.

laurmaedje:
Thanks!

I didn't want to block merging on this, but for the future: Not that `#include("content/01.typ")` is canonically written as `#include "content/01.typ"` as its distinct syntax and not a function. The parenthesis just happen to be accepted.

SeSodesa:
> Thanks!
> 
> I didn't want to block merging on this, but for the future: Not that `#include("content/01.typ")` is canonically written as `#include "content/01.typ"` as its distinct syntax and not a function. The parenthesis just happen to be accepted.

I guess the same applies to imports. I'll change this in a future release.

PR #478, opened 2024-03-25
ilm:0.1.2

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

---

### Fixed

- Check if a heading's body has a text field before using it (https://github.com/talal/ilm/commit/d6eb41cabc587c7135baccc09f3a6400adbb0b84)

---


reknih:
Thank you!

PR #484, opened 2024-03-29
modern-russian-dissertation:0.0.1

SergeyGorchakov:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Russian PhD thesis.


laurmaedje (commented):
Currently, all of the implementation of this template is in the `template` directory which will be copied into the user's project. The intented way for templates to work, in contrast, is to encapsulate the implementation in the package itself and only provide the scaffolding the user will adjust in the `template` directory. That way it's simpler for users to see what's what and also to upgrade to future versions. You can check out some of the other templates to see how it works.

That aside, the name is unfortunately a bit too descriptive. While it's of course not an institution, the disambiguation `russian` just by itself is too generic. It would be great if you pick a somewhat more unique name! See also the [list of other templates](https://typst.app/universe/search?kind=templates) for inspiration.

SergeyGorchakov:
Thank you for your additions. I fixed the template

laurmaedje:
If I'm not mistaken GOST is a Russian standards body, which makes this name unfortunately also not permissible, as it would be akin to naming something e.g. `iso-paper`. The adjective should be not directly descriptive or affiliated with some standard / organization / similar thing. Other than that, you can pick anything, can be fun / a pun, or really just anything.

The naming policy might seem and be a bit strange, but we think it's in the interest of everybody that it's clear what is and isn't affiliated with an official entity!

SergeyGorchakov:
Thank you. 

How about the name: “gost-7.0.11-phd-thesis”?

The fact is that in Russia there is a standard for the preparation of candidate and doctoral dissertations. Therefore, I would like the user to see that the presented template and its improvements exactly meet the requirements of the specified standard. Briefly about the standard:

_Area of ​​use_

_This standard establishes general requirements for the preparation of candidate and doctoral dissertations and abstracts of dissertations in all fields of knowledge._

laurmaedje:
> How about the name: “gost-7.0.11-phd-thesis”?

That doesn't really fix the problem. It makes it _even more_ descriptive. The reason for the naming rule is that other people can also submit a package for the task. 

If you look at the existing templates, you'll note that they all have some component that is not descriptive at all (e.g. "charged" in `charged-ieee`, which is just a pun of it being an electronics conference). There's also various university templates there that also have fixed guidelines, but still each has a unique naming component, so that multiple people from that university could submit a package without an unfair naming advantage for one.

You can name it `<something>-gost-phd-thesis` where `<something>` is _not_ anything descriptive or also just `<something>-phd-thesis` where `<something>` could somehow hint on what it's about  (e.g. some Russian concept or something like that) it without being directly descriptive.

As to informing the user that is adheres to the standard: The ideal place to do that is in the package description.

SergeyGorchakov:
Okay, thank you. Your policy is now clearer. How about the name: “modern-russian-dissertation”?

laurmaedje:
That works.

SergeyGorchakov:
I corrected the template name

laurmaedje:
Thanks!

PR #492, opened 2024-04-01
linguify:0.4.0

jomaway:
I am submitting
- [x] an update for a package

Description: Simplify API and add [fluent](https://projectfluent.org/) support


packages/preview/linguify/0.4.0/lib/linguify.typ
```diff
@@ -0,0 +1,163 @@
...
+      // let lang_section = database.at(key, default: none)
+      if key not in _database.keys() {
+        _database.insert(key, value)
+      } else {
+        let new = _database.at(key) + value
+        _database.insert(key, new)
+      }
+    }
+    database.update(_database);
+  }
```
laurmaedje, line 51: Didn't see this before, but I recommend using an update closure instead of get() + update(). Calling `update_database` more than five times will likely lead to a convergence error.

packages/preview/linguify/0.4.0/lib/linguify.typ
```diff
@@ -0,0 +1,163 @@
...
+#let database = state("linguify-database", none)
+
+
+/// Set the default linguify database
+///
+/// The data must contain at least a lang section like described at @@database.
+///
+/// - data (dictionary): the database which will be set to @@database
+/// -> content (state-update)
+#let set_database(data) = {
```
laurmaedje, line 23: Also a side note: kebab-case would be a bit more idiomatic.

packages/preview/linguify/0.4.0/lib/linguify_fluent_rs/linguify_fluent_rs.wasm
laurmaedje, line 1: this _seems_ pretty big for just fluent but maybe there is not much that can be done about it

packages/preview/linguify/0.4.0/README.md
```diff
@@ -0,0 +1,59 @@
...
+Example:
+```typst
+#import "@preview/linguify:0.4.0": *
+
+#let lang_data = toml("lang.toml")
+#linguify_set_database(lang_data);
+
+#set text(lang: "de")
+
+#linguify(abstract)  // Shows "Zusammenfassung" in the document.
```
laurmaedje, line 22:
```suggestion
#linguify("abstract")  // Shows "Zusammenfassung" in the document.
```

packages/preview/linguify/0.4.0/lib/linguify_fluent_rs/linguify_fluent_rs.wasm
jomaway: Actually i don't know as this was a contribution from someone else. And i don't have any experience with wasm. The source code is [here](https://github.com/jomaway/typst-linguify/tree/main/lib/linguify_fluent_rs).

packages/preview/linguify/0.4.0/lib/linguify.typ
jomaway: Ok thanks for the info, i will change this later. ;)

packages/preview/linguify/0.4.0/lib/linguify_fluent_rs/linguify_fluent_rs.wasm
laurmaedje: `wasm-opt -O3 linguify_fluent_rs.wasm -o opt.wasm` brings it down to 1.4MB, but the order of magnitude doesn't change. The wasm file seems to mostly consist of code rather than data, so I guess fluent is larger than I thought.

packages/preview/linguify/0.4.0/lib/linguify.typ
jomaway: Decided to remove this function, as it does not fit for all current usecases anymore.

packages/preview/linguify/0.4.0/lib/linguify_fluent_rs/linguify_fluent_rs.wasm
jomaway: Do you want me to change this? Or can we leave it as is?

laurmaedje: You can leave it

PR #493, opened 2024-04-01
starter-journal-article:0.2.0

HPDell:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Make the template configurable via the argument `template`. Users can set self-defined templates for title, author list, institutions, abstract, and main text in this argument to override the default template.


packages/preview/starter-journal-article/0.2.0/README.md
```diff
@@ -0,0 +1,206 @@
+# A configurable starter template for journal articles
+
+This package provides a template for writing journal articles
+to organise authors, institutions, and information of corresponding authors.
+
+![](./thumbnail.png)
```
laurmaedje, line 6: I recommend not adding the thumbnail here as Typst Universe will show a click-to-enlarge thumbnail anyway.

packages/preview/starter-journal-article/0.2.0/template/main.typ
```diff
@@ -0,0 +1,38 @@
+#import "@preview/starter-journal-article:0.2.0": article, author-meta
+
+#show: article.with(
+    title: "Article Title",
```
laurmaedje, line 4: There's one unnecessary extra level of indent here.

packages/preview/starter-journal-article/0.2.0/README.md
HPDell: Thanks. This has been removed in cee5160deee5f229c2755c49dc4fa39cf62ae991 .

packages/preview/starter-journal-article/0.2.0/template/main.typ
HPDell: Thanks. The extra indent has been removed in cee5160deee5f229c2755c49dc4fa39cf62ae991 .

PR #503, opened 2024-04-02
`modern-cv:0.2.0`

DeveloperPaul123:
I am submitting
- [ ] a new package
- [x] an update for a package

Improved the readability of the template and cleaned up the design to make it more readable and flexible. 


packages/preview/modern-cv/0.2.0/typst.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.2.0"
+compiler = "0.11.0"
+entrypoint = "lib.typ"
+authors = ["Paul Tsouchlos <https://github.com/DeveloperPaul123>"]
+license = "MIT"
+description = "A modern resume template based on the Awesome-CV Latex template."
+keywords = ["CV", "Curriculum Vitae", "Resume", "Minimalistic", "Professional", "Modern"]
+categories = ["cv"]
+disciplines = []
+exclude = ["resume-documentation*", "scripts*"]
```
laurmaedje, line 12: The only assets which seems to be used by the package itself is `square-phone-solid.svg`. The image assets for the README should thus be excluded here to keep the package size down.

packages/preview/modern-cv/0.2.0/template/resume.typ
```diff
@@ -0,0 +1,95 @@
...
+      email: "js@example.com", 
+      phone: "(+1) 111-111-1111",
+      github: "DeveloperPaul123",
+      linkedin: "Example",
+      address: "111 Example St. Example City, EX 11111",
+      positions: (
+        "Software Engineer",
+        "Software Architect",
+        "Developer"
+      )
```
laurmaedje, line 16: There are some double indents here in various places, which might you want to remove.

packages/preview/modern-cv/0.2.0/template/resume.typ
```diff
@@ -0,0 +1,95 @@
...
+  ),
+  date: datetime.today().display(),
+  colored_headers: true
```
laurmaedje, line 19: The idiomatic Typst case would be kebab-case, in case you want to change it.

packages/preview/modern-cv/0.2.0/template/coverletter.pdf
laurmaedje, line 1: These PDFs should be removed.

packages/preview/modern-cv/0.2.0/template/resume.typ
DeveloperPaul123: Thanks! I will address this in a future release.

DeveloperPaul123:
Thanks for the review @laurmaedje, I think I addressed all your points.

PR #516, opened 2024-04-08
modern-nju-thesis:0.3.1

OrangeX4:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: excluding the images directory in typst.toml will cause the images directory in the template directory to be excluded as well, resulting in initialization failures. So I changed one of the images directories to `imgs`.


laurmaedje:
Ah, that's unfortunate and also hard/impossible to catch with local testing.

OrangeX4:
Yes, it's hard to notice i guess.

PR #523, opened 2024-04-09
typpuccino:0.1.0

TeddyHuang-00:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This simply adds four palette of Catppuccin for use in Typst.


laurmaedje:
Since this is a port of an existing project, the name needs to have some unique component beyond the original name, so that another person could potentially also create a port of it. (At least unless you are a maintainer of the upstream project.) This is similar to how we handle plugins that provide bindings to other libraries.

TeddyHuang-00:
> Since this is a port of an existing project, the name needs to have some unique component beyond the original name

Now renamed to ca**typ**uccin, which is close to original while indicating this is for Typst. Is that okay now?

laurmaedje:
It's maybe a _bit_ close so that one could possibly not notice the change, but I think it's okay.

TeddyHuang-00:
How about Typpuccino? Is it better than this one?

laurmaedje:
That would definitely also work. I'll leave it up to you!

TeddyHuang-00:
Renamed to Typpuccino then. Anything else missing here?

laurmaedje:
Nope, all good. Thanks!

PR #529, opened 2024-04-11
outline-summaryst:0.1.0

aarneng:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: `outline-summaryst` is a basic template designed for including a summary for each entry in the table of contents, particularly useful for writing books. It provides a simple structure for organizing content and generating formatted documents with summary sections.


laurmaedje:
From the description, it seems to me that this is more a building block rather than a complete template. Would it perhaps make more sense for it be just a package, not a template package?

aarneng:
> From the description, it seems to me that this is more a building block rather than a complete template. Would it perhaps make more sense for it be just a package, not a template package?

I'm honestly not sure, I guess it's currently kind of in the middle of the two. The main idea is that it implements a way to add a summary into the `#outline()`,  but also changes the styling of said outline and each heading. Here is a link to a version running on the web app, if you want to quickly look through it yourself: https://typst.app/project/rREF3kavhUHhGJqNFgHETM

If you feel that the current state makes it more of a package than a template, I can make changes re-submit it as that. If so, do you have any pointers on how to make it more template-like for future versions?

laurmaedje:
Conceptually, it does feel more like a package to me. Of course, right now some of it is set up like a template, which would need to be changed if it were just a package. 

> If so, do you have any pointers on how to make it more template-like for future versions?

Is that desirable? Being a template mostly means that it's not mix-and-match with other templates. If it's a package that can be flexibly used alongside other stuff, great!

But ultimately, it's totally up to you whether you want to go with this, those are just my 2ct.

aarneng:
I think that's fair, i've changed this and removed the template part. I hadn't considered that making it a template limits usability to a degree, that is nice feedback :)

packages/preview/outline-summaryst/0.1.0/src/outline-summaryst.typ
```diff
@@ -0,0 +1,123 @@
+#import "helpers.typ": calc-elem-size, make-title, make-foreword
+
+// randomly generated IDs
+#let gs = state("VIvouA", ())
+#let elemcounter = counter("k1lRPA")
+
+#let supl = [Custom Heading]
+
+
+#let outline-summaryst(
```
laurmaedje, line 10: Is this function needed at all? Wouldn't it make more sense to provide a function that generates just the fancy outline instead? 

Since your headings use a special `make-heading` function instead of the syntax, the heading show rule could also be included with that. That way it would be even more composable.

aarneng: I've changed the behavior of the both functions to be more composable based on your feedback (note that I renamed `outline-summaryst` to `style-outline`).

To answer your question of why `outline-summaryst` is not a function that generates the fancy outline, this is because some templates include an `#outline()` by default, so this package could not be used alongside such templates. The current behavior, where we have a show rule for `outline.entry`, solves this issue

laurmaedje:
Thanks!

PR #559, opened 2024-04-18
chore: add schema for typst.toml

sjfhsjfh:
Maybe create a schema also [here](https://www.schemastore.org)?


elegaanz:
From what I understand, it would be better to contribute this file to the schemastore repository directly, so that VSCode 
extensions (and maybe other tools) pick it up automatically, without needing to point it to the self-hosted file.

I think the schema should follow what universe does, not what the local compiler does, if we consider that this schema is useful to get editor completions and diagnostics: I think that most of the time when you edit a `typst.toml` it will be submitted to Universe at some point. Maybe another schema file with only compiler-required fields marked as such could still exist alongside the main one, to make it possible to run lighter checks by giving an explicit schema path (I believe this is something you can do with most tools that work with JSON schemas).

If you open a PR on the schemastore repository, you can ping me in the discussion over there, I will happily give it a look.

PR #561, opened 2024-04-18
smooth-tmlr:0.4.0

daskol:
I am submitting
- [ ] a new package
- [x] an update for a package

**Description:** Fix caption alignment of figures (https://github.com/daskol/typst-templates/issues/6).


packages/preview/smooth-tmlr/0.4.0/template/main.pdf
laurmaedje, line 1: This file should be deleted.

daskol: Fixed. Sorry for that. What do you think about adding `*.pdf` to `.gitignore` in repo root?

laurmaedje: Some PRs push PDF manuals, which they link to, which is okay. It just needs to be excluded at the package level. But we'll add more automated checks for stuff like this soon.

PR #574, opened 2024-04-20
isc-hei-report:0.1.0

pmudry:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: This package is an official template for writing reports for the ISC bachelor degree programme at the School of Engineering in Sion, Switzerland.


laurmaedje:
There are by now various templates starting with "modern-". We think that the profileration of this term isn't great for users and not really in the spirit of the naming rules. While it's not forbidden, we strongly discourage its use. Instead, we recommend you choose an adjective that more closely relates to your template.

pmudry:
Would something like "official-" be more appropriate in your opinion (as it is an official template for the school) ?

laurmaedje:
If you have an official role in that university and the template is thus truly official, then a descriptive name would be allowed. Then, we would need to verify that. If you're a student and it's not truly official, but rather just adhering to the official guidelines, then the prefix "official-" would not be allowed, as it'd be misleading. In this case, I would suggest just picking some word that fits well with your university. Since it is an engineering school, it might be something engineering related, or really just any other interesting thing you can come up with.

pmudry:
Yes the template is truly official as I am the head of the CS department and I plan to release templates for reports and thesis in my university. What do you need to proceed ? I have official websites that I can send you as well as as official emails.

laurmaedje:
Okay, that changes things. 

If ISC is a term exclusive to your university, you can just call it `isc-report`. If it's a common French abbrevation, then we'd encourage you to include the abbrevation of your university in the name as well.

Regarding verification, you could just send us an email from your official address to hello@typst.app.

laurmaedje:
I got the mail, thank you! It's great to have someone submitting an official template! :)

I think `isc-hei-report` was ideal, as it includes the university name, whereas `isc-engineering-report` is more general again. Note, that the folder name also needs to change for the package to be functional (and CI to pass) again.

pmudry:
Fine by me, renamed to `isc-hei-report` and directory name changed as well. Thank you for the comments !

packages/preview/isc-hei-report/0.1.0/template/report.typ
```diff
@@ -0,0 +1,223 @@
+#import "@preview/modern-isc-report:0.1.0" : *
```
laurmaedje, line 1: There are still a few places where it says `modern-isc-report` (also in the README). Here, in particular, it prevents the template from working. Once changed here, it works on my system.

pmudry: This should now be resolved. Sorry for the mess, everything has now been double checked and should be working as expected.

PR #536, opened 2024-04-12
iconic-salmon-svg:1.0.0

Bi0T1N:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Easily reference your social media profiles with the typical icon of the service plus a link to your profile.

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name that isn't the most obvious or canonical name for what the package does
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] `exclude`d PDFs or README images, if any, but not the LICENSE

- [ ] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #218, opened 2023-11-04
socialhub-fa:1.0.0

Bi0T1N:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:
The `socialhub-fa` package is designed to help you create your curriculum vitae (CV). It allows you to easily reference your social media profiles with the typical icon of the service plus a link to your profile.

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name in conformance with the guidelines
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] ensured that my submission does not infringe upon the rights of a third party
- [x] tested my package locally on my system and it worked
- [x] named this PR as `name:version` of the submitted package
- [x] agree that my package will not be removed without good reason

PR #575, opened 2024-04-20
iconic-salmon-fa:1.0.0

Bi0T1N:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is the renamed version of the `socialhub-fa` (#218) package with the new name introduced in #536. It also fixes a small inconsistency in the generated example PDF (different icon for GitLab due to initially wrong build environment).


Bi0T1N:
Is there any special process to remove or deprecate the existing `socialhub-fa` package?

packages/preview/iconic-salmon-fa/1.0.0/examples/examples.typ
```diff
@@ -0,0 +1,104 @@
+#import "../iconic-salmon-fa.typ": *
```
elegaanz, line 1: This import should use the package specification:
```suggestion
#import "@preview/iconic-salmon-fa:1.0.0": *
```

packages/preview/iconic-salmon-fa/1.0.0/examples/minimal_example.typ
```diff
@@ -0,0 +1,4 @@
+// #import "@preview/iconic-salmon-fa:1.0.0": github-info, gitlab-info
+#import "../iconic-salmon-fa.typ": github-info, gitlab-info
```
elegaanz, line 2: Same here:
```suggestion
#import "@preview/iconic-salmon-fa:1.0.0": github-info, gitlab-info
```

elegaanz (commented):
Just two import statements that should be changed, but other than that it looks good. Thanks!

elegaanz:
> Is there a process defined on how to remove or deprecate the existing `socialhub-fa` package?

What you could do is to publish a new minor version of `socialhub-fa`, that only edits the contents of the README.md to tell to use this package instead.
There isn't a proper way to deprecate packages for the moment, but once it will be implemented we will notify you.

packages/preview/iconic-salmon-fa/1.0.0/examples/examples.typ
Bi0T1N: Is there any hard requirement for this? I see that several other packages also reference the local version (which btw is the only way to access a new version to generate a new PDF file before publishing)

elegaanz: No, it is not a hard requirement. It is just better because then people can copy/paste your example and start from here. It also makes it easier to tell which version of the package is used if the file is given out-of-context.
I agree that this is however not ideal to generate PDF files. The best solution as far as I know is to generate them with the relative import, and then update it to use the package import before submitting here.

Bi0T1N: That's why I have included the instructions in the `README.md` :smile:

elegaanz: Ok, I'll merge it as is then.

PR #578, opened 2024-04-21
fletcher:0.4.4

Jollywatt:
I am submitting
- [ ] a new package
- [x] an update for a package

Features since v0.4.3:
- Add node groups via the `enclose` option of `node()`.
- Node labels can be aligned inside the node with `align()`.
- Node labels wrap naturally when label text is wider than the node. **Note:** This may change diagram layout from previous versions.
- Add a `layer` option to nodes and edges to control drawing order.
- Add node shapes: `ellipse`, `octagon`.


Jollywatt:
Sorry, I meant to exclude the `tests/` directory and other development files. Will fix later...

PR #580, opened 2024-04-22
guided-resume-starter-cgc:2.0.0

Spelkington:
I am submitting
- [x] a new package
- [ ] an update for a package

---

After [porting a previous template over to Typst](https://typst.app/project/rGgmv6JmmpPSDFdMwsHWfe), I was sent the recommendation to add it to Typst Universe. The original has now been proofread and typos appropriately swept away — after an additional proofread and local test on my end, this appears to be ready to go!

This is a port of a previous LaTeX package used by undergraduates for creating new resumes. On top of a standard styling, it also includes a set of best practices on how to write good content for a resume, such as:

- Structuring concise bullet points
- Formatting non-standard experience as an addition to work experience
- Methods for tuning resumes to particular job entries

Please let me know if there are any necessary additions needed. Thanks!

---


packages/preview/guided-resume-starter-cgc/2.0.0/template/templates/resume.template.typ
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  // Main body.
+  set par(justify: true)
+
+  body
+}
+
+// Allows hiding or showing full resume dynamically using global variable
+#let hide(shouldHide, content) = {
```
elegaanz, line 58: Typst packages generally use kebab-case for naming functions and their arguments. But this is not a hard requirement, and you can decide to leave this as is if you prefer.
```suggestion
#let hide(should-hide, content) = {
```

packages/preview/guided-resume-starter-cgc/2.0.0/template/templates/resume.template.typ
```diff
@@ -0,0 +1,148 @@
+#let resume(
```
elegaanz, line 1: Unless you expect people to edit the contents of these functions, I think this file should be outside of the `template/` folder (either directly in the root of the package, or in a dedicated sub-folder). As it is done now, this template file would be duplicated in every new project initialized from this template.

packages/preview/guided-resume-starter-cgc/2.0.0/template/templates/resume.template.typ
Spelkington: This should be fixed in [`49d62cd`](https://github.com/typst/packages/pull/580/commits/f9d62cdadd5ab001bb625970125161d153255840)!

packages/preview/guided-resume-starter-cgc/2.0.0/template/templates/resume.template.typ
Spelkington: This project is a bit more "to each their own" than something like IEEE-compliant formatting, so the hope was to include the template file directly to more easily allow users to tweak it to their comfort. However, I'd imagine that Typst is looking to encourage adoption for the packaging system.

As a comfortable compromise, in [`49d62cd`](https://github.com/typst/packages/pull/580/commits/f9d62cdadd5ab001bb625970125161d153255840) I've kept the local copy of the template in the repository, switched the `#import` to the package system copy, and provided a brief instruction for users to use the local copy to tweak the editing if they'd like.

I've also added additional comments to the template file itself to make it a bit friendlier to power users who do want to tinker around with the template directly

Hopefully this is a comfortable in-between!

elegaanz: It is, thank you :)

PR #589, opened 2024-04-24
fractusist:0.1.0

liuguangxi:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: Create a variety of wonderful fractals (dragon curve, Koch snowflake, etc.) in Typst.


packages/preview/fractusist/0.1.0/examples/dragon-curve-n12.typ
```diff
@@ -0,0 +1,18 @@
+#set document(date: none)
+
+
+#import "/src/lib.typ": *
```
elegaanz, line 4: I think this example should use the package import.
```suggestion
#import "@preview/fractusist:0.1.0": *
```

elegaanz (commented):
I left a small remark, but otherwise it looks good to me.
Also, could you remove the compiled PDF files, please? Even if they wouldn't be included in the final archive, they still take space in the git repository.

packages/preview/fractusist/0.1.0/examples/dragon-curve-n12.typ
liuguangxi: I've modified it as suggested.

PR #601, opened 2024-04-28
numblex:0.1.0

sjfhsjfh:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Allow users to customize the numbering function more conveniently


packages/preview/numblex/0.1.0/lib/circle_numbers.typ
Enter-tainer, line 1: it would be better if you can pr this into typst/typst

packages/preview/numblex/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "numblex"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["ParaN3xus@https://github.com/ParaN3xus", "sjfhsjfh@https://github.com/sjfhsjfh"]
```
elegaanz, line 5: I think this format is not valid. You can either use:
- just a name: `ParaN3xus`
- a name and an email address: `ParaN3xus <you@example.org>`
- a name and a homepage URL: `ParaN3xus <https://example.org>`
- a name and a GitHub username: `ParaN3xus <@ParaN3xus>` (which is probably the one you are looking for in that case)

elegaanz (commented):
Thank you for your submission! It just needs a small fix in the manifest, and it will be good to merge I think.

sjfhsjfh:
Done

elegaanz:
Thank you!

PR #607, opened 2024-04-29
scholarly-epfl-thesis:0.1.0

augustebaum:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A template for a thesis at EPFL.


laurmaedje:
There are by now various templates starting with "modern-". We think that the profileration of this term isn't great for users and not really in the spirit of the naming rules. While it's not forbidden, we strongly discourage its use. Instead, we recommend you choose an adjective that more closely relates to your template.

augustebaum:
> ... we recommend you choose an adjective that more closely relates to your template.

How about "scholarly"?

laurmaedje:
That would work.

packages/preview/scholarly-epfl-thesis/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+compiler = "0.11.0"
+entrypoint = "src/lib.typ"
+repository = "https://github.com/augustebaum/epfl-thesis-typst"
+authors = ["Auguste Baum <https://github.com/augustebaum>"]
+license = "MIT"
+description = "A template for a thesis at EPFL"
+keywords = ["thesis", "university", "EPFL", "Lausanne"]
+categories = ["thesis"]
+exclude = ["example", "latex-reference", "screenshots"]
```
elegaanz, line 12: I think `example` should not be excluded here, as it is the directory that is used to initialize the template.

augustebaum: Thanks for your review! I misunderstood the point of `exclude` I guess. This is now corrected.

elegaanz: `exclude` tells which files should not be in the archive that will be downloaded by Typst when your package is imported. It is generally a good practice to exclude example files, but in your case you also use this directory to initialize the template associated with your package, which would not work if the directory is not part of the downloaded archive.

PR #609, opened 2024-05-01
delimitizer: 0.1.0

Enter-tainer:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This packages defines shorthand for `\big`, `\Big` ... and provide correct size for them by researching obscure LaTeX macro definition.


Enivex:
I think the better way to approach this is what mathtools’ `\DeclarePairedDelimiter` does. There's for instance no need to write `\big` twice

Enter-tainer:
> I think the better way to approach this is what mathtools’ `\DeclarePairedDelimiter` does. There's for instance no need to write `\big` twice

good idea. I've add `paired-delimiter` for this. Now the demo looks like this

```typst
#let Bigp = paired-delimiter(Big[(], Big[)])
#let bigp = paired-delimiter(big[(], big[)])

$
Bigp(
  bigp((a+b)times (a-b))
div
  bigp((c+d)times (c-d))
) + d \ = (a^2-b^2) / (c^2-d^2)+d
$
```

Enivex:
> > I think the better way to approach this is what mathtools’ `\DeclarePairedDelimiter` does. There's for instance no need to write `\big` twice
> 
> good idea. I've add `paired-delimiter` for this. Now the demo looks like this
> 
> ```typst
> #let Bigp = paired-delimiter(Big[(], Big[)])
> #let bigp = paired-delimiter(big[(], big[)])
> 
> $
> Bigp(
>   bigp((a+b)times (a-b))
> div
>   bigp((c+d)times (c-d))
> ) + d \ = (a^2-b^2) / (c^2-d^2)+d
> $
> ```

The actual benefit of `DeclarePairedDelimiter` is the ability to get all of them with with a single definition. Let's say I've defined `\parn` to be the pair (). The `\parn{1+1}` would give me unscaled parentheses, `\parn*{1+1}` would give me auto-scaled parentheses, and `\parn[\Big]{1+1}` would give manually scaled parentheses

Enivex:
> > The actual benefit of DeclarePairedDelimiter
> 
> I didn't use DeclarePairedDelimiter before. Is there a place i can learn how exactly it works?(and maybe all its benefits?)

http://mirrors.ctan.org/macros/latex/contrib/mathtools/mathtools.pdf

Enter-tainer:
`\parn[\Big]{1+1}` I'm a little bit worried about the API if we port it to typst. It will either become `$parn(Big, 1+1)$ ` or `$parn(Big)[$1+1$]$`. Both of them don't look clean to me

Enivex:
> `\parn[\Big]{1+1}` I'm a little bit worried about the API if we port it to typst. It will either become `$parn(Big, 1+1)$ ` or `$parn(Big)[$1+1$]$`. Both of them don't look clean to me

The first one doesn't look too bad to me. But it's worth considering whether typst syntax could be changed to allow optional arguments like this in a pleasant way than currently. It would be very useful in many contexts.

Enter-tainer:
> `$parn(Big, 1+1)$`

I just tried this and it doesn't work well. It seems that within the function call, typst switch to code mode and you can no longer use symbols. I decide to keep current API if there is not a good way to provide a clean interface

Enivex:
> > `$parn(Big, 1+1)$`
> 
> I just tried this and it doesn't work well. It seems that within the function call, typst switch to code mode and you can no longer use symbols. I decide to keep current API if there is not a good way to provide a clean interface

```
#let big = 1.2em
#let Big = 1.5em
#let bigg = 2em
#let Bigg = 2.5em

#let parn(size,expr) = {
  $lr((#expr), size: #size)$
}

$ parn(bigg,integral_0^oo e^(-x) d x) $
```

This seems to work fine, though ideally it should be possible to get both unscaled (by not providing two inputs) and auto scaled.

I'm not sure the sizes are correct. In your code you use `%`, but that will still scale based on the expression contained inside the delimiters @Enter-tainer

Enter-tainer:
> I just tried this and it doesn't work well. It seems that within the function call, typst switch to code mode and you can no longer use symbols. 

I got this wrong. it indeed works.

> In your code you use %, but that will still scale based on the expression contained inside the delimiters

good catch

PR #611, opened 2024-05-02
fletcher:0.4.4

Jollywatt:
I am submitting
- [ ] a new package
- [x] an update for a package

A featureful update to fletcher, which in particular adds a user-extendable arrowheads/marks system and the ability to easily enclose nodes in groups.

Changes since v0.4.3:

- Support fully customisable marks/arrowheads!
  - Added new mark styles and tweaked some existing defaults. **Note.** This may change the micro-typography of diagrams from previous versions.
- Add node groups via the `enclose` option of `node()`.
- Node labels can be aligned inside the node with `align()`.
- Node labels wrap naturally when label text is wider than the node. **Note:** This may change diagram layout from previous versions.
- Add a `layer` option to nodes and edges to control drawing order.
- Add node shapes: `ellipse`, `octagon`.


packages/preview/fletcher/0.4.4/docs/gallery/algebra-cube.typ
```diff
@@ -0,0 +1,47 @@
+#import "@preview/fletcher:0.4.2" as fletcher: diagram, node, edge
```
elegaanz, line 1: There are a few imports in the examples that are still using some older versions.

elegaanz (commented):
Thank you :) just a small remark about outdated imports in your examples, that you may want to fix. But it is not mandatory to get this merged, I just thought that maybe you would like to update them.

packages/preview/fletcher/0.4.4/docs/gallery/algebra-cube.typ
Jollywatt: Thanks! I've just fixed these.

elegaanz:
Thank you!

PR #615, opened 2024-05-04
ilm:1.1.1

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

### Added

- Added: Language setting definition in template. https://github.com/talal/ilm/commit/bdfd05935e42054cf313dfaa8f8728cd62f7b0d7

### Fixed

- Removed hardcoded heading for table of contents so that the heading is defined as per
  Typst's language settings. https://github.com/talal/ilm/commit/4316bd0af72eac3e1727159f63dabfa2e7fb9ee9

---


elegaanz:
Thanks!

PR #622, opened 2024-05-06
fix typo in zen-zine README example

tomeichlersmith:
A simple one-character typo which may confuse some users is removed. I've resolved this issue in the source repository as well, does this warrant a new release of the package? Or since there aren't any code changes its okay?

I am submitting
- [ ] a new package
- [x] an update for a package

Description: `zen-zine` is a simple package for typsetting 8-page zines onto U.S. Letter paper.


elegaanz:
Thank you!

PR #631, opened 2024-05-10
dotenv: v0.1.0 (new)

chillcicada:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Load environment variables from a .env file.


packages/preview/dotenv/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "dotenv"
+version = "0.1.0"
+entrypoint = "dotenv.typ"
+authors = ["chillcicada"]
+license = "MIT"
+description = "Load environment variables from a .env file."
+homepage = "https://github.com/chillcicada/typst-dotenv#readme"
+repository = "https://github.com/chillcicada/typst-dotenv"
```
elegaanz, line 9: These two links feel redundant, I think having only `repository` would be less confusing.

elegaanz (commented):
Thank you for your submission :) I have two remarks:
- first, the name is too descriptive of what the package does (see the ["Naming" section here](https://github.com/typst/packages/?tab=readme-ov-file#submission-guidelines). Could you please come up with an alternative name? It can just be an adjective or a suffix/prefix added to `dotenv`.
- I also think that generated example files should not be included in this repository, as they can easily be re-created if needed. Could you please delete them? (this remark does not apply to `/example.png` as it is needed for your README to render correctly on Typst Universe).
Other than that and the comment below, it should be good to merge.

elegaanz:
Thanks

PR #635, opened 2024-05-10
chuli-cv:0.1.0

npujol:
I am submitting
- [x] a new package
- [ ] an update for a package


packages/preview/chuli-cv/0.1.0/template/cv.typ
```diff
@@ -0,0 +1,174 @@
...
+      text: [Faye Valentine],
+      link: "https://linkedin.com/in/npujolm/"
+    ),
+    (
+      icon: icons.homepage,
+      text: [Earth],
+      link: "#"
+    ),
+  ),
+  profile-picture: image("media/avatar.png")
```
elegaanz, line 53: The `media` directory won't be copied to projects using this template, so this wouldn't work. If you want to include an example image, it should go in the `template` directory (it can be in a dedicated sub-directory if you want).

packages/preview/chuli-cv/0.1.0/template/letter.typ
```diff
@@ -0,0 +1,27 @@
...
+
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+
+Sincerely,
+#letter_footer(  
+  name: "Faye Valentine",
+)
+#letter_signature(image("/media/signature.jpeg", width: 25%))
```
elegaanz, line 27: Same remark here.

packages/preview/chuli-cv/0.1.0/lib.typ
```diff
@@ -0,0 +1,259 @@
...
+#import "modules/languages.typ": *
+#import "modules/piechart.typ": *
+#import "modules/education.typ": *
+#import "modules/entry.typ": *
+
+
+// Cover letter Components
+//--------------------------------------------
+
+#let letter_header(
```
elegaanz, line 15: It is recommended to use kebab-case names. For instance, it would be `letter-header` here, but there are other places in the package where snake_case or camelCase is being used. This is however not mandatory, and if you prefer we can merge the package as it is.

elegaanz (commented):
I think the example images that you used are subject to copyright, so I would rather not include them in this repository. Could you please replace them with some images you made yourself or ones that are under a free license?

packages/preview/chuli-cv/0.1.0/lib.typ
npujol: Thanks for the feedback! I've updated the commit to use kebab-case naming convention, as you recommended.

packages/preview/chuli-cv/0.1.0/template/cv.typ
npujol: I think I might have missed something. `media` folder is already located inside `template` folder, isn't that how it should be?
image

elegaanz: Yes, my bad, you are right.

elegaanz:
Thanks for the update, everything looks good now :) I'll merge it.

PR #637, opened 2024-05-10
mitex:0.2.4

OrangeX4:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- Fix `\boxed` command.
- Support basic figure and tabular environments.
- Add escape symbols for plus, minus and percent.
- Fix `\left` and `\right`: size of lr use em.


elegaanz:
Thank you :)

PR #641, opened 2024-05-12
in-dexter:0.3.0

RolfBremer:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes
* Support multiple named indexes. Also allow the generation of
  combined index pages.
* Support for LaTeX index group syntax (`#index("Group1!Group2!Entry"`).
* Support for advanced case handling for the entries in the index. 
  Note: The new default ist to ignore the casing for the sorting of the entries.
  The behavior can be changed by providing a `sort-order()` function to the
  `make-index` function.
* The casing for the index entry can also be altered by providing a  
  `entry-casing()` function to the `make-index` function. So it is possible
  that all entries have an uppercase first letter (which is  also the new default!).


elegaanz:
Thanks for the update!

PR #642, opened 2024-05-12
ascii-ipa:2.0.0

imatpot:
I am submitting
- [ ] a new package
- [x] an update for a package

> [!CAUTION]
> This release contains breaking changes

- General
  - **BREAKING:** Dropped support for `override-font` to avoid [`str`](https://typst.app/docs/reference/foundations/str/)/[`content`](https://typst.app/docs/reference/foundations/content/) ambiguity
  - Added bracket & braces support for [precise, morphophonemic, indistinguishable, obscured, and transliterated](https://en.wikipedia.org/wiki/International_Phonetic_Alphabet#Brackets_and_transcription_delimiters) notations
  - Added support for converting [`raw`](https://typst.app/docs/reference/text/raw/)
- Branner
  - **BREAKING:** Infixed `))` (e.g. `t))s`) is no longer supported and must now be used as a postfix (e.g. `ts))`) as per the official specification
  - Added support for more IPA characters and diacritics according to official specification
- Praat
  - Added support for more IPA characters and diacritics according to official specification
- SIL
  - Added support for more IPA characters and diacritics according to official specification
  - Added support for subscripts
  - Added support for right-bar tone glides
  - Added support for left-bar tones
  - Added Unicode support for hooks (palatal, retroflex)
  - Added Unicode support for middle tildes (velar, pharyngeal)
  - Added Unicode support for superscripts
- X-SAMPA
  - Added support for more IPA characters and diacritics according to official specification

PR #650, opened 2024-05-13
ascii-ipa:2.0.0 (metadata update)

imatpot:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This is a metadata update as I had accidentally linked the wrong repository in https://github.com/typst/packages/pull/642 which is not where e.g. issues should be reported. There are no changes to the package itself.


elegaanz:
Nice catch, thanks.

PR #653, opened 2024-05-14
modern-cv:0.3.1

DeveloperPaul123:
I am submitting
- [ ] a new package
- [x] an update for a package

Bug fixes and added a new language translation.


elegaanz:
Thanks :)

PR #672, opened 2024-05-19
simplebnf:0.1.0

Zeta611:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: simplebnf is a simple package to format Backus-Naur form. The package provides a simple way to format Backus-Naur form (BNF). It provides constructs to denote BNF expressions, possibly with annotations.


packages/preview/simplebnf/0.1.0/simplebnf.typ
```diff
@@ -0,0 +1,53 @@
...
+      left,
+      left,
+    ),
+    inset: 2.7pt,
+    stroke: none,
+    ..content,
+  )
+}
+
+#let Prod(
```
elegaanz, line 27: Typst packages generally use kebab-case names, without capital letters. If you want to match this convention, you should rename `Prod` and `Or`. But if there is a good reason to have them capitalized (or if you just prefer it this way), you can keep it as is too.

elegaanz (commented):
Thanks for the new package. Everything looks good, just a remark about naming conventions. If you want to keep your function named as they are, just tell me and I'll merge.

Zeta611:
Yes, I am aware of the convention, but renaming Or to or is not possible as it is reserved. So I capitalized them as it matches a common convention of capitalizing constructors in PL.

elegaanz:
Oh right, LGTM then.

PR #681, opened 2024-05-21
modern-sysu-thesis:0.1.1

huangjj27:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update the README so that users can reach to the repository and use the template on their local environments.


elegaanz:
Thank you.

PR #685, opened 2024-05-22
bytefield:0.0.6

jomaway:
I am submitting
- [x] an update for a package

Description: Removed tablex dependencies and some minor improvements

### DIFF
--- a/packages/preview/bytefield/0.0.6/lib/api.typ
+++ b/packages/preview/bytefield/0.0.6/lib/api.typ
@@ -26,7 +26,7 @@
     bpr: bpr,
     msb: msb,
     rows: rows,
-    side: (left_cols: pre, right_cols: post)
+    side: (left-cols: pre, right-cols: post)
   )
 
   let meta = generate-meta(fields.pos(), args)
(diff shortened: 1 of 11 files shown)


elegaanz:
There are a few places where you use snake_case for public functions or their arguments, while you seem to use kebab-case in most place. Is that intended or not? And if not, do you want me to list all the occurrences of snake_case I found?

jomaway:
I tried to replace all snake_case functions. But maybe I forgot some.

If you could highlight those you found would be very nice 👍.

I will then replace them later this evening.

packages/preview/bytefield/0.0.6/lib/api.typ
```diff
@@ -0,0 +1,247 @@
...
+  note(side, level:level, rowspan:rowspan, bracket:bracket, content)
+}
+
+/// Shows a special note with a start_addr (top aligned) and end_addr (bottom aligned) on the left of the associated row.
+///
+/// #emoji.warning *experimental:* This will probably change in a future version.
+///
+/// - start_addr (string, content):  The start address will be top aligned
+/// - end_addr (string, content): The end address will be bottom aligned
+#let section(start_addr, end_addr, span: 1) = {
```
elegaanz, line 166: The two first arguments here.

packages/preview/bytefield/0.0.6/lib/states.typ
```diff
@@ -0,0 +1,51 @@
...
+#let default_header_border = state("bf-header-border", none);
+
+// function to use with show rule
+#let bf-config(
+  row_height: 2.5em,
+  field_font_size: auto,
+  note_font_size: auto,
+  header_font_size: 9pt,
+  header_background: none,
+  header_border: none,
```
elegaanz, line 16: All these arguments.

elegaanz (commented):
There are other places where snake_case is still used, but it is not in the public interface, so it is not as important I think. Thank you for taking time to fix this. :)

jomaway:
Oh true I just refactored function names. Didn't think about parameters 🙈

packages/preview/bytefield/0.0.6/lib/states.typ
jomaway: fixed

jomaway:
OK, i fixed the one you mentioned and some more i found. So you can merge.

elegaanz:
Perfect, thank you :)

PR #686, opened 2024-05-22
diagraph 0.2.4

Robotechnic:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added support for xlabels which are now rendered by Typst
- Added support for cluster labels which are now rendered by Typst
- Fix a margin problem with the clusters


elegaanz:
Thanks!

PR #698, opened 2024-05-27
ttt-exam:0.1.2 - fix typo in  readme

jomaway:
I am submitting
- [x] an update for a package

Description: 
Not sure if this is intended, but i found a typo in the README which may confuse the user. So i only removed one character in the readme and thought this might not need a new version? Please let me know.


elegaanz:
Nice catch. This does not require a new version, indeed.

PR #702, opened 2024-05-27
glossarium:0.4.1

slashformotion:
I am submitting
- [ ] a new package
- [x] an update for a package

### 0.4.1

- Resolved an issue causing Glossarium to crash when all entries had a defined, non-empty group.

### DIFF
--- a/packages/preview/glossarium/0.4.1/LICENSE
+++ b/packages/preview/glossarium/0.4.1/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) [2023] [Enib Community]
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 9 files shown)


packages/preview/glossarium/0.4.1/LICENCE
elegaanz, line 1: There is a typo in the file name, it should be `LICENSE` with a S.

slashformotion: I wrote it in french 😂

slashformotion: I will correct it this evening

elegaanz: > I wrote it in french 😂

Yeah that's what I thought, I often do the same mistake :smile:

elegaanz:
Thank you :)

PR #713, opened 2024-06-03
add tufte-memo:0.1.0

nogula:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: this template package provides a memo document template inspired by the design of Edward Tufte's books.

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name that isn't the most obvious or canonical name for what the package does
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] `exclude`d PDFs or README images, if any, but not the LICENSE

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #718, opened 2024-06-05
tufte-memo:0.1.1

nogula:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: In #713 a reviewer asked that I rename a variable in the template, but I failed to rename all instances of that variable leading to an error when attempting to use the template. This PR (for tufte-memo:0.1.1) addresses that mistake.


elegaanz:
Thank you, and sorry to not have double-checked that everything was correctly renamed in the previous PR.

PR #722, opened 2024-06-06
silky-letter-insa:0.2.0 and silky-report-insa:0.2.0

SkytAsul:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Changed fonts for both templates, added features for the report template, restructured some code in both.


elegaanz:
Thank you

PR #725, opened 2024-06-08
equate:0.2.0

EpicEricEE:
I am submitting
- [ ] a new package
- [x] an update for a package

Changelog: 
- Added `breakable` parameter (defaults to `auto` which uses `block.breakable` of the current equation context)
- Made `number-mode: "label"` only number labelled lines, even when an outer label is added
- Improved using the `equate` function for only selected equations (without global show-all rule)
- Improved alignment algorithm (fixed some spacing issues)
- Improved performance (it still has a noticeable impact, but should be a bit faster)
- Improved parameter validation (better auto-completion in tinymist)
- Fixed figure show rules interfering with equation layout
- Fixed errors with `cases` and `vec` elements


elegaanz:
Thanks :)

PR #727, opened 2024-06-09
modern-sysu-thesis:0.2.0

huangjj27:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: modify the package to fit the specification of thesis for SYSU.


elegaanz:
Thank you

PR #729, opened 2024-06-10
unofficial-fhict-document-template:0.10.1

TomVer99:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: A couple of bug fixes regarding white pages, and behaviour change for the font fallback.


elegaanz:
Thanks :blush:

PR #731, opened 2024-06-10
diagraph 0.2.5

Robotechnic:
I am submitting
- [ ] a new package
- [x] an update for a package

- If the shape is point, the label isn't displayed
- Now a minimum size is not enforced if the node label is empty
- Added support for font alternatives


elegaanz:
Thanks !

PR #732, opened 2024-06-10
modern-resume:0.1.0

peterpf:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A package that provides a template for resumes/CVs.


packages/preview/modern-resume/0.1.0/lib.typ
```diff
@@ -0,0 +1,283 @@
+#let colors = (
+  primary: rgb("#313C4E"),
+  secondary: rgb("#222A33"),
+  accentColor: rgb("#449399"),
+  textPrimary: black,
+  textSecondary: rgb("#7C7C7C"),
+  textTertiary: white,
+)
+#let pageMargin = 16pt
```
elegaanz, line 9: Typst packages generally use kebab-case names for public variables, functions and arguments. This is only a convention, and if you personally prefer camelCase, this PR can be merged as is.

elegaanz (commented):
Thank you for the package. Just a remark about naming conventions, tell me if you want to ignore that and merge the package as is.

packages/preview/modern-resume/0.1.0/lib.typ
peterpf: Thank you for the fast review. I updated the code to follow the Typst convention (note that some functions are still in camelCase because they are not intended to be public).

Do you have a reference/best-practice guide on how to separate public from private attributes?

elegaanz: Thanks!

I'm not sure what your question is. Do you want to know how to tell if something is publicly exposed or not? Or do you want to know how to make sure something is private/public by following some pattern?

peterpf: Exactly this: "Or do you want to know how to make sure something is private/public by following some pattern".

elegaanz: Okay, then what I can recommend you is to have your package entry point explicitly import from other files what should be public. For instance:

```typ
// In functions.typ
let add = (a, b) => a + b
let mul = (a, b) => a * b
```

```typ
// In lib.typ (the entry point)
#import "./functions.typ": add
```

Here, it is easy to see that only the `add` function can be accessed by users of the package.

An alternative that may be simpler to implement is to prefix any private value/function with an `_`, but that's just a naming convention, it will not actually prevent anybody from accessing these values.

peterpf: Thank you very much for the thorough feedback and guidance.
For now I would like to merge the PR, if this is fine with you, and improve the code in future versions of the template.

elegaanz: Yes, sure.

PR #733, opened 2024-06-11
fletcher:0.5.0

Jollywatt:
I am submitting
- [ ] a new package
- [x] an update for a package

This is a major update to fletcher, which adds:

- Greatly enhance coordinate system.
  - Support CeTZ-style coordinate expressions (relative, polar, interpolating, named coordinates, etc).
  - Absolute coordinates (physical lengths) can be used alongside "elastic" coordinates (row/column positions).
- Add `label-angle` option to `edge()`.
- Add `label-wrapper` option to allow changing the label inset, outline stroke, and so on.
- Add `label-size` option to control default edge label text size.
- Add `trapezium` node shape.
- Disallow string labels to be passed as positional arguments to `edge()` (to eliminate ambiguity). Used named argument or pass content instead.


elegaanz:
Thanks!

PR #734, opened 2024-06-11
minimalbc:0.0.1

sevehub:
I am submitting
- [X] a new package
- [ ] an update for a package

This Typst template is designed for the creation of sleek and minimalist professional business cards. It offers rapid and extensive customization options for the default layout of the card.

- [x ] selected a name that isn't the most obvious or canonical name for what the package does


max397574:
tests failing because no categories were provided
also
>  excluded PDFs or README images, if any, but not the LICENSE

packages/preview/minimalbc/0.0.1/company_logo.svg
elegaanz, line 1: I don't know if this file should really be included here, as it does not seem to be actually used anywhere.

packages/preview/minimalbc/0.0.1/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+The function, **minimalbc**, allows you to customize the majority of the business card's elements. 
+
+By default, the layout is horizontal. However, it can be easily switched to a vertical layout by passing the value true to the flip argument in the minimalbc function.
+
+Here’s an example of how to use the minimalbc function:
+
+```Typst
+
+#import "lib.typ":minimalbc
```
elegaanz, line 13: It would be better to use a package import here so that people can copy/paste this example and start from here directly:

```suggestion
#import "@preview/minimalbc:0.1.0": minimalbc
```

packages/preview/minimalbc/0.0.1/lib.typ
```diff
@@ -0,0 +1,88 @@
...
+
+    grid(columns:(100%),
+        rows:(50%,10%,35%),
+        gutter:0.5mm,
+        grid(
+    columns: (60%, 40%),
+    rows: (40%, 15%, 20%,  20% ),
+    gutter: 1mm,
+    companytext(14pt,company_name),
+    grid.cell(rowspan:2, image(company_logo), align:center),
```
elegaanz, line 44: File path in packages are resolved relatively to the package: this means that the only value that will work for the `company_logo` are the images shipped with the package. The workaround is to let the user call `image` themselves and pass that as an argument to your package.

packages/preview/minimalbc/0.0.1/lib.typ
```diff
@@ -0,0 +1,88 @@
...
+  weight: "medium",
+)[#body]
+#let roletext(size, body, style: "normal") = text(
+  size: size,
+  style: style,
+  weight: "light",
+)[#body]
+
+#let minimalbc(
+  company_name: "Company Name",
```
elegaanz, line 18: Typst packages generally use kebab-case names (but this is only a convention and you can keep snake_case if you prefer).

packages/preview/minimalbc/0.0.1/README.md
```diff
@@ -0,0 +1,34 @@
...
+    email_address: "me@me.com",
+    website: "example.com",
+    company_logo: "company_logo.png",
+    bg_color: "ffffff",
+)
+
+
+```
+
+When compiled, this will produce a PDF file named businesscard.pdf located in this repository. Feel free to download and use this as a starting point for your own business cards.
```
elegaanz, line 32: Maybe I'm nitpicking here, but this is not necessarily true, you have to write your code in `businesscard.typ` or to explicitly tell the compiler to write to `businesscard.pdf`. Also the generated file will only go in this directory if people clone your project and work from here, not if they use the package from Typst Universe.

packages/preview/minimalbc/0.0.1/businesscard.pdf
elegaanz, line 1: This file seems to be a compiled test file, can you please delete it?

packages/preview/minimalbc/0.0.1/lib.pdf
elegaanz, line 1: Same for this file.

packages/preview/minimalbc/0.0.1/lib.typ
sevehub: I'm not entirely certain about this, but it's expected that users save the company logo using the specific filename 'company_logo.png'. I plan to elaborate on this and provide further clarification in the README. Currently, my time to devote to this project is limited. However, please feel free to make additions and modifications as you see fit.

packages/preview/minimalbc/0.0.1/lib.pdf
sevehub: Yes, this can be removed.

packages/preview/minimalbc/0.0.1/lib.typ
elegaanz: Even if they save their own logo under a file with the same name, this won't work as expected. Passing the `image` element directly instead of just a path is the only way to make it work at the moment.

sevehub: Done

elegaanz:
Thank you :)

PR #738, opened 2024-06-12
typearea:0.2.0

freundTech:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: typearea is inspired by the LaTeX package of the same name, which is part of KOMA-Script. It allows setting the size of the type area and margins according to typographic practices by specifying a division number and binding correction instead of a fixed margin size.


elegaanz:
Thanks!

PR #743, opened 2024-06-15
modern-sysu-thesis:0.3.0

huangjj27:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
1. add an enhencement feature of pages' header, with context of the thesis's title and chapters
2. fix numbering of figures, tables


elegaanz:
Thank you :)

PR #747, opened 2024-06-16
ccicons:1.0.0

jneug:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package offers authors who want to publish their documents under a [Creative Commons](http://creativecommons.org/) license an easy way to include the relevant icons in their documents. The package is a port of the LaTeX package of the same name by @ummels. The original author gave his permission to reuse the name for this package.


packages/preview/ccicons/1.0.0/typst.toml
```diff
@@ -0,0 +1,13 @@
...
+name = "ccicons"
+version = "1.0.0"
+entrypoint = "src/ccicons.typ"
+authors = ["J. Neugebauer"]
+license = "MIT"
+description = "A port of the ccicon LaTeX package for Typst."
+repository = "https://github.com/jneug/typst-ccicons"
+keywords = ["creative commons", "commons", "license", "icons"]
+categories = ["components"]
+exclude = ["README.md", "docs"]
```
elegaanz, line 13: It is better to not exclude the README, some tools (like language servers) may want to use it.

packages/preview/ccicons/1.0.0/src/ccicons.typ
```diff
@@ -0,0 +1,53 @@
+#import "utils.typ": ccicon, cc-url, cc-is-valid, _cc-colorize
+
+// Logos
+#let ccLogo = ccicon("logo")
```
elegaanz, line 4: Typst packages use kebab-case and not camelCase by convention (and you seem to use that in other parts of your package as well). Following this is not mandatory to get your package accepted though, so if you prefer to keep camelCase names, just tell me.

jneug: I kept the same function names as in the original package. I think I would like to keep the current naming scheme. 

The individual icons will probably not get much use, anyway. Most people will use the license icons like `#cc-by-nc-sa` and those are in the usual format.

PR #751, opened 2024-06-16
isc-hei-report:0.1.5

pmudry:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added multi-language support, captions as markdown, multi-pages code listing, new listing format for line numbers, headers


packages/preview/isc-hei-report/0.1.5/isc_template_report.pdf
elegaanz, line 1: This file was probably automatically generated, can you please delete it? And do the same for other PDF files that you committed.

packages/preview/isc-hei-report/0.1.5/isc_template_report.typ
```diff
@@ -0,0 +1,372 @@
...
+    panic("I18n key " + key + "doesn't exist")
+  }
+  return keys.at(key)
+}
+
+
+//
+// Source code inclusion
+//
+#let luma_background = luma(250)
```
elegaanz, line 91: If this value is public it would be better to use kebab-case to be consistent with the rest of the package. If you want to keep the snake_case name and it is not intended to be used outside of this file, maybe you can make that more explicit by following the underscore-prefix convention and renaming to `_luma_background`.

pmudry: This has been fixed in 23008a5
PDFs have been removed as well, sorry.

elegaanz: No worries, thank you for taking time to fix it.

PR #754, opened 2024-06-18
ilm:1.1.2

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog

- Improve template and removed old config options.

---


elegaanz:
Thanks!

PR #756, opened 2024-06-19
ilm-catppuccin:1.1.2

Quantalabs:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A fork of [`ilm`](https://typst.app/universe/package/ilm) that is themed with Catppuccin Mocha.


JamesxX:
This should be a pr to ilm surely?

elegaanz:
I think it is fine to have a fork if @Quantalabs wants to maintain one. But maybe a cleaner solution, if that suits `ilm` authors, would be to make the colors configurable in the template somehow, and then this package could just override these settings (or just not exist as it would be really easy to reproduce the same behavior by hand, by importing typpuccino and ilm).

In any case, there are some parts that were not updated with the fork, such as the `repository` field in `typst.toml` or the template thumbnail.

max397574:
In my opinion it definitely doesn't make sense to add such a thing as a new template
Making the colors configurable would definitely be the better approach
Otherwise this will just fill up the repo with a lot of unnecessary code

We can't just accept variations for all the templates with different colorschemes just because someone might want to use that
So imo we shouldn't even start with it
But in the end that is up to the maintainers whether we want such templates or not

Quantalabs:
> Making the colors configurable would definitely be the better approach

I agree, I'll do this instead.

PR #757, opened 2024-06-19
universal-hit-thesis:0.2.1

vonbrank:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: fixed some format issues, and refactored parts of the template


drupol:
Hello,

I just stumbled upon this project on [Typst Universe](https://typst.app/universe/package/universal-hit-thesis)... Sadly, I couldn't understand anything since it's not written in a language I know.

image

I have some questions:

1. Can I submit projects in any language?
2. Is there a policy enforcing the use of a language?

vonbrank:
> I have some questions:
> 
> 1. Can I submit projects in any language?
> 2. Is there a policy enforcing the use of a language?

Hi!

It seems that you can submit projects in any language. There are a bunch of thesis templates, the READMEs of which are written in Chinese,  in Typst Universe. Most of them are developed for universities in China. And you can see some templates are written in Russian.

We will provide an English version README document for `universal-hit-thesis` later.

elegaanz:
I confirm that it is allowed to submit a package with documentation written in any language.

drupol:
Thanks for all your answers.

If I understand correctly, I could tehrefore submit a package in a language that nobody understands, right?

I would like to point out that Typst Universe is primarily in English, yet we currently see projects in various languages. This creates a complicated user experience, as users may have difficulty navigating or understanding documentation that isn't in a language they know.

I have nothing against Chinese or any other languages and I fully support the diversity of the community. However, I firmly believe that for better accessibility and promotion of packages, it would be beneficial to agree on one common language.

Here are a few reasons, from the top of my head:

- Consistency: having all documentation in one language ensures a consistent user experience. Users will know what to expect and won’t have to navigate language barriers
- Accessibility: a common language which is widely used and understood globally, can make projects more accessible to a broader audience. This inclusivity can encourage more contributions and user engagement
- Ease of collaboration: when everyone uses the same language, collaboration and communication among contributors become much smoother. It minimizes misunderstandings and reduces the need for translations
- Searchability: a unified language makes it easier to search for and find relevant information. Users can more effectively use search functions in Typst Universe
- Standardization: adopting a single language for documentation can provide a sense of professionalism and standardization

While I understand and respect the diversity of culture and languages, I believe selecting one common language for documentation will significantly improve the user experience.

To balance the benefits of diversity and the need for a common language, I propose that we ask contributors to provide their `README` files in English. This way, the core information and documentation are accessible to the majority of users while allowing contributors to write the rest of the documentation or comments in their preferred language if needed. Additionally, this approach maintains the inclusivity of the community while improving the user and collaboration experience.

Thank you for considering my suggestion. I believe this would enhance the usability and appeal of Typst Universe for a global audience.

elegaanz:
For packages such as this one, that will only be used by people speaking Chinese anyways, I don't see any reason to force an English translation. Even without an English README, you can understand from the name that this package is not a general purpose one, and that you won't need to use it if you don't speak Chinese.

For other packages, I think they would naturally tend to adopt English as their main language, because package authors realize that it will be beneficial to their package to do so (for all the reasons you listed). So I don't think there is a need to enforce the rule in that case either, as it would happen anyways.

PR #771, opened 2024-06-28
vartable 0.1.0

Le-foucheur:
I am submitting
- [X] a new package
- [ ] an update for a package

it's a packet used to create variation tables, 
because at the moment the only way to do it is by hand


packages/preview/vartable/0.1.0/tabvar.typ
```diff
@@ -0,0 +1,358 @@
+#import "@preview/fletcher:0.4.5": *
+#let signe = "Sign"
+#let variation = "Variation"
+#let bo(x) = table(columns: 2cm, stroke: 0pt)[#table.cell(align: center + horizon)[#x]]
+
+// les deux prochaine fonctions son pour connaitre le nombre d’éléments à skip
+#let prochainNonVideSigne(x, i) = {
```
elegaanz, line 7: I think this function and the two below are meant to be private. If you want to properly mark them as such, you can either prefix their name with a `_` (doesn't actually make it private, but it is a convention that is adopted in the Typst ecosystem) or move all your functions to another file and only re-export the public ones from the entrypoint.

packages/preview/vartable/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "vartable"
+version = "0.1.0"
+entrypoint = "tabvar.typ"
+authors = ["Le_Foucheur"]
+license = "MIT"
+description = "A simple package to make variation table"
+repository = "https://github.com/Le-foucheur/Typst-VarTable"
+categories = ["visualization"]
+keywords = ["variation table", "variation", "table", "variationtable"]
+exclude = ["./exemples", "documentation.pdf", "documentation.typ"]
```
elegaanz, line 11: There is a typo.

```suggestion
exclude = ["./examples", "documentation.pdf", "documentation.typ"]
```

packages/preview/vartable/0.1.0/tabvar.typ
Le-foucheur: ah I didn't know this, I'll change it, thanks

PR #772, opened 2024-06-29
sweet-graduate-resume:0.1.0

InnocentZero:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a simple and neat graduate resume template useful for college students. It was designed for engineering students but I don't see why it wouldn't work for anyone else.


packages/preview/sweet-graduate-resume/0.1.0/lib_fontawesome.typ
```diff
@@ -0,0 +1,26 @@
+//! typst-fontawesome
+//!
+//! https://github.com/duskmoon314/typst-fontawesome
```
elegaanz, line 3: Is there a reason you copied the contents of this package instead of importing it?

packages/preview/sweet-graduate-resume/0.1.0/typst.toml
```diff
@@ -0,0 +1,15 @@
...
+entrypoint = "template.typ"
+authors = ["innocentzero (Md Isfarul Haque) <https://innocent_zero.codeberg.page>"]
+license = "MIT"
+description = "A simple graduate student resume template"
+repository = "https://codeberg.org/innocent_zero/typst-resume"
+categories = ["cv"]
+disciplines = ["engineering"]
+
+[template]
+path = "."
```
elegaanz, line 13: This won't work as expected: with this configuration, the whole package directory would be copied when creating a project from this template. You should move `main.typ` and the images it uses to a new sub-directory.

packages/preview/sweet-graduate-resume/0.1.0/template.typ
```diff
@@ -0,0 +1,139 @@
...
+      #box(height: 1em, baseline: 20%)[#img]
+    ] + if value == urls.last() {
+      ""
+    } else {
+      "|"
+    }
+  }
+}
+
+#let section_header(head) = {
```
elegaanz, line 65: Typst packages generally follow the kebab-case naming convention. I would recommend to rename public function and their arguments to use that instead of snake_case. However, the choice is up to you, and if you really prefer snake_case, just tell me, it won't block the package from being merged.

packages/preview/sweet-graduate-resume/0.1.0/typst.toml
InnocentZero: Oh ok. My bad. The latest version should fix it.

packages/preview/sweet-graduate-resume/0.1.0/lib_fontawesome.typ
InnocentZero: I wasn't aware of typst having local imports as well when I started making/using this and later on just went with it. I hope it is not a problem since the relevant people have been acknowledged in the source files.

packages/preview/sweet-graduate-resume/0.1.0/template.typ
InnocentZero: It was a force of habit due to rust conventions. I believe the latest commit should fix it.

packages/preview/sweet-graduate-resume/0.1.0/lib_fontawesome.typ
elegaanz: No it is not a real problem, but it would be cleaner and easier for you to maintain to just use the actual package. If you prefer to keep it as is, it is possible though.

elegaanz:
Also, the new structure is better, but not exactly what most people will expect: the code that should not be edited by users (`template/template.typ`, and the fontawesome copy if you decide to keep it) should not be in the `template` directory, but outside of it.

packages/preview/sweet-graduate-resume/0.1.0/lib_fontawesome.typ
InnocentZero: I'll probably get around to do this when I have a minor release bump. Thanks for the suggestion.

InnocentZero:
Hi @elegaanz . I finally understood what you meant when I saw. I've pushed the updated version.

packages/preview/sweet-graduate-resume/0.1.0/template.typ
```diff
@@ -0,0 +1,139 @@
...
+  set text(7pt)
+  for value in urls {
+    let img = if value.fa {
+      if value.brand {
+        fa-icon(value.svg)
+      } else {
+        fa-icon(value.svg, solid: value.solid)
+      }
+    } else {
+      image("template/svg/" + value.svg)
```
elegaanz, line 51: Unfortunately, this will not work (sorry, I could probably have noticed that in the last review too). `svg` would have to be passed as an `image` element directly, and not a string giving the filename. This is because packages can't access files outside of their folder (the template folder would be copied to the user's project, it is not considered to be part of the package, and I think you want to allow people to add their own SVG files, not just use the two Codeberg logos you provide).

So this line should become : `value.svg`, and the elements of `urls` should look like that:

```typ
#let urls = (
  (
    name: "Codeberg",
    url: "https://codeberg.org/innocent_zero",
    svg: image("svg/codeberg.svg"), // Here is the change
    fa: false,
    brand: false,
    solid: false,
  ),
  (name: "Website", url: "https://example.com", svg: "link", fa: true, brand: false, solid: false),
  (name: "Website", url: "https://example.com", svg: "github", fa: true, brand: true, solid: true),
)
```

InnocentZero: Hi, really sorry for not checking it. I simply checked if typst was able to compile it or not without caring about the pdf. I've fixed it now. Thanks for your help!

elegaanz:
Thanks :)

PR #777, opened 2024-07-01
splash:0.4.0

kaarmu:
I am submitting
- [ ] a new package
- [x] an update for a package

splash provides a library of different color palettes. This version contains new palettes from seaborn.


packages/preview/splash/0.4.0/src/palettes/xkcd.typ
```diff
@@ -0,0 +1,951 @@
+#let xkcd_rgb = (
```
elegaanz, line 1: You may want to rename this palette to `xkcd-rgb` to be more consistent with the rest of the package.

PR #792, opened 2024-07-04
tufte-memo:0.1.2

nogula:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Corrected implementation of sans-fonts in template, and turned off justify text in par's.


JamesxX:
Why is the package starting at 0.1.2?

elegaanz:
The 0.1.0 and 0.1.1 were already published from what I can see, but somehow GitHub still shows them in the diff for this PR. My guess is that the commits for these previous versions were squashed and thus are re-included when diffing from an older point of history. I think rebasing `main` in your fork @nogula would fix the issue. Could you please do it? I can also do it manually before merging this PR if you prefer.

nogula:
Hi @elegaanz, I've rebased my main branch - apologies for that mistake. Does this look correct now?

packages/preview/tufte-memo/0.1.0/template/main.typ
```diff
@@ -1,4 +1,4 @@
-#import "@preview/tufte-memo:0.1.0": *
+#import "@preview/tufte-memo:0.1.2": *
```
elegaanz, line 1: This file should not change, this version has been published and can't be changed anymore.

elegaanz (commented):
Thanks for rebasing, everything looks good apart from one small thing.

nogula:
That was unintentional and might've been a mistake doing the rebase stuff above. Sorry.

elegaanz:
Thank you :)

PR #793, opened 2024-07-05
Update package metadata

EpicEricEE:
As I split the monorepo of all my packages into separate repositories, I wanted to update the repository links (and ended up adding and updating some other entries of the `typst.toml` files as well).

No changes were made to any code, so I hope this is fine without any version bumps.


elegaanz:
Yes, it's fine. Thank you :)

PR #801, opened 2024-07-10
chordx:0.4.0

ljgago:
I am submitting
- [ ] a new package
- [x] an update for a package

### New:

- Background color in chord names.
- Chart position below or above of the chord name.
- Updated the old styles-measure syntax by the new context-measure syntax.
- Refactored the render and the graphic objects, now each object calculates its own relative position.
- Removed closure functions in favor of the "with" property.


elegaanz:
Thank you!

PR #814, opened 2024-07-15
sunny-famnit:0.2.0

Tiggax:
I am submitting
- [ ] a new package
- [x] an update for a package

Update to the package `sunny-famnit` changes:
- changed how `mentor` and `comentor` are written in the settings
- fixed wrong referencing number
- added `work_mentor` as an option


packages/preview/sunny-famnit/0.2.0/examples/example.typ
```diff
@@ -0,0 +1,85 @@
...
+  date: datetime(day: 1, month: 1, year: 2024),
+  author: "Tilen Gimpelj",
+  studij: "Bioinformatika",
+  mentor:(
+    name: "Oge So-Kul", 
+    sl: ("dr.",""),
+    en: ("",".phD"),
+    ),
+  somentor: none, // optional somentor, written in the same way as mentor
+  work_mentor: none, // optional work mentor can be written the same way as mentor
```
JamesxX, line 36: Prefer snake case for variable names

laurmaedje: I think you mean kebab-case? If yes, indeed, it's preferred. It's not a hard requirement, but I'll hold off on merging until there's a response.

JamesxX: Whoops absolutely I meant kebab case (I dread to think how many times I have mixed those two up 😭)

PR #821, opened 2024-07-20
unofficial-FHICT-Document-template:0.11.0

TomVer99:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- Removed two packages which were effectively bloat
- Added support for multiple languages
- Updated three base packages to the newest versions
- Updated the README for the typst universe page


TomVer99:
`Removed two packages which were effectively bloat`

This is within the template, nothing else.

elegaanz:
Thanks!

PR #822, opened 2024-07-21
grape-suite:1.0.0

piepert:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added new functionality, added possibilities for some more customization, fixed forced language issues.


elegaanz:
Thanks :)

PR #824, opened 2024-07-21
preavis:0.1.0

mathias-aparicio:
I am submitting
- [x ] a new package
- [ ] an update for a package

Description: The package is a template for writing a french move out letter. I am a student and need to write a letter to move out from my appartement typst help me make it standard, and be able to write it multiple times, I think it might be helpful for other french people that are in my case.

### DIFF
--- a/packages/preview/preavis/example.typ
+++ b/packages/preview/preavis/example.typ
@@ -1,0 +1,24 @@
+#import "lib.typ": lettre_preavis, locataire, adresse, proprietaire
+#lettre_preavis(
+  locataire: locataire(
+     "Dupont locataire",
+     "Jean",
+     adresse(
+       "123 rue de la Paix",
+       "75000",
+       "Paris",
+       complement: "Appartement 2"
+    )
+  ),
+  proprietaire: proprietaire(
+     "Martin proprietaire",
+    "Sophie",
+     adresse(
+       "456 avenue des Champs-Élysées",
+       "75008",
+       "Paris"
+    ),
+    "Madame"
+  ),
+  date_etat_des_lieux: datetime(year:2024, month:9, day:21)
+)
--- a/packages/preview/preavis/LICENSE
+++ b/packages/preview/preavis/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Mathias APARICIO
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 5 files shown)


packages/preview/preavis/example.typ
```diff
@@ -0,0 +1,24 @@
+#import "lib.typ": lettre_preavis, locataire, adresse, proprietaire
+#lettre_preavis(
```
elegaanz, line 2: The convention in Typst packages is to use kebab-case names for functions and their arguments. However, if you prefer, you can keep snake_case names.

elegaanz (commented):
Two things:
- each version of a package should have its own sub-directory: you should create `packages/preview/preavis/0.1.0/` and move all your files there.
- the name is too descriptive for what the package does, could you please change it for something more unique?

elegaanz:
Thank you :)

elegaanz:
You still need to upload a thumbnail for your template and specify that in your `typst.toml`, but then it should be good to merge.

packages/preview/simple-preavis/0.1.0/typst.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "simple-preavis"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Mathias APARICIO <https://github.com/mathias-aparicio/>"]
+license = "MIT"
+description = "📖 a french move out letter"
+repository = "https://github.com/mathias-aparicio/simple-preavis"
+categories = ["letter"]
```
elegaanz, line 9: This is not a valid category, see [this list](https://github.com/typst/packages/blob/main/CATEGORIES.md).

packages/preview/simple-preavis/0.1.0/typst.toml
```diff
@@ -0,0 +1,15 @@
...
+exclude = [".gitignore"]
+
+[template]
+entrypoint="lib.typ"
+path="."
+thumbnail = "thumbnail.png"
```
elegaanz, line 15: This won't work as expected: I think you should create a `template` subfolder in which you move your `example.typ` file, and then update the manifest to the following:

```suggestion
[template]
entrypoint = "example.typ"
path = "template"
thumbnail = "thumbnail.png"
```

With your current code, the whole package would be copied when creating a project from this template.

elegaanz (commented):
Two more things I didn't catch, sorry.

elegaanz:
Thanks!

PR #830, opened 2024-07-23
bubble:0.2.0

hzkonor:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Simple template with a main color that gets applied to list items, headings...
This update fixes the problem with coloring words that didn't properly work and removes some useless code


elegaanz:
Thanks :)

packages/preview/bubble/0.2.0/lib.typ
```diff
@@ -0,0 +1,126 @@
...
+  let body-font = "Source Sans Pro"
+  let title-font = "Barlow"
+
+  // Set colors
+  let primary-color = rgb(main-color) // alpha = 100%
+  // change alpha of primary color
+  let secondary-color = color.mix(color.rgb(100%, 100%, 100%, alpha), primary-color, space:rgb)
+
+  // highlight important words
+  show regex(color-words.join("|")): text.with(fill: primary-color)
```
elegaanz, line 28: If `color-words` is an empty array, this will crash (`().join("|")` returns `none` and not a string), which will probably happen often as this is the default value of the argument. If this is expected, please tell me, otherwise you can probably only apply this show rule if the array is not empty.

elegaanz:
Thank you! :)

PR #833, opened 2024-07-23
ilm:1.1.3

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog
-  fix: chapter page breaks inside page container (https://github.com/talal/ilm/commit/a10a76de19803cedc21840f1b8cbc905ac464c79)

---


elegaanz:
Thanks!

PR #840, opened 2024-07-26
elsearticle:0.2.0

maucejo:
I am submitting
- [ ] a new package
- [x] an update for a package

This update adds support for the `subpar` package. This implies that : 
- The previous API for `subfigure` is is deprecated and no more supported
- The API for `subfigure` is now a wraper around `subpar.grid` with `numbering` and `numbering-sub-ref` adapted to the `context` (main body or `appendix`)
- The customization of `ref` and `figure` is simplified

Description: This template will allow the users to write papers formatted for submission to Elsevier


packages/preview/elsearticle/0.2.0/template/main.typ
```diff
@@ -0,0 +1,138 @@
...
+#show: appendix
+
+= Appendix A
+
+== Figures
+
+In @fig:app
+
+#figure(
+  image("../template/images/typst-logo.svg", width: 50%),
```
elegaanz, line 99: The image paths were correct in v0.1.0: they should be relative to the current file, which will live in the root of the project created from the template, not in a template subdirectory. You can also use absolute paths if you want, the root being the root of the project.

maucejo: Of course, you are right. I forget to modify these liens when submitting the update version. Thanks for your careful check.

packages/preview/elsearticle/0.2.0/template/main.typ
```diff
@@ -0,0 +1,138 @@
...
+    kind: table,
+    caption : [Example]
+) <tab:tab1>
+
+=== Figures
+
+Below is @fig:logo.
+
+#figure(
+  image("../template/images/typst-logo.svg", width: 50%),
```
elegaanz, line 74: This one is still wrong.

maucejo: Sorry for that. I think it should be okay...

elegaanz:
Thank you :)

PR #847, opened 2024-07-29
aio-studi-and-thesis:0.1.0

fuchs-fabian:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:

With this unofficial template, students can easily create documentation or a thesis without having to create the layout themselves. It is highly customisable when it comes to tables of contents, tables of figures, etc. The colours, font colour, adjustments to the tables of contents etc. can be made in one central location. It can be used in both German and English. The cover page and the explanation can be easily exchanged if required. If a special table of contents is missing, it can simply be inserted so that it is in the right place. This saves a lot of time, especially when professors at a university are constantly making new demands on the design of the document.


fuchs-fabian:
https://github.com/typst/packages/pull/845

I had previously opened a pull request, but it didn't work because the thumbnail wasn't the right size. Now I had to enlarge it. I hope this was the right approach :)

fuchs-fabian:
> I think you could have just pushed another commit to your previous PR, but re-doing it works as well :) Thanks!

@elegaanz Thank you very much and especially that it went so quickly :)

PR #850, opened 2024-07-30
quetta:0.1.0

FlorentCLMichel:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: The goal of this package is to make it easier to write Tengwar in Typst. Tengwar is one of the scripts invented by J. R. R. Tolkien, and primarily used for the fictional elvish languages. While it is technically possible to write Tengwar in vanilla Typst with the right fonts, this requires memorising the (not always intuitive) map from ascii symbols. This package uses a phonetic approach, providing a command `quenya` turning a string in the Latin alphabet into Tengwar in such a way that each sound is mapped to the correct symbol. It also implements several spelling rules to pick the right symbol when the map between sound and symbol is ambiguous. 

It is of course a bit niche - few people write in Tengwar on a regular basis. However, as there is, as far as I am aware, currently no open-source tool for writing Tengwar efficiently with the wealth of formatting options that Typst provides, I would imagine it can be useful, if only to fans of Tolkien for generating content inspired by his works. Please feel free to close this pull request if you think it is of too narrow interest.


packages/preview/quetta/0.1.0/src/tengwar_proto.pdf
elegaanz, line 1: Can you please delete this file?

FlorentCLMichel: Ah, indeed I missed it when cleaning up my local version.
Thanks for having noticed it! Should be deleted now.

elegaanz:
Thanks :)

PR #862, opened 2024-08-05
typographix-polytechnique-reports:0.1.3

remigerme:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog
- enabling user-specified logo on cover page (horizontal and vertical mode)
- updating pdf guide (still excluded from the package)


packages/preview/typographix-polytechnique-reports/0.1.3/cover.typ
```diff
@@ -0,0 +1,100 @@
...
+  smallcaps(author)
+
+  v(1fr)
+
+  let logo-height = if (logo-horizontal) { 20mm } else { 30mm }
+  let path-logo-x = if (logo-horizontal) { "assets/logo-x-ip-paris.svg" } else { "assets/logo-x.svg" }
+
+  grid(
+    columns: (1fr, 1fr), align: center + horizon,
+    if (logo != none) { image(logo, height: logo-height) } else { smallcaps("Insert your logo") },
```
elegaanz, line 80: If you expect `logo` to be a path to an image file, this won't work, as packages can only access files in their own directory, not in the user's project. What you should do instead, is directly take an `image` element as input, and let the user of your template call `image` with the correct path (you can still automatically apply height with a `set` rule).

elegaanz:
Thank you!

PR #874, opened 2024-08-09
keyle:0.1.1

magicwenli:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- Refactor the theme switching logic. Switch from `state` to factory method pattern.
- Add Linux Libertine Biolinum Font style


elegaanz:
Thank you.

PR #885, opened 2024-08-15
unofficial-fhict-document-template:1.0.0

TomVer99:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

Updated the following (roughly described):

- Updated table styling
- Added the option to use the template in a multitude of languages and made sure people can easily add their own
- Added an `assessor` option
- Fixed cover page title/subtitle/author/assessor layout
- Added chapter layout option for every chapter on a new page if not already on a new page
- Added a header with chapter information and document title
- Fixed documentation
- Added Table of tables option
- And some more...

Regarding the use of `version 1.0.0`. I believe that the current state of the template will only require maintenance. All features I, and a few other students could think of over the past year have been implemented in a way I am satisfied with. Thus I believe `1.0.0` is a fitting version number for this template.


elegaanz:
Thank you :)

PR #894, opened 2024-08-19
flyingcircus:3.0.0

Tetragramm:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: Template and functions for formatting content for the Flying Circus RPG. Integrated with the aircraft builder and vehicle builder, as well as styling the document.


packages/preview/flyingcircus/3.0.0/fonts/Balthazar-Regular.ttf
elegaanz, line 1: Bundling fonts with packages is not supported at the moment. What we generally recommend to do as a workaround is to have a link to download the fonts in the README, and a brief explanation on how to use them with Typst (system wide installation, `--font-path`, or upload of the TTF files if you use the web app).

packages/preview/flyingcircus/3.0.0/template/Main.typ
```diff
@@ -0,0 +1,130 @@
...
+  + Sub-Lists!
++ Things! But you can have multiple lines for an item by indenting the next one, or just one long line.
+  - Sub-items!
+
+- Unnumbered list
+ 
+Break the column where you want it with `#colbreak()`
+
+Images can be added by doing `#image(path, width:blah, blah)`. The FC functions do fancy stuff though, so they need you
+to do `read(path,encoding:none)`
```
elegaanz, line 35: I feel like a more idiomatic way would be to have the properties of `image` configured with a `set` rule, and take `image`s as parameters rather than bytes. For instance, this:

```typst
#let func = (img) => {
    image.decode(img, width: 80%)
}
```

Would become:

```typst
#let func = (img) => {
    set image(width: 80%)
    img
}
```

elegaanz (commented):
Typst packages generally use kebab-case for public function names and their arguments. This is a convention, but it is not necessary to adhere to it to get your package published, so if you want to stick to PascalCase you can. Otherwise, I'll let you push an update.

I have also left two other comments.

Tetragramm:
I have taken your suggestions, except the kebab case.  I'd like it to match the existing latex template in that regard.

Tetragramm:
Actually, could the fonts go in the template folder?  Since it copies that folder into the workspace, that seems like it would work for anyone who creates a project based on the template.
Or is there a filter of some sort that would remove them?

packages/preview/flyingcircus/3.0.0/src/Impl.typ
```diff
@@ -0,0 +1,687 @@
...
+  }
+   
+  //Reset Page counter to 1, and let's go!
+  counter(page).update(1)
+   
+  body
+}
+
+#let MaybeImage(img, ..args) = if (img != none) {
+  image.decode(img, ..args)
```
elegaanz, line 206: For the new way of passing images to work correctly you need to update this function to something like:
```suggestion
set image(..args)
img
```

Tetragramm: Apologies, I made a mistake in manually copying changes. Fixed now.

elegaanz:
For the fonts, I think that would work, but only in the web app. If the template was used with the CLI, users would still have to specify `--font-path` themselves. Font files can also be quite heavy, so it is better if they don't bloat the history of this repository too much, it is already large enough.

Anyway, thanks for the package and for taking into account review comments :)

PR #903, opened 2024-08-22
fh-joanneum-iit-thesis:1.1.0

internettechnik:
I am submitting
- [ x ] a new package
- [ ] an update for a package

Description: This template is for a Bachelor's or Master's thesis at FH JOANNEUM, IIT (Institute of Software Design and Security).

- [ x ] selected a name that isn't the most obvious or canonical name for what the package does
- [ x ] added a `typst.toml` file with all required keys
- [ x ] added a `README.md` with documentation for my package
- [ x ] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [ x ] tested my package locally on my system and it worked
- [ x ] `exclude`d PDFs or README images, if any, but not the LICENSE

- [ x ] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.


elegaanz:
Hi and thanks for the package. I feel like the name is a bit too descriptive of what it does (which we prefer to avoid to not give your package an unfair advantage in case someone else wants to provide an alternative template for the same purpose). Could you please change it? It does not have to be super unique, adding an adjective would be enough for example.

Also there are a few naming warnings that were reported, I'll let you fix them if you want to follow the kebab-case convention that most Typst packages use (if you don't want to, you can keep it as is).

laurmaedje:
Hey, it depends on whether you are officially affiliated with the department. If that's the case, just drop us a mail to hello@typst.app from your official email and then a descriptive name is okay.

If not, the reason for disallowing it is that, should there ever be an official template, it might want to have the descriptive name.

PR #910, opened 2024-08-24
thesist:0.1.0

tfachada:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A Master's thesis template for Instituto Superior Técnico (IST)


packages/preview/thesist/0.1.0/template/main.typ
```diff
@@ -0,0 +1,84 @@
...
+
+/* INIT THESIS */
+// Set language to en/pt
+// Setup title page
+// Hide unused lists
+#show: thesis.with(
+
+  lang: "en",
+
+  cover_image: image("Images/default_cover.jpg", width: 95%),
```
laurmaedje, line 11: It is recommended to use kebab-case, as in `cover-image`. It is not a requirement, but it is the common convention in Typst.

tfachada: Ok, will correct. I now have these doubts, though:

- Does that also apply to filenames? If yes, is there an issue with the example files and folders within `template`, which I used a different convention on due to them being on the user side?

- Does that also apply to the `STRING` variables I defined in the `thesis` function in `layout.typ` (which I named to look like `#define` C constants due to their immutability)?

laurmaedje: - With filename, I think there's less of a convention, but I personally would also use kebab-case.
- I haven't seen much all uppercase variables in Typst code anywhere, so I would say it's not common, but I believe the string variables are private anyway, right? Then, it doesn't matter that much.

tfachada: Ok, done! Let me know if there's anything else.

PR #911, opened 2024-08-24
vienna-tech:0.1.1

npikall:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: added some functionality where toc can be disabled.


packages/preview/vienna-tech/0.1.1/typst.toml
```diff
@@ -0,0 +1,28 @@
...
+repository = "https://github.com/npikall/vienna-tech.git"
+keywords = ["thesis", "tuw", "TU Wien", "civil engineering", "environmental engineering"]
+categories = ["thesis"]
+exclude = [
+	".github",
+	"docs",
+	"scripts",
+	"tests",
+	".typstignore",
+	"Justfile",
```
laurmaedje, line 19: I recommend excluding the `thumbnail-long.png` file to keep the package size down.

PR #927, opened 2024-08-31
not-tudabeamer-2023:0.1.0

mohe2015:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Mimics the TU Darmstadt beamer presentation template in its 2023 edition to allow other students to use Typst to create their presentations.


laurmaedje:
Would it make sense to call it just `not-tudabeamer` and just bump the major version number when there's a 2024/2025/... version?

mohe2015:
The current situation is a bit weird I would say. There have been official PowerPoint templates by the university for a while but the Latex Template (https://github.com/tudace/tuda_latex_templates/tree/main/examples) has still not released an update with the new template. Also it seems to keep both version for now. Additionally I have personally heard from some students that they don't like this style and therefore I think it could make sense that somebody wants to publish a typst template with the old style. So to summarize I think it makes at least a bit of sense to clearly mark this in the name but I see how this is going to be problem when the university will publish a new style at some point. I think it could then again make sense to just publish a new package so they can be used simultaneously (and I won't have the burden of maintaining two styles or not being able to patch the old style).

laurmaedje:
If you prefer to keep the 2023 suffix, it's fine. Just let me know what you decided.

mohe2015:
I would prefer to keep it.

OrangeX4:
Are you willing to upgrade to touying 0.5.1? This will keep the current version of the touying theme in typst universe consistent. Of course, it's also feasible to still keep 0.4.2, it's up to you. Thank you very much for creating this beautiful theme. :-)

packages/preview/not-tudabeamer-2023/0.1.0/template/main.typ
```diff
@@ -0,0 +1,56 @@
+#import "@preview/touying:0.4.2": utils
+#import "@preview/not-tudabeamer-2023:0.1.0": register
+
+#let tuda_logo = read("tuda_logo.svg")
```
OrangeX4, line 4: Would you like to include this logo file in the template?

packages/preview/not-tudabeamer-2023/0.1.0/lib.typ
```diff
@@ -0,0 +1,178 @@
...
+  ..args,
+) = {
+  self = (self.methods.datetime-format)(self: self, "[day].[month].[year]")
+  self.enable-header = true
+  let header = self => align(top)[
+    #if self.enable-header {
+      v(0.39in)
+      header-font(upper(self.info.short-title + " / " + self.info.short-author))
+    }
+    #place(top + right, dx: 0.34in, dy: 0.2in)[#image.decode(tuda_logo, height: 0.99in)]
```
OrangeX4, line 127: Maybe it would be better here for it to show any logo content instead of an image. This way we can also use emoji as a logo. You can refer to the practice of the touying stargazer theme. Of course, this is not necessary, but it will be more user-friendly.

packages/preview/not-tudabeamer-2023/0.1.0/lib.typ
```diff
@@ -0,0 +1,178 @@
...
+  }
+  if outline-slide {
+    (self.methods.outline-slide)(self: self)
+  }
+  (self.methods.touying-slides)(self: self, slide-level: slide-level, ..args)
+}
+
+#let register(
+  self: themes.default.register(),
+  tuda_logo,
```
OrangeX4, line 117: Consider putting it in `self.info.logo`.

packages/preview/not-tudabeamer-2023/0.1.0/template/main.typ
mohe2015: Unfortunately according to our university website this is not easily possible.
> The TU Darmstadt logo is registered as a word and image mark with the German Patent and Trademark Office and enjoys special protection.

OrangeX4: OK.

mohe2015:
This is ready for review/merge.

Thank you both for your awesome work with typst and touying!

mohe2015:
While the thumbnail doesn't exactly represent the default initialized template (the template uses an emoji instead of the university logo) I would prefer to keep the thumbnail as is if possible because the primary user of this template are probably students from TU Darmstadt and in this way they can more easily identify this template.

laurmaedje:
> While the thumbnail doesn't exactly represent the default initialized template (the template uses an emoji instead of the university logo) I would prefer to keep the thumbnail as is if possible because the primary users of this template are probably students from TU Darmstadt and in this way they can more easily identify this template.

That's okay. Thanks!

PR #936, opened 2024-09-03
touying-buaa:0.2.0

Coekjan:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Because @OrangeX4 made stargazer as touying built-in theme and released touying 0.5.0 (a disruptive version), I simplified the implementation of this package/template using the built-in theme and bumped to touying 0.5.0


Coekjan:
Wait for #937

packages/preview/touying-buaa/0.2.0/lib.typ
```diff
@@ -0,0 +1,30 @@
+// Stargazer theme - adapted for BUAA
+#import "@preview/touying:0.5.0": *
```
OrangeX4, line 2: #import "@preview/touying:0.5.2": *

Coekjan: I saw https://github.com/touying-typ/touying/commit/334f8a2190ae1767c269b75474c7d33f0aa8f0bc . Would it indicate releasing 0.5.3 soon?

OrangeX4: I don't think so, because it's a minor bug and it may be a week or two, or even longer, before 0.5.3 arrives. Maybe when typst 0.12.0 comes out, the incompatible content will be dealt with as well.

Coekjan: Ok, let us use 0.5.2 for now.

PR #940, opened 2024-09-03
minor fix for touying 0.5.1

OrangeX4:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I'm sorry for missing a simple bug fix. Perhaps I can include it as a minor fix? If not, I'll consider moving it to a later version. Sorry to bother, anyway many thanks.


laurmaedje:
Bug fixes cannot be submitted as minor fixes because people might have already downloaded the buggy version and then they have no way to get the fix (unless they clear their cache).

OrangeX4:
ok, i will create a new version, thanks.

PR #967, opened 2024-09-12
unichar:0.1.0

MDLC01:
I am submitting
- [x] a new package
- [ ] an update for a package

Unichar ports part of the [Unicode Character Database](https://www.unicode.org/reports/tr44/) to Typst. Notably, it includes information from [UnicodeData.txt](https://unicode.org/reports/tr44/#UnicodeData.txt) and [Blocks.txt](https://unicode.org/reports/tr44/#Blocks.txt).

The name comes from [Template:Unichar](https://en.wikipedia.org/wiki/Template:Unichar), a template used on Wikipedia to display Unicode characters. Since the package's main goal is more general than just displaying Unicode characters, and the name "Unichar" is not strongly associated with the Wikipedia template for most people, I figured this was fine. I can change the name if necessary.


laurmaedje:
Thanks!

PR #975, opened 2024-09-14
tgm-hit-thesis:0.1.2 (template import correction)

SillyFreak:
I am submitting
- [ ] a new package
- [x] an update for a package

Once again an error in my initial submission, sorry...

this corrects the template to actually point at the published package, not the local version

### DIFF
--- a/packages/preview/tgm-hit-thesis/0.1.2/template/lib.typ
+++ b/packages/preview/tgm-hit-thesis/0.1.2/template/lib.typ
@@ -1,4 +1,3 @@
-// #import "@preview/tgm-hit-thesis:0.1.2": *
-#import "../src/lib.typ": *
+#import "@preview/tgm-hit-thesis:0.1.2": *
 
 #import "assets.typ"


laurmaedje:
This _does_ change behaviour though, so it would need a new version. :/

SillyFreak:
Just in case you missed that, the changed import is in the template. So this change won't change behavior for someone who imports package version 0.1.2. It _will_ change what happens on `typst init tgm-hit-thesis:0.1.2`, by not pointing at a file outside the template. The only case in which the original version would compile is if the project root is set one level up and is next to a local checkout of this package.

If that makes it ok in your opinion, sorry for not including enough context initially. If not, I still think 0.1.2 as it currently is should be yanked, which is not less breaking.

laurmaedje:
The thing is that people who've already downloaded 0.1.2 won't be able to get the fix. Even if they redo `typst init`, they already have a cached bad version.

SillyFreak:
so 0.1.3 then? is yanking an option, or no?

SillyFreak:
ok, this PR is superseded now. If anything, this teaches me to not do manually what I have already automated :P

let me know if yanking 0.1.2 is possible in your opinion, then I'll make a PR for that.

laurmaedje:
What do you mean with yanking? There isn't really a process for that. If you mean just deleting, then I think it's better to not do it, since if you adjusted the template 0.1.2 _does_ work and in theory people could rely on it. (Even if that is improbable for the package in question, this would set a precedent.) I think the fact that there is a newer version means that only people who already used and fixed 0.1.2 would ever try to use it.

SillyFreak:
as in [`cargo yank`](https://doc.rust-lang.org/cargo/commands/cargo-yank.html), but in case of typst packages I meant just deleting the version, which is of course a bit different. Yeah, probably not a good precedent.

PR #984, opened 2024-09-17
package/unofficial-FHICT-Document-template:1.0.2

TomVer99:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added a version nr display on the cover page

This PR is very small in terms of new features, but this is required for the current internship I'm doing so I need it as soon as I can manage.


TomVer99:
Thanks for the very quick review ❤️

PR #998, opened 2024-09-23
frackable:0.1.0

jamesrswift:
I am submitting
- [x] a new package

Description: Creates vulgar fractions using Unicode symbols

example


EpicEricEE:
Do you really want the parameters to be named? I feel like typing out "numerator" and "denominator" can get annoying very quickly, and the meaning of the parameters should still be relatively clear even when they are positional.

jamesrswift:
> Do you really want the parameters to be named? I feel like typing out "numerator" and "denominator" can get annoying very quickly, and the meaning of the parameters should still be relatively clear even when they are positional.

done

elegaanz:
Thank you :)

PR #1005, opened 2024-09-25
haw-hamburg:0.1.0

LasseRosenow:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This allows Typst users to write their bachelor or master thesis using the design of the haw-hamburg.


LasseRosenow:
I haven't excluded some PDFs from examples so users can have a fast view how the template looks like.
I saw that many packages have example PDFs. If you  think I should remove them, I can remove them of course.

elegaanz:
Because you excluded the whole `examples` directory, the PDF files are excluded as well. And it makes sense to do so in my opinion, because these files are not necessary to have in the local cache for the package to work correctly (and most people will have a look in your GitHub repo if they want to see the examples, not in the package cache on their machine).

The bot listed a few variables and functions that are not named following Typst conventions, but you can choose to ignore these warnings if you want to, tell me what you prefer.

LasseRosenow:
Thanks for letting me now, I have just fixed all the warnings :)
I always try to adhere to the naming conventions of course, but still need to get used to it :)

elegaanz:
Thank you for making the changes, and for the package :)

PR #1031, opened 2024-10-02
orange-book:0.2.0

flavio20002:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I fixed a bug regarding the figure and equations label.


elegaanz:
Thank you :)

PR #1035, opened 2024-10-03
zero:0.2.0

Mc-Zen:
I am submitting
- [x] an update for a package

- Added support for using non-math fonts for `num` via the option `math`. This can be activated by calling `#set-num(math: false)`. 
- Performance improvements for both `num()` and `ztable(9)`


elegaanz:
Thanks :)

PR #1044, opened 2024-10-05
basic-resume:0.1.2

stuxf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Emails, github, etc. in the resume header are now optional.


elegaanz:
Thank you!

PR #1055, opened 2024-10-08
haw-hamburg:0.2.0

LasseRosenow:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This package allows to write a bachelor thesis, master thesis or report using the style of the haw hamburg.

What is new in this version:

- Improved Readme documentation around fonts
- Added List of Figures
- Added List of Tables
- Added Listings
- Added abbreviations to the template
- Fixed header showing chapter with number "0" for chapters without numbering


elegaanz:
Hello, can you please change the name of your package to something more unique? The idea is to not have an unfair advantage in case someone else wants to submit an alternative package for the same purpose. If you don't have much inspiration you can simply add an adjective to the current name, or try to find a pun.

LasseRosenow:
So that means also changing the name of the already released 0.1.0 package?

elegaanz:
Oh sorry, I didn't notice it was an update and not a new package. My mistake then, I should have asked you to change the name the first time. You can keep it.

LasseRosenow:
Thanks a lot :) I will try to come up with better names for future packages :)

LasseRosenow:
Sorry for the confusion I think I have learned now that it is not allowed to publish templates that don't contain a lib file. So templates would always be packages correct? 

In this case maybe it is not needed to revert this since it only changes the visibility of this package from package to template and that I guess can also just be reverted in the future or extended if new features come. I think it should be fine to just wait if my problem is addressed eventually :) No need to worry about it

PR #1062, opened 2024-10-09
bookletic:0.3.0

harrellbm:
I am submitting
- [ ] a new package
- [x] an update for a package

#### 0.3.0 Changes
- Remove internal dependency on native page function. This allows the user to set the page function separately with full control over paper type, outer margins and everything else defined by the native page function.
- Add p-num-halign-alternate to page number layout allowing setting page numbers to alternate on facing pages making it possible to place page numbers along the outside or inside edges of facing pages.
- Internal improvements for ordering algorithm.
- Add `num-layout` function helper.


elegaanz:
Thank you for the package update :) For the `exclude` that is not working, I think it is now correct and the bot just has a bug (and even if it's right, it's not that bad if these two files end up in the archive).

PR #1067, opened 2024-10-10
grotesk-cv:0.1.5

AsiSkarp:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixes relative file path string for import function, to correctly import sections from local preview


elegaanz:
Thanks :)

PR #1072, opened 2024-10-10
cetz:0.3.0

johannes-wolf:
I am submitting
- [ ] a new package
- [x] an update for a package

# Changes

CeTZ 0.3.0 requires Typst 0.11.0
The licence changed from Apache-2.0 to LGPLv3.

CeTZ' plotting and charting functionality has been moved to a separate 
package called `cetz-plot`.

## Canvas
- Add runtime cetz version check support (see `assert-version`).
- Fixed a bug with `#set place(float: true)` affecting the canvas.
- Transformation matrices are now rounded.
- The default coordinate system changed to a right-hand side system.
  Use `scale(z: -1)` if you want to change back to a left-hand system.
- The `on-<axes>` functions lead to wrong anchors and got fixed. The offset
  argument is now behaving as translation instead of hard setting the coordinate!
- A new `scope(...)` element got added that behaves like an unnamed group but
  leaking child elements to the outside. This element can be used for scoping
  transformations, without having to scope children under a group name.
- The center anchor of `content()` with two coordinates got fixed when using
  negative cordinates.
- Element names are now checked to not contain a "." character.
- Fixed intersection bug for content with `anchor:` set.
- **BREAKING** The winding order of _all_ elements has been changed to CCW.
- You can now add custom marks via `register-mark`.
- Mark anchor (tip, base, center) calcuation got fixed.
- Content border anchors for `text` now default to the texts bounds.
- Content emit new anchors for accessing the baseline of the content.

## Draw
- Added `floating` function for drawing elements without affecting bounding boxes.
- The `ortho` function gained a `sorted` and `cull-face` argument to enable
  depth ordering and face culling of drawables. Ordering is enabled by default.
- Closed `line` and `merge-path` elements now have a `"centroid"` anchor that
  is the calculated centroid of the (non-self-intersecting!) shape.

## Marks
- Added support for mark `anchor` style key, to adjust mark placement and
  allow for centered marks.

## Plot
- **BREAKING** The plot library has been moved out of cetz
- Added support for automatically adding axis breaks (sawtooth lines) by setting the `break`
  attribute of an axis to `true`.
- Added a new errorbar function: `add-errorbar`
- Added errorbar support to the `add-bar` function
- Improved the performance & memory consumption of the clipping algorithm
- **BREAKING** Legend anchors got renamed and do not use the legend prefix anymore

## Chart
- **BREAKING** The chart library has been moved out of cetz
- Added errorbar support for bar- and columncharts
- Piecharts now support a legend (see `legend.label` style)
- **BREAKING** Legend anchors got renamed and do not use the legend prefix anymore

## Anchors
- `copy-anchors` no longer requires copied anchors to have a default, allowing the copying of an element's anchors from a group as expected.

## Util
- `apply-transform` now allows passing a function that takes and returns a vector, instead of a transformation matrix.
- Fixed matrix-vector multiplication.


elegaanz:
Thank you :)

PR #1078, opened 2024-10-12
scrutinize:0.3.0

SillyFreak:
I am submitting
- [ ] a new package
- [x] an update for a package

Scrutinize is a library for building exams, tests, etc. with Typst. This is a major breaking release that makes writing questions less intrusive and adds more question writing utilities, including new question types and more configuration options for existing question types.


elegaanz:
Thanks :)

PR #1085, opened 2024-10-14
minerva-report-fcfm:0.2.1

Dav1com:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixing importing and older version in some files of the template.


elegaanz:
Thanks!

PR #1090, opened 2024-10-14
rich-counters:0.2.0

jbirnick:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Counters that can depend on other counters. This PR is a minor fix for version `0.2.0`.


elegaanz:
Can you please make that a new patch release? Once a package is published, its code should not change.

jbirnick:
also if it was just accepted yesterday?

elegaanz:
I agree that the rule of having to make a new release is a bit strict but I prefer to enforce it at all times rather than judging individually for each package if this should be a patch release or if it can be hotfixed in the previous release. Also, patch releases exist for this kind of changes, it makes sense to have a new release just for that bug fix. Sorry, I understand that this is a bit more work for you.

PR #1115, opened 2024-10-19
graceful-genetics:0.2.0

jamesrswift:
I am submitting
- [x] an update for a package

Description: Updated template to work with the new column layout workflow in Typst 0.12. Additionally exposed means of altering venue at the top of the page.


elegaanz:
Can you remove the empty `homepage` field in your `typst.toml` please? After that it should be good to merge.

elegaanz:
Thank you!

PR #1123, opened 2024-10-20
fontawesome:0.5.0

duskmoon314:
I am submitting
- [ ] a new package
- [x] an update for a package

In type 0.12.0, a warning will be thrown if a font is not found. This causes several warnings when using the pro sets. The new version adds a state to control the pro sets. If the state is set to true, the pro sets will be used; otherwise, the pro sets will not be used.

Users will not receive unexpected warnings in the new version.

### DIFF
--- a/packages/preview/fontawesome/0.5.0/README.md
+++ b/packages/preview/fontawesome/0.5.0/README.md
@@ -95,10 +95,10 @@ Any feedback is appreciated.
 ##### Enable Pro sets
 
 Typst 0.12.0 raise a warning when the font is not found.
-To use the Pro set, `typst #fa_use_pro()` should be called before any `fa-*` functions.
+To use the Pro set, `#fa-use-pro()` should be called before any `fa-*` functions.
 
 ```typst
-#fa_use_pro()                 // Enable Pro sets
+#fa-use-pro()                 // Enable Pro sets
 
 #fa-icon("chess-queen-piece") // Use icons from Pro sets
 ```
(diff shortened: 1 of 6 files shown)


duskmoon314:
Where can I see the rules of code styles?

duskmoon314:
Oh, I found it in README:

> Style: No specific code style is mandated, but two spaces of indent and kebab-case for variable and function names are recommended.

packages/preview/fontawesome/0.5.0/README.md
```diff
@@ -0,0 +1,162 @@
...
+To use other sets or specify one set, you can pass the `font` parameter to the inner `text` function: \
+`fa-icon("github", font: "Font Awesome 6 Pro Solid")`
+
+If you have Font Awesome Pro, please help me test the library with the Pro set.
+Any feedback is appreciated.
+
+##### Enable Pro sets
+
+Typst 0.12.0 raise a warning when the font is not found.
+To use the Pro set, `typst #fa_use_pro()` should be called before any `fa-*` functions.
```
elegaanz, line 98: If you want to have Typst syntax highlighting, I think the correct syntax is with three backticks. Also, the function should be renamed here too.

duskmoon314: Oh, this is a copy-paste mistake.

Thanks for helping to find the remaining unchanged function names. A commit will be pushed ASAP

elegaanz:
Thanks :)

PR #1126, opened 2024-10-20
weave:0.1.0 and weave:0.2.0

leana8959:
I am submitting
- [x] a new package
- [x] an update for a package

Description:
Rewrote library functions to work with `.with()` application, making the library more versatile. Is a breaking change.

PS: I also corrected my name in the license. I simply copied someone else's license and didn't pay attention to that last time.

### DIFF
--- a/packages/preview/weave/0.1.0/LICENSE
+++ b/packages/preview/weave/0.1.0/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2024 Jason Daniel Pieck
+Copyright (c) 2024 Léana 江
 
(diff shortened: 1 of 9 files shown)


elegaanz:
Would it be possible to not exclude the README and LICENSE files actually? They are not necessary for the package to work correctly, but may still be useful to have in the package archive (for instance if a language server wants to display your README when showing docs for your package, or if a tool wants to check for license compatibility automatically).

leana8959:
Sure of course!
I misread the guidelines in the README, I thought it says we should exclude LICENSE and README.

elegaanz:
It is valid, the CI is just broken for some reason. Sorry about that.

PR #1132, opened 2024-10-21
bubble:0.2.2

hzkonor:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fix hardcoded language in the template and add some minor features (new `other` field)


elegaanz:
It would be nice if you could fix the error reported in CI. You can ignore the warnings.

PR #1142, opened 2024-10-22
acrostiche:0.3.4

Grisely:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This update mainly removes the dependency on the `display` function to print a state. It replaces it with `get` for accessing the value and moves the test on the value to outside. Overall, these changes make the code simpler.

This update also documents a useless feature for defining acronyms with content blocks instead of strings. It's futile but fun.


elegaanz:
Thanks :)

PR #1143, opened 2024-10-22
orange-book:0.3.0

flavio20002:
I am submitting
- [ ] a new package
- [x] an update for a package

Migrate template to Typst 0.12.0 and minor fixes.


elegaanz:
Thanks!

PR #1146, opened 2024-10-22
fix(pro-letter): update default font for 0.12 compatibility

swaits:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Updates default font for typst 0.12. Bumps version to 0.1.1; everything is exactly the same except for the version change in `typst.toml` and the font update in `lib.typ`.


swaits:
Note the warning on the 0.1.0 version is about the default font. I'm assuming we shouldn't modify any functionality of published packages, so I'm leaving that alone for now. If you prefer I change the font there, too, glad to update it.

elegaanz:
Yes, you're right. Thank you for the update.

PR #1150, opened 2024-10-22
ctheorems:1.1.3

sahasatvik:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed alignment and block-breaking issues resulting from breaking changes in Typst 0.12.


sahasatvik:
I have switched to kebab case in my unreleased [v2.0.0](https://github.com/sahasatvik/typst-theorems/tree/v2.0.0) branch; since meant to be a quick update fixing some breaking changes, I don't think it merits refactoring everything right now.

Does `exclude = ["/assets/*"]` not exclude the pdf's there? The checks seem to complain about `assets/manual_examples.pdf`.

elegaanz:
The way you wrote `exclude` works, but there is a bug in CI, sorry about that. And no problem for kebab-case.

PR #1153, opened 2024-10-22
minimal-presentation:0.2.0

flavio20002:
I am submitting
- [ ] a new package
- [X] an update for a package

Migrate to typst 0.12.0


packages/preview/minimal-presentation/0.2.0/template/main.typ
```diff
@@ -0,0 +1,109 @@
+#import "@preview/minimal-presentation:0.1.0": *
```
elegaanz, line 1: You should update that import, it should fix all the other warnings, that come from the older version of the package (they are just reported in the wrong place for some reason).

```suggestion
#import "@preview/minimal-presentation:0.2.0": *
```

flavio20002:
Done, thanks!

elegaanz:
Thank you :)

PR #1154, opened 2024-10-22
formalettre:0.1.1

Brndan:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Make signature implementation more robust.


elegaanz:
Thanks!

PR #1160, opened 2024-10-23
glossy:0.1.0 and glossy:0.1.1

swaits:
I am submitting
- [x] a new package
- [x] an update for a package

Description: Cleanup, refactoring, bug fix.

0.1.0: remove errant PDF

0.1.1: new version

- removed errant PDF
- cleaned up logic in term selection
- improved the example (readme.typ)
- unified around terms with no groups having an empty group string: ""


elegaanz:
Thanks

PR #1162, opened 2024-10-23
quill:0.5.0

Mc-Zen:
I am submitting
- [x] an update for a package

This release 
- adds support for multi-controlled gates with Tequila. 
- switches to using `context` instead of the now deprecated `style()` for measurement. 

Note: Starting with this version, Typst 0.11.0 or higher is required.

### DIFF
--- a/packages/preview/quill/0.5.0/LICENSE
+++ b/packages/preview/quill/0.5.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Mc-Zen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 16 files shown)


Mc-Zen:
Wow, the bot is so picky 😅

The gate `targX` has always been named like that and I'd rather not change it. Actually, it is not even used anymore and can always be replaced by `swap(0)`. It sort of just sticks around, maybe to help LaTeX users find their symbols. I might remove this altogether in the next release.

elegaanz:
Yeah the bot can be picky, especially with kebab-case. But it is not a requirement to have your package accepted.

Anyway, thanks for the update!

Mc-Zen:
Good, I'm glad 😅

Thank you

PR #1165, opened 2024-10-24
pesha:0.4.0

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

### Added

- `page-numbering-format` option so that end users can specify the numbering format pattern.
- Keywords to document metadata: author name, `curriculum vitae`, `cv`, `resume`.

### Changed

- Minimum required compiler version to 0.12.0.
- Decreased page margins by 1%.
- Decreased font size for experience time by 0.1em.

### Fixed

- Footer text not using correct styling for single page documents.

---


talal:
I just tested the template in this PR with Typst web app and everything works as expected. Cantarell font is still available in the web app (see screenshot). Due to the warning by the GH workflow, I thought it might've been removed in the recent 0.12.0 update.

image

elegaanz:
Don't worry, it's just that the bot only supports the fonts embedded in the compiler, not all the ones available in the web app. It would silently fallback to another font in previous versions, but since 0.12 it reports a warning. But everything will actually work as expected if people use your template in the web app (or if they have Cantarell installed on their machine).

PR #1167, opened 2024-10-24
grotesk-cv:0.1.0

AsiSkarp:
Remove non-functional pre-release versions

I am submitting
- [x] a new package
- [ ] an update for a package

Deleting a non-functional pre-release package


elegaanz:
Hello, we don't delete packages from Universe, even broken ones. It means that people can technically install an old broken version, but in practice nobody will do it or they will notice very quickly and update. I'm closing this PR and all the others you opened.

AsiSkarp:
All right, that is nice to know!

PR #1182, opened 2024-10-26
zero:0.3.0

Mc-Zen:
I am submitting
- [x] an update for a package

- Adds a `nonum[]` function that can be used to mark content in cells as _not belonging to the number_. The remaining content may still be recognized as a number and formatted/aligned accordingly. The content wrapped by `nonum[]` is preserved. 
- Fixes number alignment tables with new version Typst 0.12. Apparently `context` for measurements is now not anymore provided by default within show rules


elegaanz:
Thanks!

PR #1184, opened 2024-10-26
README: Fix local packages path for Windows

DanikVitek:


EpicEricEE:
`%APPDATA%` already points to the `Roaming` directory, so it was correct before.

PR #1196, opened 2024-10-29
fh-joanneum-iit-thesis:2.0.3

internettechnik:
I am submitting
- [ ] a new package
- [x] an update for a package

Updated to latest typst compiler version. Fixed warning about missing font.


elegaanz:
If you want you can restore the STIX Two Text font. The warning is a false positive that only happens in CI. If the users of your package correctly install the font, they should never encounter it.

internettechnik:
No problem at all, we have no need to use STIX Two Text font. So we removed it and get rid of the warning :)

PR #1220, opened 2024-11-02
ilm:1.3.0

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

- Use spacing attribute for paragraph spacing instead of block.
- Minimum required compiler version to 0.12.0.


elegaanz:
Thanks :)

PR #1229, opened 2024-11-04
Test PR Workflow with sparse checkout

quachpas:
# Add a new Test workflow for PR specifically using sparse checkout

Based on the PR title in format `[package]:[version]`, the checkout action
uses `sparse-checkout` to reduce the scope of the test for each PR.

~~Example run at:~~
~~- single package: https://github.com/quachpas/packages/actions/runs/11666315963/job/32480980631~~
~~- multiple package: https://github.com/quachpas/packages/actions/runs/11666311358/job/32480966073~~

Test for PR are separated in a different workflow file `test_pr.yml`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -13,7 +13,7 @@ jobs:
     name: Deploy
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           # Download the whole history for the package timestamps
           fetch-depth: 0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     name: Test
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
         with:
--- a/.github/workflows/test_pr.yml
+++ b/.github/workflows/test_pr.yml
@@ -34,7 +34,7 @@ jobs:
             }
             core.setOutput('sparse_checkout_paths', sparseCheckoutPaths);
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           sparse-checkout: |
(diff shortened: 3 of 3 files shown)


elegaanz:
I'm not sure how well this would work, knowing that some people open PRs with title that don't follow this format, and they only get renamed by the review bot once opened.
Also, for PRs with multiple packages, all the package names get listed in the PR title, so that should be taken into account.

quachpas:
> they only get renamed by the review bot once opened

ah this was one of my interrogations, the PR naming seemed awfully uniform

> Also, for PRs with multiple packages, all the package names get listed in the PR title, so that should be taken into account.

I can do that I think

quachpas:
> they only get renamed by the review bot once opened.

From what I've seen, the bot seems to be "first" at renaming PRs?

I've added the capability to sparse-checkout multiple packages

elegaanz:
> I've added the capability to sparse-checkout multiple packages

Thanks :)

> From what I've seen, the bot seems to be "first" at renaming PRs?

It does it quite quickly once the PR is opened, but I think the CI job is started just before (it just doesn't show in the PR timeline when it does). If GitHub lazily evaluates the `context` object when you read it, it will probably get the new PR title, but I doubt it does, I would tend to believe it builds it once before starting the CI job and keeps it as it is afterwards.

Maybe there is a better way to tell which packages were introduced in a given PR than relying on its title? For instance, the review bot asks Git what files changed to be able to tell and then rename PRs accordingly, GitHub probably offers some mechanism to query this info (the Git command-line and a little bit of shell scripting could also work).

quachpas:
How about relying on `github.rest.pulls.listFiles`? It's octokit.rest.

https://github.com/quachpas/packages/actions/runs/11667822857/job/32486031946
https://github.com/quachpas/packages/actions/runs/11667805686/job/32485974936

quachpas:
I set it to "packages/" by default, "/" cannot be used.

quachpas:
BTW, is there a reason why checkout/v3 is used instead of v4?

elegaanz:
No particular reason, no. I just think it was the latest version at the time CI was set up. You can update it in this PR if you want. Otherwise, it looks good to me, I will merge.

quachpas:
Should be good!

elegaanz:
Thanks a lot!

PR #1232, opened 2024-11-04
update abbr:0.1.0

jhalmen:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

this just updates the readme with contributing information and an example output the package can produce. Hence it doesn't even get a version bump


elegaanz:
Thank you.

PR #1233, opened 2024-11-05
tinyset:0.1.0

SylvanFranklin:
I am submitting
- [x] a new package
- [ ] an update for a package

Tinyset is a minimal and opinionated library that provides a handful of utility functions for writing pure math pros for homework / psets. It contains a prebuilt header along with several environments for proof / question / sub question etc. I also added some custom plain English shorthand for common math symbols such as implication. 


SylvanFranklin:
Hullo! I read the warning regarding kebab case but did not realize that this name would trigger it. I could update the name if that would better, but I don't think of the two words as separate, but rather as a hybrid name. Thanks.

packages/preview/tinyset/0.1.0/lib.typ
```diff
@@ -0,0 +1,31 @@
...
+#let header(number: 1, class: "Math 3551 - Fall 2024", name: "Your Name") = {
+  set text(font: "New Computer Modern Math")
+  align(center)[
+    #class \
+    #name - Homework #number
+  ]
+  set par(justify: true)
+}
+
+// custom shorthands, feel free to comment these out if they are in the way.
```
elegaanz, line 11: For information it is not possible for package users to edit the contents of a package.

packages/preview/tinyset/0.1.0/lib.typ
```diff
@@ -0,0 +1,31 @@
+/// You must create a header to get the rest of the style
+#let header(number: 1, class: "Math 3551 - Fall 2024", name: "Your Name") = {
+  set text(font: "New Computer Modern Math")
+  align(center)[
+    #class \
+    #name - Homework #number
+  ]
+  set par(justify: true)
```
elegaanz, line 8: I don't think that will work as expected, since `set` rules are scoped. This would only apply until the end of the current scope, which is on the next line. If you want to apply this rule to the whole document, the correct way to do it is to use a global `show` rule. There is [a tutorial on how to create one](https://typst.app/docs/tutorial/making-a-template/).

elegaanz (commented):
For the warning, don't worry about it, kebab-case is not mandatory in any case, and here it makes sense to keep the capital letter.

packages/preview/tinyset/0.1.0/lib.typ
SylvanFranklin: Yes of course! Thanks for pointing that out, I will move the show rules into the individual components, just for ease of use of the library.

SylvanFranklin:
Thank you for being so friendly and welcoming to a beginner! I'm glad the the Typst foundation is in such good hands, this really made my day!

PR #1234, opened 2024-11-06
based:0.2.0

EpicEricEE:
I am submitting
- [ ] a new package
- [x] an update for a package

**Description:** Switched from pure Typst implementation to WASM, as performance for larger data (e.g. images) was terrible. This now uses the rust crates [`base64`](https://crates.io/crates/base64), [`base32`](https://crates.io/crates/base32), and [`hex`](https://crates.io/crates/hex).


elegaanz:
Thank you.

PR #1237, opened 2024-11-06
vienna-tech:0.1.2

npikall:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added new customization functionality and adapted to the newest typst version


elegaanz:
Thanks :)

PR #1240, opened 2024-11-08
scholarly-tauthesis:0.9.0

SeSodesa:
I am submitting
- [ ] a new package
- [x] an update for a package

A thesis template for Tampere University Bachelor's and Master's theses.

In this update, a fatal error with the typst packaging system was mended. Basically, the template module (the file `tauthesis.typ`) was including files like `meta.typ` and the abstracts relative to itself, which of course resulted in users' changes not showing up in their work on typst.app or when initializing a package with `typst init`. This was not detected before, because to make testing not related to packaging easier, symbolic links to the Git repo were generated in the local Typst package cache directory, but an actual `typst init` was omitted from local testing, at least for the past few releases.

The fix is to give all user information and file locations in `main.typ`. This makes the implementation leak a bit, but it seems unavoidable.

Also fixes an index-out-of-range error with `preamble.num`. Although, now that the `zero` package is available, it might be recommended instead.


SeSodesa:
Looks like I am supposedly doing too many things again. Last time it was a false alert, so I guess that is the case this time as well.

elegaanz:
Yes, it is because of [this issue](https://github.com/typst/package-check/issues/14) and the `tiivistelmä.typ` file. So as long as you have this file in your package you can expect this error (until the issue is fixed). :sweat_smile: I tested locally and a few warnings are reported, but nothing that prevent the new version from being accepted.

PR #1242, opened 2024-11-09
bamdone-rebuttal:0.1.1

avonmoll:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Changed quote formatting to fix the problem that all child blocks were getting styled.


elegaanz:
Thanks!

PR #1252, opened 2024-11-11
package/unofficial-fhict-document-template:1.1.1

TomVer99:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: A bug in the front glossary option was fixed in this minor version. It made the front glossay w/ groups unusable so I created this minor version with a fix.


elegaanz:
Thanks :)

PR #1254, opened 2024-11-11
clean-math-thesis:0.1.0

sebaseb98:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: After finishing my master's thesis, I thought it could be useful to create a template from it. I did not theme it to a particular university but created a general math template which uses `great-theorems` and also configures it in a useful way.  Additionally, I included my equation settings, which also took me some time to find when writing my thesis. I hope to make life easier for people writing a mathematical thesis in the future.

I am not entirely sure about this last point, but as I have seen other template packages under the MIT license I guess it should be alright.  

Thanks for processing my template, let me know if anything needs to be changed :)


elegaanz:
Thank you!

PR #1259, opened 2024-11-12
slydst:0.1.2

glambrechts:
I am submitting
- [x] an update for a package

Description: Some deprecation warnings were fixed (resulting from the new Typst version), there are also more customisation options, and I added a default template, which could be useful since it is a package for slides.


elegaanz:
Thank you.

elegaanz:
Sorry, I merged a bit too fast, and I didn't notice but the thumbnail file should not be part of the template sub-directory, otherwise it will be copied to every new project created from this template. Could you please make a `0.1.3` release to fix that?

glambrechts:
Can't I just modify version 0.1.2?

elegaanz:
We generally don't allow modifying a package after it was published, unless you just want to update some metadata and not touch the code.

glambrechts:
No problem, I will submit a new PR then.

PR #1264, opened 2024-11-12
slydst:0.1.3

glambrechts:
I am submitting
- [x] an update for a package

Description: Moved the thumbnail out of the template (as requested in https://github.com/typst/packages/pull/1259). Also, fixed a typo in `lib.typ`.


elegaanz:
Thanks!

PR #1269, opened 2024-11-13
derive-it:0.1.1

0rphee:
Fix examples in README.md

I am submitting
- [ ] a new package
- [x] an update for a package

Changes: only an update in the README.md, the first version (0.1.1) had broken examples (because of a change in camel case -> kebab case. Sorry for the trouble...


elegaanz:
I'll merge this as a new version to avoid you some more work, but for info it's okay to edit an older release in case you are just fixing the README.

PR #1272, opened 2024-11-14
quick-maths:0.2.0

EpicEricEE:
I am submitting
- [ ] a new package
- [x] an update for a package

**Description:** Added more cursed behavior, namely
- Single-element shorthands are now applied to non-sequence elements (via dynamically created show rules).
- Improved interaction of shorthands with attachments and fractions. For example, a shorthand `+-` is now correctly detected in the cases `+- / 2`, `1 / +-` or `+-^2` (but not in the case `x^+-`) even though Typst would normally require parentheses.


elegaanz:
Thank you. :)

PR #1273, opened 2024-11-14
minimal-presentation:0.3.0

flavio20002:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: A fix on figure caption size that impacted on text size inside the figure itself.

- [X ] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.


elegaanz:
Thanks :)

PR #1297, opened 2024-11-19
ilm:1.4.0

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

# Changelog
Added:
- `appendix` option for defining appendix.
- Documentation of template to template's `main.typ` file so that users of Typst's web app
  can see how the template works and how to customize it without leaving the web app, i.e.
  they don't have to read the README on GitHub.

---


elegaanz:
Thank you.

PR #1300, opened 2024-11-20
accelerated-jacow:0.1.1

eltos:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- Place title footnotes (funding, author email) in left column only
- Automatically number labelled equations (no need for `eqnum` any more)
- Add shorthands for author list definition
- Avoid hyphenations and undesired breaks in author list, avoid hyphenations in headings
- Add helper for column balancing on last page


elegaanz:
Thank you.

PR #1308, opened 2024-11-21
touying-flow:1.0.0 and touying-ustc:0.1.0

Quaternijkon:
I am submitting
- [x] a new package
- [ ] an update for a package

This package is a template file created based on the Touying project, with simplicity and practicality as its primary features. It integrates a variety of functions, making it convenient for users to learn and use Typst to create presentations. It is particularly well-suited for academic exchanges in settings such as universities.


Quaternijkon:
> Hello, I think your package need to be reworked a little bit. From what I understand, they both include a copy of an old version of touying, which I believe is not the correct way to define themes. You can have a look at what [`touying-brandred-uobristol`]'https://github.com/typst/packages/tree/main/packages/preview/touying-brandred-uobristol/0.1.2) does. If I'm mistaken and there is a good reason for copying all of touying, please tell me.

At first, I felt that the styles of components in Touying did not align with my theme, so I referred to Touying's source code and made some modifications. For convenience, I copied some of Touying's original components, but certain components have been implemented differently from Touying. If this structure is not considered best practice, I will replace them in future updates. By the way, the USTC project does not include packages from Touying.

Quaternijkon:
@elegaanz 
I have checked, and the components I copied from Touying Dewdrop are not the same as those imported via `import touying`. For example, the `side-by-side` implementation is not available in the imported Touying package but is implemented in the Dewdrop theme. I wanted to customize these components to better adapt them to my template, so I copied some components from Touying Dewdrop. Although they are all originally created by Touying, they cannot be achieved through `import @preview/touying:xyz`. 

I have copied more component implementations than I currently use because I hope to gradually adapt them to my theme in the future while reserving appropriate interfaces in advance. To be honest, these components are not large in size, but they are definitely not redundant parts of the Touying package.

OrangeX4:
First of all, thank you for your contribution! However, I don't think you should fork touying now, because it will cause you to not be able to keep up with touying's updates in the future. Touying doesn't make assumptions about how you should write your layout functions, and it ensures the decoupling of various interfaces. You can just copy a small part of it and update it with `config-xxx()`; Or you can use the built-in layout functions using syntax like `components.side-by-side`. If you think touying should add or change any part of the code, you are welcome to open an issue or PR, and you are welcome to join the touying-typ organization. If you just need to write a new theme or template, you can also refer to the existing touying themes and templates, I believe they can inspire you.

Quaternijkon:
@OrangeX4 thanks for your reply, I will refactor it later.

PR #1313, opened 2024-11-22
typslides:1.2.1

manjavacas:
I am submitting
- [x] an update for a package

This release includes several bug fixes and some new features.

- Corrected the spacing of `grayed` boxes.
- Included the `blank-slide`, which does not include header bar vs. no-titled slides.
- The background of all slides is now white-filled. Some were previously defined as transparent.
- Modified the separations between slides' text and header.
- Fixed bug that prevented correct text alignment. This is due to text wrapping inside a `box`. It has been corrected returning to the previous version format. Included changes in the header (now in the background).
- Docs and README updated.


elegaanz:
Thanks.

PR #1319, opened 2024-11-23
marginalia:0.1.1

nleanba:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: The package provides simple utilities to create margin-notes and wide blocks that extend into the margin. It provides a simple way to set this up such that everything is aligned nicely.

Updates:
- Notes will now avoid each other for any amount of notes.
- They will also avoid wide blocks and the page margin.
- Added documentation for figures and improved compatibility.
- Added `notefigure` command for figures in the margin column.
- Added `clearance` configuration option

### DIFF
--- a/packages/preview/marginalia/0.1.1/README.md
+++ b/packages/preview/marginalia/0.1.1/README.md
@@ -1,0 +1,78 @@
+# Marginalia
+
+## Setup
+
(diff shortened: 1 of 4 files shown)


elegaanz:
Thank you!

PR #1328, opened 2024-11-25
wenyuan-campaign:0.1.0

yanwenywan:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package recreates the LaTeX DnD 5e-style campaign document for Typst, for all game masters who want to write (or publish!) their campaign documents in a format like published campaigns, and for reasons prefer to use typst or do not want to go through the trouble of manually installing a LaTeX package, dealing with long compile times, etc.


yanwenywan:
Not quite sure why the package check is bringing up "This PR does too many things", please advise if there's any issues, thanks

packages/preview/dndcampaign/0.1.0/typst.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "dndcampaign"
```
elegaanz, line 2: Could you please change the name to something less canonical / a bit more unique? Other people might want to submit templates for the same purpose in the future.

packages/preview/dndcampaign/0.1.0/typst.toml
```diff
@@ -0,0 +1,18 @@
...
+version = "0.1.0"
+entrypoint = "campaign.typ"
+authors = ["Yan Xin <@yanwenywan>"]
+description = "Easily write DnD 5e style campaign documents."
+categories = ["layout", "text", "book"]
+license = "Apache-2.0"
+keywords = ["rpg", "5e", "dnd", "campaign"]
+repository = "https://github.com/yanwenywan/typst-packages/tree/master/dndcampaign"
+exclude = ["**/*.pdf", "template/fonts/*"]
```
elegaanz, line 11: The template files should not be excluded: if they are they won't be in the package archive and won't be copied in the end. However in that case, because the fonts files are quite large, I think it would just be best to not include them in the template at all, and give people instructions on how to download them and use them with Typst (in your README and/or in a comment of `template/main.typ`)

elegaanz (commented):
Thank you for the nice template!

Sorry about the CI bug, it is a [known](https://github.com/typst/package-check/issues/14). I did a manual review instead.

For info, it is more common in the Typst ecosystem to use kebab-case names for functions and their arguments, so you may want to change that (but you can also keep camelCase if you prefer).

yanwenywan:
Thanks for you feedback. It's no problem to change the names, I don't think I ever realised Typst allows kebab-case until recently lol

yanwenywan:
Okay, it's still unhappy about a camelCase name, but that's not really part of the intended public interface 😅 everything that's meant to be used by a user should be now kebab case

elegaanz:
Thank you for renaming the package and fixing the issues.

FYI, you can silence warnings about kebab-case by prefixing names with a `_`. The item would still be accessible, be it is clearer that it is supposed to be private. If you want it to not be public at all, the correct solution is to have an `entrypoint` file that only imports items that should be public from other files. I'm not suggesting you to do that here, just giving you the information so that you have it next time.

> Also, sorry about the few last straggler bugs and features that need to be added in before merge

Please tell me when you're ready for another review then :)

yanwenywan:
Thanks for your advice. I've gone through the package once more and ported one of my own small campaigns to typst, I can't see any more problems so it's probably good to release to the wild

Ready for review whenever you are.

elegaanz:
Great, everything looks good to me, thanks!

PR #1337, opened 2024-11-27
unify:0.7.0

ChHecker:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- added option to add units at runtime (finally!)
- added new per mode (inline units, e.g. m/s)
- smaller fixes


elegaanz:
Thanks!

PR #1342, opened 2024-11-28
echarm:0.1.1

lublak:
I am submitting
- [ ] a new package
- [x] an update for a package

just updated to the latext ctxjs version to add eval-later support


lublak:
yes the current version of echarm is a bit bigger than 2MB
Please publish anyway, as the function of eval-later has been improved here.
I will definitely analyze what I can do to make the kbc1 files smaller again. But that will take a long time (months I think).

elegaanz:
Don't worry it's fine, this warning is here in case you pushed files that are large and not strictly necessary for the package to work (for instance a big example document). Thank you for submitting the update :)

ramkumarkb:
@elegaanz & @lublak - thank you so much for this update 🎉🙏✅

PR #1346, opened 2024-11-29
add `.history` exclusion to .gitignore

neuralpain:
Add exclusion for the automatic `.history` folder generated by the Local History VSCode extension (xyz.local-history).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ desktop.ini
 
 # Rust
 /bundler/target
+
+# Local History (xyz.local-history)
+.history/


elegaanz:
I'm a bit split about this: on one side it's really a small change and it wouldn't hurt to add that line, and on the other hand it seems to be very specific and I feel like it should be the responsibility of this extension to not fill all the projects you touch with its internal state, and [a lot of people seem to agree with that](https://github.com/zabel-xyz/local-history/issues/129).

I think a better solution would be to add this to your personal VS Code config:

```json5
{
  "local-history.path": "~/.cache/vscode-local-history", // or some other global directory
  "local-history.absolute": false, 
}
```

neuralpain:
> I feel like it should be the responsibility of this extension to not fill all the projects you touch with its internal state

Oh yeah, I totally agree with you.

> I think a better solution would be to add this to your personal VS Code config:

The settings worked so I'll leave it as that for now. Thanks.

PgBiel:
I'll add that, as someone who often dumps lots of testing stuff in repositories I'm working on, you can add `.history`, or any other file you'd like to locally ignore, to `.git/info/exclude` in the repository. It ensures the file won't be tracked, but only for you, without having to change `.gitignore`.

PR #1348, opened 2024-11-30
basic-resume:0.2.1

stuxf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Support orcid as a social link. Also fix contact-item's prefix argument.


Moelf:
TIL this whole process is manual, huh

elegaanz:
Thanks :)

PR #1356, opened 2024-11-30
codly-languages:0.1.2

swaits:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Version 0.1.2:

- just make #codly-languages return the full dictionary instead of
  trying to add the aliases dynamically
- add test.typ
- ignore `**/test*.pdf`


elegaanz:
Thanks!

PR #1360, opened 2024-11-30
modern-hsh-thesis:1.0.1

MrToWy:
I am submitting an update for a package

## Added
- Optional Parameter `chapter-break-mode` to control whether a new chapter should start on a new page.

## Changed
- Upgraded Typst to version 0.12
- Upgraded Hydra to 0.5.1

## Removed
- get-current-heading (instead use get-current-heading-hydra)

## Misc
- Added "Just" install step to readme


elegaanz:
Thanks :)

PR #1364, opened 2024-12-01
oasis-align:0.2.0

jdpieck:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changes
- removed `oasis-align-images`  
- added `swap` to parameters
    - switches positions of content
- cleaned up `debug` output to be more readable
- breaking scenarios are now treated as such with `panic`
- no longer show error message if content is "incompatible" and will instead display state at `int-frac`
    - should make using `oasis-align` way less frustrating to use
- default max iterations reduced from 50 to 30

## License


elegaanz:
Thank you!

PR #1366, opened 2024-12-01
rfc-vibe:0.1.0

swaits:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Just updating tests.


elegaanz:
Does that require a new release? Is there any visible difference for the users of the package?

elegaanz:
Sorry, I think I wasn't clear. I was actually asking the opposite question of what you understood: is it worth it to make a new release now if the only change is related to test files? My reasoning is that users won't see the difference, so there is no point to update for them. I would have waited for more changes to make the `0.1.1` release (or `0.2.0` depending on how much you change).
But if you consider it is worth it, I can merge, no problem.

swaits:
@elegaanz - accidentally filled out the form wrong, so this is tagged as "new" instead of "update". Don't see how I can change the tags. Can you change it before merging, please?

swaits:
Ignore the above comment. It's correct now.

elegaanz:
Yes, the bot doesn't handle updates to previous versions correctly, but it's fine. Thanks for the update.

PR #1373, opened 2024-12-03
xyznote:0.2.0

wardenxyz:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: xyznote is a template designed for efficient and organized note-taking with Typst.


packages/preview/xyznote/0.2.0/template.typ
```diff
@@ -0,0 +1,265 @@
...
+  counter(page).update(1)
+
+  body
+
+  //参考文献
+  if bibliography-file != none {
+    set text(font: ("Times New Roman", "KaiTi")) //设置参考文献字体
+    pagebreak()
+    show bibliography: set text(10.5pt)
+    bibliography("template/refs.bib", style: bibliographystyle)
```
elegaanz, line 225: Users of your package won't be able to give their own bibliography file with that code. The `refs.bib` file in your package and in their project are not the same, and here you are referencing the one from the package, which people can't edit.
The proper approach here is to let users pass a `bibliography(...)` call as a parameter to your template, that you simply display.

```typst
#let xyznote(
  // ...
  bib: none
) {
  // ...
  if bib != none {
    // Add the required configuration as before
    set text(font: ("Times New Roman", "KaiTi"))
    pagebreak()
    show bibliography: set text(10.5pt)
    // Set the bibliography style with a set rule
    set bibliography(style: bibliographystyle)
    // Show the bibliography the user supplied
    #bib
  }
}
```

```typst
#show: xyznote.with(bib: bibliography("my-refs.bib"))
```

wardenxyz: Okay, I'll make some more changes.

elegaanz:
Thank you!

PR #1378, opened 2024-12-03
bamdone-aiaa:0.1.2

isaacew:
I am submitting
- [ ] a new package
- [x] an update for a package


isaacew:
I have a refs.bib file. I'm not sure what the issue is.

packages/preview/bamdone-aiaa/0.1.2/template/main.typ
```diff
@@ -0,0 +1,227 @@
...
+// Date: 03 DEC 2024
+// BAMDONE!
+//***************************************************************
+
+#import "@preview/bamdone-aiaa:0.1.2": *
+#import "@preview/droplet:0.3.1": dropcap
+
+#show: aiaa.with(
+    title: "Preparation of Papers for AIAA Technical Conferences",
+    bibliography-file: "refs.bib",
```
elegaanz, line 21: The proper way to fix this is to take a `bibliography(...)` call directly as a parameter here and not just the file (see [this package](https://github.com/typst/packages/blob/main/packages/preview/charged-ieee/0.1.3/template/main.typ#L25) if you need a reference). If you just give a path here, it will be resolved relative to the package root and not to the user's project (in general, it is resolved relative to where `bibliography` is called), so the only possible path will be `refs.bib` and users won't be able to edit the contents of that file.

isaacew: I was able to fix this per your guidance. Thanks!

elegaanz:
Thanks!

PR #1381, opened 2024-12-04
typslides:1.2.3

manjavacas:
I am submitting
- [x] an update for a package

This new version of Typslides includes the following changes:

- Added functions for colouring text with the different theme colours (i.e., `bluey`, `reddy`, `greey`, `yelly`, `purply`, `dusky`).
- Added `template/main.typ` and `thumbnail.png` to the project.
- Minor documentation changes.
- `bibliography-slide` is now usable via API (see https://github.com/manjavacas/typslides/issues/2, https://github.com/typst/packages/pull/1381#issuecomment-2516876444).
- Footnote references are now coloured.


elegaanz:
If you want to properly fix the bibliography issue, you can take a `bibliography(...)` call instead of just the path to the file as an argument. The file path is resolved relatively to where the `bibliography` call is made, that's why it has to be done in the user's project and not in your package (from which you can only reference other package files).

manjavacas:
> If you want to properly fix the bibliography issue, you can take a `bibliography(...)` call instead of just the path to the file as an argument. The file path is resolved relatively to where the `bibliography` call is made, that's why it has to be done in the user's project and not in your package (from which you can only reference other package files).

Thank you for the feedback @elegaanz! 😃 

So do you refer to something like this?

```typst
#let bib = bibliography("bibliography.bib")
#bibliography-slide(bib)
```

```typst
#let bibliography-slide(
  bib-call,
) = {
  set bibliography(title: text(size: 30pt)[References])
  bib-call
}
```

I have tested it locally and it seems to work. If you can confirm I will add it to this update 👍🏻

elegaanz:
Yes, exactly, thank you!

PR #1387, opened 2024-12-04
t4t:0.4.0

jneug:
I am submitting
- [ ] a new package
- [x] an update for a package

This new version has some breaking changes to some of the main submodules, but should make `t4t` future proof with some of the changes in upcoming Typst versions.

- :warning: Removed `alias` module in favor of the Typst 0.12 `std` module.
- :warning: Changed argument order in `def` module functions.
	- Old behaviour is still available in the `def.compat` module for now.
- :warning: Renamed `is` module to `test` and moved some tests to the top-level module.
	- `is` will become a reserved word in future Typst versions (see [typst/typst#5229](https://github.com/typst/typst/pull/5229))
- Refactored `get.stroke-paint` and `get.stroke-thickness` to use the native `stroke` type.
- Refactored `get.x-align` and `get.y-align` to use the native `align.x` and `align.y` values.


jneug:
`test.typ` is a module of the package and not a test. Should I rename it?

elegaanz:
No it's fine, thank you for the update!

PR #1389, opened 2024-12-04
mephistypsteles:0.2.0

T0mstone:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- Fixes:
  - `let` no longer has two `kind` fields (which caused an error before).
- Breaking changes:
  - The structure of parsed `ImportItem`s was changed.
  - The `parse-flat` and `parse-markup` functions got unified into `parse`.
- Additions:
  - Expressions and flat nodes now have a `span` field describing their location in the source string.


elegaanz:
Thank you.

PR #1390, opened 2024-12-04
bamdone-ieeeconf:0.1.0

isaacew:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: The charged-ieee.typ package does not conform to the IEEE's conference style. This one does.


jamesrswift:
Would it not be preferable to make a PR to the other package?

isaacew:
I feel they are different. And it appears that ieee has two separate formats that accept. One is IEEEtrans which is the one that charged-ieee uses. And the one that I have provided which is ieeeconf. They should both exist.

-Isaac
________________________________

elegaanz:
Thanks!

PR #1395, opened 2024-12-06
finite:0.4.0

jneug:
I am submitting
- [ ] a new package
- [x] an update for a package

### Version 0.4.0

:warning: Version 0.4.0 is a major rewrite of finite to make it compatible with Typst 0.12 and CeTZ 0.3.0. This includes some minor breaking changes, mostly in how layouts work, but overall functionality should be the same.


packages/preview/finite/0.4.0/CHANGELOG.md
elegaanz, line 1: Did you forget to update the changelog? And if not, maybe it would be better to just not include it, as it doesn't seem very helpful in its current state.

packages/preview/finite/0.4.0/tests/test-1/test.typ
elegaanz, line 1: I'm not sure how typst-test works, but I think you should either remove this file (and the corresponding reference image) or add some actual tests for your package in here.

elegaanz (commented):
Just some remarks about files that could be deleted (or improved) in my opinion, but otherwise it looks good!

packages/preview/finite/0.4.0/CHANGELOG.md
jneug: True, I migrated to the typst-package-template and never moved the changelog out of the readme. Did that now.

packages/preview/finite/0.4.0/tests/test-1/test.typ
jneug: Somehow, I had a bad version of typst-test installed that didn't work properly. I want to add more tests in the future, but I pushed the first few to the repo now. Or should I remove tests in general before creating a pull request? Don't know how other packages do this.

jneug: Scratch that, it probably makes sense to just submit the files here, that are actually needed by Typst. I removed the tests and scripts altogether.

elegaanz:
Thanks :)

PR #1400, opened 2024-12-07
flautomat:0.1.0

Kuchenmampfer:
I am submitting
- [x] a new package
- [ ] an update for a package

This package can be used to visualize abstract automata. It relies on the json format that [FLACI](https://flaci.com/autoedit) ([FLACI on github](https://github.com/gitmh/FLACI/tree/main)) saves to. This allows for intuitive automata design with FLACI and then an import into Typst without relying on raster graphics. Therefore the package can be useful for anyone who studies abstract automata and wants to visualize them.

### DIFF
--- a/packages/preview/flautomat/0.1.0/README.md
+++ b/packages/preview/flautomat/0.1.0/README.md
@@ -1,0 +1,44 @@
+# Typst package for the visualization of abstract automata
+
+This package can be used to visualize abstract automata in [Typst](https://typst.app/docs/) using [fletcher](https://typst.app/universe/package/fletcher). It relies on the json format that [FLACI](https://flaci.com/autoedit) ([FLACI on github](https://github.com/gitmh/FLACI/tree/main)) saves to. This allows you to intuitively design your automata with FLACI and then import them into Typst without relying on raster graphics.
+
+## Usage
+
+- Create an automaton via [FLACI](https://flaci.com/autoedit) or one of their local clients and save it locally to json.
+- Sadly, the exported json starts with a weird character that prevents us from directly importing it. You can remove that char for example with the following command line:
+```bash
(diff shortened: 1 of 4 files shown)


jneug:
Nice package. I was looking into supporting more of FLACI in `finite` myself. 

One suggestion: I think you can get around the json read problem by first reading the file and then decoding it with `json.decode`. In my tests it worked fine:

```typst
#flautomat(json.decode(read("Automaton_TEIL2_ Wort mit gerader Anzahl an Nullen.json")), scaling_factor: 1)
```

packages/preview/flautomat/0.1.0/README.md
```diff
@@ -0,0 +1,44 @@
...
+```bash
+cat Automaton_MY_AUTOMAT.json | jq . > MY_AUTOMAT.json
+```
+- Now import this package into your typst file and use it:
+```typst
+#import "@preview/flautomat:0.1.0": flautomat
+
+...
+
+#flautomat(json("MY_AUTOMAT.json"), scaling_factor: 1)
```
jneug, line 18: Should be `scaling-factor`.

Kuchenmampfer: Right, thank you

Kuchenmampfer:
Thank you :)

I have to admit that I only got aware of `finite` after I made this package. I wondered if a new package was really necessary or if I should make a PR to `finite` to add FLACI-support. But I think our architecture is too different to easily translate. And I now also think that having one powerful and one simple and easy to use package is a good thing.

I tried your approach with reading first, but for me it did not work :/ So I guess I will leave this part in the README.

jneug:
I had a similar experience with Fletcher. It came out right after I published `finite`. Otherwise I probably would have build it on top of Fletcher. 

I think the packages are different enough and having a choice is always a good thing. 😀

elegaanz:
Thanks!

PR #1412, opened 2024-12-09
zero:0.3.1

Mc-Zen:
I am submitting
- [x] an update for a package

- Improved support for `math: false` mode: numbers can now contain uncertainties and exponents without the use of `math.equation`. 
- More selective number recognition in tables: Numerals need to start with one of `0123456789+-.,` to be automatically recognized as a number (⚠️Breaking change). 
- Fixes show rules on tables: it is now possible to apply rules like
  ```typ
    #show table.cell.where(y: 1): strong
    #show table.cell.where(y: 1): text.with(blue)
  ```
  and have it also affect cells where automatic number recognition and alignment is applied.

### DIFF
--- a/packages/preview/zero/0.3.1/LICENSE
+++ b/packages/preview/zero/0.3.1/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Mc-Zen
+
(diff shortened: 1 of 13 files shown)


elegaanz:
Thank you.

PR #1414, opened 2024-12-09
bamdone-ieeeconf:0.1.1

isaacew:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed the figure numbering issue where it starts at 2 instead of 1


elegaanz:
Thank you.

PR #1418, opened 2024-12-11
acrostiche:0.5.0

Grisely:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Add some features inspired from the `acronym` package from LaTeX. Fix some bugs. Refactor code a bit to access short version of acronyms in all cases.


elegaanz:
Thanks!

PR #1420, opened 2024-12-11
diatypst:0.4.0

skriptum:
I am submitting
- [ ] a new package
- [x] an update for a package

- add a new theme to display slides (full theme, more similar to a beamer look)
- fix some minor errors in display
- custom footer title / footer subtite option

New Full Theme
Full-Example-Slide


skriptum:
also a question to the maintainers: is there a description of the distinction between package and template? diatypst is coming to be more of a package i would say, with a lot of different options / themes / useful tricks etc. 
At what point should I change it to a package / does that need to fullfull certain things? merci

elegaanz:
All templates are packages, but the reverse is not true: a template is just a package with a special directory that can be copied to a new project easily with the command line or the web app. So `diatypst` already is a package, that includes a template.
If you feel like the template is no longer relevant, you can delete it in the next release (or even in this one). If you feel like you could offer various different templates (one per theme for instance), the best solution for now is to make one package for each of these options. The package itself would be very small, maybe re-exporting the main `diatypst` package and the theme information, and the template would contain a Typst file that imports that theme and uses it.
I hope this makes things clearer. If not feel free to ask more specific questions.

If you want to change something in light of this information, let me know, otherwise I'll merge your PR.

skriptum:
okay thanks for the clarification, ill think about it maybe for another release, so its ready to merge. merci

PR #1424, opened 2024-12-14
tidy:0.4.0

Mc-Zen:
I am submitting
- [x] an update for a package

This release is a major redesign of the documentation syntax, intending to pave the way for eventual built-in documentation support. There is a [migration guide][migration guide] for adopting the new syntax. 

Instead of (**old**)

```typ
/// This function computes the cardinal sine, $sinc(x)=sin(x)/x$. 
///
/// - x (int, float): The argument for the cardinal sine function. 
/// -> float
#let sinc(x) = if x == 0 {1} else {calc.sin(x) / x}
```
we now write (**new**)
```typ
/// This function computes the cardinal sine, $sinc(x)=sin(x)/x$. 
///
/// -> float
#let sinc(
  /// The argument for the cardinal sine function. 
  /// -> int | float
  x
) = if x == 0 {1} else {calc.sin(x) / x}
```

### The new style 
- removes almost all extra syntax and makes use of pure Typst syntax (cross references are now handled via single `@` instead of `@@`). 
- moves the parameter descriptions right in front of each parameter. 
- has a temporary syntax for type annotations that makes it relatively easy to switch to the built-in solution that we are expecting to have soon in Typst. 
### The old syntax
Note that you can also keep using the old syntax which will still be around for some time. It can be activated via `tidy.show-module(old-syntax: true, ...)`.

### New features
- New parser for the new documentation syntax. The old parser is still available and can be activated via `tidy.show-module(old-syntax: true)`. 
- Cross-references to function arguments.
- Support for detecting _curried functions_, i.e., function aliases with prepended arguments using the `.with()` function. 
- Auto-previewed examples can now be written with (inspired by the official Typst documentation)
  `````
  ```example
  #rect()
  ```
  `````
  or 
  `````
  ```examplec
  rect()
  ```
  `````

I'm happy to get feedback. Enjoy!
  
[migration guide]: https://github.com/Mc-Zen/tidy/tree/v0.4.0/docs/migration-to-0.4.0.md

### DIFF
--- a/packages/preview/tidy/0.4.0/LICENSE
+++ b/packages/preview/tidy/0.4.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Mc-Zen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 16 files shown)


Mc-Zen:
I can confirm that `src/testing.typ` is not a test file and that `src/show-example.typ` is not an example file 😅.

elegaanz:
Yes, sorry about that, the heuristics for this warning is a bit too easy to trigger. Anyways, thank you for the update!

PR #1441, opened 2024-12-17
psl-thesis:0.1.0

sdiebolt:
I am submitting
- [x] a new package
- [ ] an update for a package

This package adapts the [PSL University PhD manuscript covers](https://psl.eu/formation/choisir-sa-formation/college-doctoral-de-luniversite-psl) in Typst.

In essence, the package simply provides a `psl-thesis-covers` function to set the front and back covers of a thesis manuscript. A template is provided with an example thesis layout.  


sdiebolt:
I missed the part about "OSI approved" in the submission guidelines. Since I adapted an existing Overleaf template licensed under CC-BY-4.0, which is unusual for software, I'm not sure how to proceed here. Note that I only adapted the looks of the template, not the LaTeX code itself. How should this package be licensed?

packages/preview/psl-thesis/0.1.0/lang.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[lang.en]
+conducted_at = "Conducted at"
+presented_by = "Presented by"
+doctoral_school = "Doctoral school n°"
+specialty = "Specialty"
+jury = "Jury members"
+
+[lang.fr]
+conducted_at = "Préparée"
```
elegaanz, line 12: Not sure, but isn't it supposed to say "Préparée à" ?

packages/preview/psl-thesis/0.1.0/template/main.typ
```diff
@@ -0,0 +1,126 @@
...
+  abstracts: (fr: lorem(128), en: lorem(128)),
+  keywords: (fr: lorem(4), en: lorem(4)),
+)
+
+#counter(page).update(1)
+#set page(numbering: "i")
+
+#heading([Acknowledgments], numbering: none)
+#lorem(128)
+
```
elegaanz, line 89: Maybe it would make more sense to include `acknowledgments.typ` here?

packages/preview/psl-thesis/0.1.0/template/main.typ
```diff
@@ -0,0 +1,126 @@
...
+  ),
+)
+
+#register-glossary(glossary)
+
+// Choose between fr and en to set the front cover language.
+#set text(lang: "fr", font: "Montserrat", size: 11pt)
+
+// Set chapter style for first level headings.
+#set heading(numbering: "1.1")
```
elegaanz, line 20: I feel like a lot of the `show` and `set` rules in this file should be in the core of the package and not in the template itself. It only makes sense to have them here if you expect most users of your template to edit them. But this is up to you, and if you want to have them in the template it is possible.

packages/preview/psl-thesis/0.1.0/typst.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "psl-thesis"
```
elegaanz, line 2: If this package is officially endorsed by the university staff, it would need to undergo a short verification process. If it is not, could you please change the name to something more unique? You can simply add a distinctive adjective to the current name if you lack inspiration.

packages/preview/psl-thesis/0.1.0/lang.toml
sdiebolt: I've intentionally left out the preposition because it could depend on the institute, e.g. "Préparée au laboratoire XXX", or "Préparée à la Faculté XXX", etc. It's up to the user to add it. Whereas in English it would always be "Conducted at" (but I could be mistaken there). What do you think about it, should I do it differently?

packages/preview/psl-thesis/0.1.0/typst.toml
sdiebolt: The official LaTeX template from this university was originally created by a student. I'll check with them quickly if they can endorse this one, and if not I'll change the name!

packages/preview/psl-thesis/0.1.0/template/main.typ
sdiebolt: Indeed, putting them in the template and not in the package was intentional: I wanted the package to only provide the front and back covers of the thesis manuscript. This is similar to what the university provides: only the covers are standardized, but PhD students should be free to use any layout or style in their manuscript. Thus, I simply suggested an example style and layout for the template.

However, I'm not at all expert on how template packages are usually organized in Typst. Knowing that only covers are standardized at this university, what would you recommend?

packages/preview/psl-thesis/0.1.0/lang.toml
elegaanz: Didn't think about that, I think you're right then. Ideally there would be a more powerful translation framework for Typst that allows to change the string depending on the context, but I don't think it exists (yet).

packages/preview/psl-thesis/0.1.0/template/main.typ
elegaanz: I think it makes sense to only have the covers in the package then.

packages/preview/psl-thesis/0.1.0/typst.toml
elegaanz: A member of the university staff will have to send an email to `hello@typst.app` stating that this package is official.

sdiebolt: Perfect, I'll shoot them an email and keep you updated!

sdiebolt: To keep you updated, I've tried contacting the university to no avail. I've a got a last contact to call, but if that does not work I'll modify the name!

elegaanz: Okay, thank you for the update.

sdiebolt: I tried one last email. If I get no answer, what would you recommend as a name? Checking Typst Universe, I see that many templates simply add "modern" as prefix, would that suffice?

elegaanz: Yes that would be enough :)

elegaanz: Hello, did you get an answer yet? Is there still hope?

sdiebolt: Hey! Indeed, I finally got an answer from my university and got forwarded to two other services for validation. The doctoral college seemed OK with adding an official Typst template alongside the existing Word/LaTeX ones, but we need to wait for final approval by the bibliography department... Sorry that it's taking so long!

reknih: Just to chime in here: If you end up going for the prefix, `modern` is really a last resort. We'd prefer if you could prepend some adjective that is unique to your university.

sdiebolt: Ok, finally got final confirmation! Putting `hello@typst.app` in CC of the email thread. The thread is in French unfortunately, but don't hesitate to answer in English if you need any other confirmation.

elegaanz: Thanks :)

PR #12, opened 2023-07-01
showybox:0.1.1

Pablo-Gonzalez-Calderon:
I am submitting
- [X] a new package
- [ ] an update for a package

I have read and followed the submission guidelines and, in particular, I
- [X] selected a name in conformance with the guidelines
- [X] added a `typst.toml` file with all required keys
- [X] added a `README.md` with documentation for my package
- [X] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [X] ensured that my submission does not infringe upon the rights of a third party
- [X] tested my package locally on my system and it worked
- [X] named this PR as `name:version` of the submitted package
- [X] agree that my package will not be removed without good reason

**Description:** This component allows the user to create colorful and customizable boxes (like the ones that can be created in LaTeX using the tcolorbox library).

_Note: This package was previously named colorbox in a discarded PR. A bug was fixed and the name was changed to "showybox" for this new PR_

PR #1448, opened 2024-12-18
glossy:0.4.0

swaits:
I am submitting
- [ ] a new package
- [x] an update for a package

# feat(glossy): 0.4.0 with indefinite article support

## New features:

**Articles with `a`/`an` modifier**

Now you can include articles using this new modifier. The articles may be defined in your dictionary entry (fields `artcile`, and `longarticle`). But, if you don't define them, Glossy will do a reasonably good job at guessing them.

This modifier is also allowed before the term. For example, you might have this term:

```yaml
www: Worldwide Web
```

And you could display it with an article by any of these:

```typst
@a:www
@an:www
@www:a
@www:an
```

These, `a` and `an` are exactly equivalent. They're just provided for your convenience. Use either or both.

The `a`/`an` modifier works with all modifiers except `pl` and `def`/`desc`.

If you want to define custom articles for your terms, do so like this:

```typst
#let glossary = (
  rf: (
    short: "RF",
    article: "an",
    long: "radio frequency",
    longarticle: "a",
  )
  uts: (
    short: "UTS",
    article: "a",
    long: "untranslatable string",
    longarticle: "an",
  )
)
```

But again, in most cases (for English) you shouldn't need to because Glossy should usually get it right.

Fixes [#12](https://github.com/swaits/typst-collection/issues/12).

## Other changes:

- added test which exhausts all reference options
- added panic if 'def'/'desc' used when no `description` field exists
- significantly improved the (English) pluralizer function
- `__gls()` function refactored, cleaned up
- removed a rogue PDF

## Contributors:

- [@eltos](https://github.com/eltos)
- [@theRealProHacker](https://github.com/theRealProHacker)


elegaanz:
Thanks!

PR #1451, opened 2024-12-20
glossy:0.5.0 and glossy:0.5.1

swaits:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

**feat(glossy): release 0.5.1**

New feature:

`init-glossary()` now accepts a `format-term()` function which gives the user control over how the short-form and long-form of the terms are assembled based on the needed mode ("short", "long", or "both").

Fix:

Use codepoints (not bytes!) in article code. Fixes [#18](https://github.com/swaits/typst-collection/issues/18).

Change:

Default theme changed to `theme-academic`, which is probably more generally useful than the prior (`theme-twocol`).


elegaanz:
Should I merge both this and #1452? From what I understand, the later just backports the bug fix of this version, but I prefer to check with you before.

swaits:
Yes, please. Your understanding is correct.

elegaanz:
Okay, thank you!

PR #1454, opened 2024-12-20
muchpdf:0.1.0

frozolotl:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: MuchPDF allows you to insert PDF files as images into your Typst document.

Everything necessary for building the package is in the linked repository. I have not included it in this submission because it does not seem necessary.


frozolotl:
There still seem to be some issues. I will need a bit of debugging time still.

frozolotl:
> There still seem to be some issues. I will need a bit of debugging time until release.

Though I improved it, I wasn't able to fix the out-of-memory issue completely.
The documentation now contains a workaround users may use.
I think that's good enough for now, especially because the intended and most common use-case would be to include PDF graphics.

PR #1468, opened 2024-12-23
derive-it:0.1.2

0rphee:
I am submitting
- [ ] a new package
- [x] an update for a package

Update: Fix for https://github.com/0rphee/derive-it/issues/4. Previously the vertical alignment of each column in a table was a bit off for the 'rule name' column.

Screenshot 2024-12-23 at 1 41 48 p m
Screenshot 2024-12-23 at 1 41 57 p m


elegaanz:
Thanks!

PR #1470, opened 2024-12-23
tatras-ieee:0.1.0

MeheheCedy22:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This packages include Typst Template for IEEE papers (two-column style) for use in Slovak language. This is updated version of the 1st party [Typst template](https://github.com/typst/templates/tree/main/charged-ieee) for IEEE papers.
Not only translations are present but also I added some minor features.


elegaanz:
Thank you for submitting this package. Before actually reviewing it, don't you think it would benefit everyone to contribute the translation and the features you added to the original `charged-ieee` package rather than forking it? I'm not sure what other members of the Typst team think about it, but I feel like it would be a better solution. If that sounds good to you, we can open an issue on typst/templates to discuss that with the rest of the team.

MeheheCedy22:
I thought about it first, but since the official template in its original form, to my knowledge, is only in English, I chose to fork it instead. Another reason was that there were no other translations in the charged template. However, the main goal was that it should be an unofficial template because we were tasked in one of the university courses to write in the style of IEEE conference paper, but it was not for a conference. It was only for the styling purposes. We could choose to write in English / Slovak and since I wanted to use Typst instead of LaTeX and I wrote in Slovak, I started to change things in the charged-ieee template and I thought someone might use it for the next year's course, so I made this unofficial template instead.

packages/preview/ieee-slovak/0.1.0/lib.typ
```diff
@@ -0,0 +1,376 @@
...
+
+  // CHANGE:
+  // For acknowledgments (Poďakovanie), just use 1st level heading with the said string
+  // CHANGE:
+  // underline_links == 3 --> underline all links
+  // underline_links == 2 --> underline all links except for email of the authors
+  // underline_links == 1 --> underline all links except for email of the authors and abstract / index terms
+  // underline_links == 0 --> underline no links
+  // default: 2
+  underline_links: 2,
```
elegaanz, line 84: To be more consistent it would be nice if this argument was named `underline-links` in my opinion.

elegaanz:
I see. Then it is probably fine to have another package. Can you please change the name to something a bit more creative and less descriptive though? The same way the original package is called `charged-ieee` and not just `ieee`.

packages/preview/ieee-slovak/0.1.0/lib.typ
MeheheCedy22: I must have missed that I used snake_case. I will fix it right away.

elegaanz:
Thanks!

PR #1482, opened 2024-12-27
vienna-tech:0.1.3

npikall:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed a bug, where the appendix label did not change with the language.


elegaanz:
Thanks:

PR #1483, opened 2024-12-27
grotesk-cv:1.0.2

AsiSkarp:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added the option for clickable links in personal info (homepage, LinkedIn, etc)
- Fixed alignment issues in cover letter template
- Added options for alignment and positioning of skill tags in CV
- Updated to `fontawesome/0.5.0`


elegaanz:
There is something I didn't catch in previous releases that I would like you to change: fonts should not be shipped with packages. Is it possible to delete the `fonts` directory and explain how to download and install/import the required files in the README and/or in a comment in the template please?

AsiSkarp:
It is possible, but it would really make for a bad user experience as the template is designed with the Hanken Grotesk font in mind, so the template would not look like it is intended to when first opening it up on Typst Universe. 
Is there no other way of providing font support out of the box?

elegaanz:
Not as of now, unfortunately. But the font is available on typst.app under the name `HK Grotesk` (at least, I believe it is the same font, maybe I'm mistaken and it is not). You can support multiple font names easily with:

```typst
#set text(font: ("Hanken Grotesk", "HK Grotesk"))
```

This way, the template would work out of the box at least on the web app.

PR #1488, opened 2024-12-29
lacy-ubc-math-project:0.1.0

lace-wing:
I am submitting
- [x] a new package
- [ ] an update for a package

A template for University of British Columbia (UBC) MATH 100 group project.


packages/preview/lacy-ubc-math-project/0.1.0/template/project1.typ
```diff
@@ -0,0 +1,45 @@
...
+    You can do it.
+  ]
+]
+
+// Use help.<section> to get help.
+#help.getting-started
+// #help.setup
+// #help.math
+// #help.drawing
+// #help.caveats
```
elegaanz, line 45: I think it's the first time I see a package ship its manual this way, and I'm not sure how I feel about it. :sweat_smile:

packages/preview/lacy-ubc-math-project/0.1.0/help.typ
```diff
@@ -0,0 +1,123 @@
...
+  "setup",
+  "author",
+  "math",
+  "drawing",
+  "question",
+  "solution",
+  "caveats",
+)
+
+#let get-orientation(tag) = {
```
elegaanz, line 15: This is not the intended use for labels, and this will produce a warning if users try to display some manual sections (because the same label is used more than once). A nicer alternative hack to achieve what you are doing here would be to have a custom language for raw blocks that indicates which layout to use, that you replace with `typst` in the end.
Otherwise, simply having a function in which to wrap raw blocks would be the clean approach here.

packages/preview/lacy-ubc-math-project/0.1.0/template/assets/madeline-math.jpg
elegaanz, line 1: Not sure if it is fine to package this image with regards to copyright. I think the original author wouldn't mind, but technically we are not allowed to distribute it. Could you either get their explicit agreement or replace it with another image that you made yourself or that is under a permissive license please?

elegaanz (commented):
Thank you for this package. Some small remarks:

packages/preview/lacy-ubc-math-project/0.1.0/template/project1.typ
lace-wing: I find having to switch to another PDF or webpage just because I forget something in the moment annoying. And since Typst compiles so fast, inline snippets of help might be useful. An experiment at this time.

lace-wing:
@elegaanz Resolved multiple label (function wrap) and image credit issue (drew my own).

lace-wing:
Seems like after the last commit which I updated the thumbnail, the package check never finished and my local compile exits with either 6 or 10, even if I change the image or remove reference to it, no idea why.

elegaanz:
Love the new image :smile: You seem to have triggered a compiler bug, I will investigate a bit more what is going on and make an issue.

elegaanz:
Okay, so I've found what the issue is, and I'm not sure if this is a compiler bug or expected behavior. Basically, with the new `showcode` function, you (via `showman`) are importing your own package. But by doing so, you trigger a new evaluation of your package, including the `showcode` function, that will in turn import itself, and so on until you have a stack overflow error that crashes the compiler. This would previously not happen because `eval` would be run only when encountering a `raw` block with the correct label, so as long as none of your manual page used the `help` module in one of the example, you were safe from this recursion.
I assume the caching of modules that Typst usually does to avoid evaluating the same file twice does not happen here because `eval` happen in an isolated context.

Now, I'm not sure what is the correct solution here is. I think wrapping each `help` section or the `showcode` function in a `context` block could be a workaround, as this would delay the evaluation until the module is actually displayed somewhere in the document. But maybe the compiler could behave better here, even if this is a bit of an edge case.

lace-wing:
I don't think it's because of the new showcode impl.
I tried `git reset --hard HEAD~2` which walks back to [dac03a2](https://github.com/typst/packages/pull/1488/commits/dac03a2fc71050e78dcd69745f0ac2509562b1b2), but still got the `6` or `10` termination.

elegaanz:
I can't reproduce the bug on this commit. What is your process to get the crash exactly?

lace-wing:
Simply `typist c/w template/project1.typ`.

elegaanz:
I made a symlink between the `packages/preview/lacy-ubc-math-project` directory and `~/.cache/typst/packages/preview/lacy-ubc-math-project` and running `typst c template/project1.typ` works fine on `HEAD~2`. On the latest commit I get a segfault.
Which operating system are you using, and which version of Typst?

lace-wing:
macOS and 0.12.0

lace-wing:
At this point I will just remove the inline doc and get it going...

elegaanz:
If it fixes the issue, this is probably the best solution yes.

lace-wing:
@elegaanz removed inline help, now everything should compile.
P.S. my errors after git reset was because I did not also reset what was in the preview registry.

packages/preview/lacy-ubc-math-project/0.1.0/template/project1.typ
```diff
@@ -37,9 +37,4 @@
   ]
 ]
 
-// Use help.<section> to get help.
-#help.getting-started
-// #help.setup
-// #help.math
-// #help.drawing
-// #help.caveats
+There is also a #link("/lace-wing/lacy-ubc-math-project/blob/master/manual.pdf")[manual] for this project.
```
elegaanz, line 40: I'm not sure this link will point to the correct URL, unless I missed something. Shouldn't it start with `https://github.com/`?

lace-wing: Fixed.

elegaanz:
Thanks!

PR #1497, opened 2024-12-30
tada:0.2.0

ntjess:
I am submitting
- [ ] a new package
- [x] an update for a package

Use `table` instead of `tablex`, and update to tidy docs 0.4.0 syntax


elegaanz:
Thank you for the update :)

PR #1520, opened 2025-01-06
unofficial-fhict-document-template:1.1.2

TomVer99:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed citation style bug (quite important), updated dependencies


elegaanz:
Thanks for the update :)

PR #1522, opened 2025-01-07
htlwienwest-da:0.2.1

peterw16:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Makes the supervisor of an author optional. This allows users to format the author-supervisor table on the front-page in a better way.


elegaanz:
Thanks!

PR #1527, opened 2025-01-08
easy-typography:0.1.0

swaits:
I am submitting
- [x] a new package
- [ ] an update for a package

**Description**: 

A package which sets up sensible typography defaults, helping Typst users get up and running quickly.


ssotoen:
Not a fan of oldstyle figures?

swaits:
@ssotoen, no. I am not a fan of them descending so much. Doesn't feel modern to me, and I think it negatively impacts readability in today's world (because of unfamiliarity). But, I'd love to know what you think -- would a typography geek disagree with me?

packages/preview/easy-typography/0.1.0/src/main.typ
```diff
@@ -0,0 +1,92 @@
+#import "constants.typ": *
+
+/// The main function for easy typography setup.
+/// - body-size: The size of body text (default 10pt).
+/// - fonts: A dictionary specifying heading/body fonts.
+/// - paper: The paper size to use (none to skip).
+/// - margin: Margin for the page (none to skip).
```
neuralpain, line 7: Documentation uses three forward slashes `///` and I would recommend leaving a space after the main description of the function and including a return type for the parameters and the function, if any.

```suggestion
/// The main function for easy typography setup.
///
/// - body-size (): The size of body text (default 10pt).
/// - fonts (): A dictionary specifying heading/body fonts.
/// - paper (str): The paper size to use (none to skip).
/// - margin (): Margin for the page (none to skip).
/// -> content
```

neuralpain:
> I think it negatively impacts readability in today's world (because of unfamiliarity)

@swaits Typography fanatic here. I personally like the look of serifs but if we're talking about the likes of Georgia, the old style numbering is quite loathsome.

Also, apart from the two changes mentioned above, your package looks good to merge (to me). I'm interested in how this grows in the future. Maybe you can provide some **suggested** free/opensource modern fonts in the package for the non-professional userbase with links to acquire them.

ssotoen:
Oldstyle figures are a matter of preference. I can see why some people don’t like them. For me it depends on the type of document and font but I usually prefer them. They’re still very common in books and newspapers so I don’t see unfamiliarity as an argument against them.

Because there is no objectively correct answer I don’t think we can set a sensible default, especially since setting this parameter makes it impossible for the user to change it. My suggestion would be to remove that line and just using the font’s default look. Libertinus and all the fonts in your tasteful pairing except Merriweather use lining figures anyway.

packages/preview/easy-typography/0.1.0/src/main.typ
```diff
@@ -0,0 +1,92 @@
...
+    size: body-size,
+    tracking: 0pt,
+    spacing: 100% + 0pt,
+    baseline: 0pt,
+    overhang: true,
+    hyphenate: true,
+    kerning: true,
+    ligatures: true,
+    number-type: "lining",
+  )
```
ssotoen, line 47: I strongly suggest you set up proportional figures. Tabular figures look bad in running text, and unfortunately a lot of digital fonts use them by default—probably because the designers expect most people use them in office programs where tabular data is common and Opentype support is not.

```suggestion
    number-width: "proportional",
  )

  #show table: set text(number-width: "tabular")
```

swaits:
> My suggestion would be to remove that line and just using the font’s default look. Libertinus and all the fonts in your tasteful pairings except Merriweather use lining figures anyway.

This is a great suggestion! I'll make the edit today.

BTW, I think I should've mentioned... in my mind the target writing style for easy-typography is probably mostly professional/corporate, maybe followed by academic. In that sense, I think making conservative choices is best.

I'm **very** open to suggestions on both this and tasteful-pairings. I'm planning to rev this to accept tasteful-pairings (i.e. "office") identifiers directly. And I'm open because even though I geek out about typesetting and typography, I don't feel I'm a true typography geek. So the collective wisdom of the community is valuable to me!

packages/preview/easy-typography/0.1.0/src/main.typ
ssotoen: Should be `show table: set text(number-width: "tabular")`, without the octothorpe. Oops!

elegaanz:
Looks good, thank you!

PR #1530, opened 2025-01-08
conchord:0.3.0

sitandr:
I am submitting
- [ ] a new package
- [x] an update for a package

A major `conchord` update, partially inspired by @soxfox42's `chordish` package. Changes include, but not limited to:

- `conchord` now has a **WASM plugin** to find all possible chord variants from their _names_ and detect the best ones. The chord generation works for arbitrary tuning and number of strings for basically every chord you can think of. Here is an example: 

- New song sheet functionality. Automatically change tonality of any part of song and use short notations of chords in song with special `show` rule. Use different display functions for line chords. Create a dynamically updated "chord library" that respects tonality shift.
- Chord drawing: optional thick nuts (thanks, @soxfox42!) .
- Tab improvements: simple `rhythm` bar, repetition macro, ghostnotes (thanks, @KMIJPH!).
- Update to `cetz:0.3.1`, removing need in deprecated functions.
- Rewritten README and new _Reference_ with tidy. Lots of documentation improvements in general.

See README for documentation and more examples!


packages/preview/conchord/0.3.0/chords/sheet.typ
```diff
@@ -0,0 +1,226 @@
...
+  for (i, c) in query(chords-selector)
+      .map(c => shift-chord-tonality(c.text.trim(), get-tonality(c)))
+      .dedup()
+      .enumerate() {
+    if c in exclude {
+      continue
+    }
+    let n = switch.at(c, default: 0)
+    let at = at.at(c, default: none)
+    box(align(center+horizon, smart-chord(c, chordgen: chordgen, n: n, at: at, scale-l: scale-l)), width: get-chordgram-width-scale(tuning.split().len()) * scale-l, height: 80* scale-l)
```
soxfox42, line 204: `tuning` should be passed to `smart-chord` here, otherwise the tuning only affects the placement of the diagrams.

sitandr:
@soxfox42 pointed me a few bugs, I would like to fix them before the release. I will try fixing them by the Friday evening, so I'll mark it as a draft for now.

packages/preview/conchord/0.3.0/chords/sheet.typ
sitandr: Fixed this in repo, thanks! Now working on `conchord-gen` to make good ukulele chords.

sitandr:
Okay, I've done it. It was faster than expected.

Warnings of the bot all seem quite irrelevant to me (sorry if it has been already discussed elsewhere). The `N` argument at putting chords in box is used to make it different from `n` used for chordgen. For the examples `../lib.typ` seems like the best import option possible, since `@preview` is not available and `@local` would be even more confusing and less convenient.

elegaanz:
For the `n` vs `N` problem I agree with you that it is not relevant here.

For the examples, the point of having them is to be able to copy paste the source code easily and start from that. But because you seem to include them again in your README, I would say it is okay to leave the source file as they are if it is more convenient for you. By the way, there is an import for version 0.2.0 in the README on line 125, you probably want to change that.

And speaking of the examples, could you please `exclude` them? I don't think users of the package will benefit much from having a copy on their machine, and it would make downloading the package slower.

sitandr:
Thanks for noting the `0.2.0` typo in README.

About the excluding: isn't this line in toml already excluding them?

```toml
exclude = ["examples", "reference"]
```

elegaanz:
Ugh, my bad, I missed that line. LGTM then!

PR #1534, opened 2025-01-08
pigmentpedia:0.3.0

neuralpain:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixed issue with scoped search in #1495 and added visual improvements.


elegaanz:
Thanks for this update :) The only remark I have before merging is that pigment names are in Uppercase-Kebab-Case, which feels inconsistent with the rest of the Typst ecosystem (where kebab-case is by far the most common convention, with only a few packages using camelCase or snake_case). It would be nice if you could change the names so that they all use lowercase-kebab-case. If you really prefer to keep the upper case version please tell me, it is not a hard requirement to follow this convention to have your package merged, I'm just making a suggestion to improve your package.

neuralpain:
> pigment names are in Uppercase-Kebab-Case, which feels inconsistent with the rest of the Typst ecosystem

Yeah, the naming convention of the pigments was a difficult decision for me. I thought about staying within the preferred `kebab-case`, however, since they were actual names, I was on the fence as to whether or not I should change their appearance within pigmentpedia.

Let me think about it some more and I'll let you know when I'm ready for another review.

elegaanz:
Thank you!

PR #1536, opened 2025-01-08
light-cv:0.2.0

AnsgarLichter:
I am submitting
- [ ] a new package
- [x] an update for a package

I updated my packages so that the user is able to adapt the styles defined in settings/styles.typ and the changes actually reflect in the document itself.


packages/preview/light-cv/0.2.0/modules/utils.typ
```diff
@@ -0,0 +1,7 @@
+#let vLine() = [
```
elegaanz, line 1: I would suggest to rename this variable to `vline` or `v-line` to be more consistent with other function names.

packages/preview/light-cv/0.2.0/modules/header.typ
```diff
@@ -0,0 +1,76 @@
+#import "../template/settings/styles.typ": *
```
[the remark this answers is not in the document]
AnsgarLichter: Removed it - it was unused anyway.

packages/preview/light-cv/0.2.0/modules/utils.typ
AnsgarLichter: Adapted it.

elegaanz:
Thanks!

PR #1538, opened 2025-01-09
handy-dora:0.1.0 and rexllent:0.2.0

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package (for 0.1.0 is from exmllent)

Description: 
- Update examples in Handy Dora's readme.
- ReXLlenT is a typst package that helps you convert Excel xlsx tables to typst tables, powered by wasm. Rename from eXMLlent to ReXLlenT is because that the package is no longer handle table through a excel exported XML file but XLSX itself.


elegaanz:
Thanks!

PR #1541, opened 2025-01-10
flow:0.1.2

MultisampledNight:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Bunch of utilities for writing notes and condensing knowledge, including todo lists, metadata management, templates, callouts, themes and other things for writing math and diagrams.


elegaanz:
There a few warnings because you depend on Polylux and they didn't yet release a version that uses `context` instead of `locate`/`style`. This means that your package will break once Typst 0.13 is released. Hopefully Polylux will have released a new version before that, and you will just be able to bump the version number. Otherwise you will have to migrate to another solution, like `touying`.
For now it is fine to have these warnings (since you are not directly responsible for them), so I'll merge this release as is, but I just wanted you to be aware of that, so that you're not surprised.
Anyway, thanks for submitting the update.

PR #1556, opened 2025-01-12
clean-dhbw:0.1.0

roland-KA:
I am submitting
- [x] a new package
- [ ] an update for a package

This is a (modernized) Typst template for DHBW documents like Bachelor theses, "Studienarbeiten", project documentation etc. and thus may be useful for all DHBW students.

- [ x] selected a name that isn't the most obvious or canonical name for what the package does
- [ x] added a `typst.toml` file with all required keys
- [x ] added a `README.md` with documentation for my package
- [x ] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x ] tested my package locally on my system and it worked
- [ x] `exclude`d PDFs or README images, if any, but not the LICENSE


roland-KA:
I need some help with this PR. Why did the check for the PR fail?

roland-KA:
Oh, I see, thanks!

roland-KA:
The warnings which now remain, occur in an external package `codelst` (which is used to display code snippets within the body text) on which I don't have any influence. Overmore this package is only an option to use within the template. So I would suggest to let the PR pass.

roland-KA:
Done! ... and thanks for the hint!

elegaanz:
Thanks!

PR #1560, opened 2025-01-13
rexllent:0.2.2

hongjr03:
fix: default value lead to panic

I am submitting
- [ ] a new package
- [x] an update for a package

Hi! The version 0.2.2 of ReXLlenT has these updates:
- fix the issue that default settings leads to panic. 
- change the default value of `parse-table-style` to `false`.


elegaanz:
Thanks!

PR #1563, opened 2025-01-13
oh-my-euxfel:1.0.0

JamesWrigley:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: the package provides a template for presentations with the European XFEL logo and style.

Note: the `thumbnail.png` and `theme-demo.png` files are very slightly different in that the header is missing from the title slide in `thumbnail.png`. This is because I compiled the file for the thumbnail under Typst 0.12 and for some reason that broke the title slide header :thinking: I'll look into it later, but I don't think it needs to block registration.


JamesWrigley:
Not sure why the preview job is stuck :eyes: Locally `presentation.typ` builds in less than a second.

packages/preview/oh-my-euxfel/1.0.0/README.md
```diff
@@ -0,0 +1,31 @@
+# An XFEL Polylux theme
+
+This is a [Polylux](https://polylux.dev/book/) theme for [European
```
elegaanz, line 3: Polylux is not very actively maintained and the latest version is using deprecated functions. I would rather not merge a package that reports deprecation warnings as they will become hard errors with the next Typst release. I contacted the Polylux maintainer to ask them to make a new release to fix that, but it may take some time before they actually do it. If you would like to have your package released rapidly, I would suggest switching to another presentation package, like `touying` which should have a similar API and that is more actively maintained.

packages/preview/oh-my-euxfel/1.0.0/logo.svg
elegaanz, line 1: I assume that this is the official logo and that you didn't design it yourself. If that's true, can you please add a note in your README and/or LICENSE file stating under which license it is distributed?

elegaanz:
And about the CI never ending, it is just that it ran out of disk space and crashed in the middle of the checks. I should migrate it to a more powerful server, but this is "normal" for now.

packages/preview/oh-my-euxfel/1.0.0/README.md
JamesWrigley: Ah I see, I wasn't aware of that... I'll have a go at switching it.

elegaanz: Actually, I already got an answer from them and they said they would make a release to fix this issue soon. So if you prefer to wait a few days, it should not be *that* long.

elegaanz: Polylux 0.4.0 was just released, updating it should fix the issues.

JamesWrigley: Thanks for the reminder, I forgot about this :sweat_smile: I have heard good things about touying though so I'll have another look at that before proceeding, will try to finish this this week.

elegaanz:
Hello, were you able to make the changes we discussed? If not, I will close this PR for now, to keep only active ones in the list. You can always open a new one when you will be ready.

JamesWrigley:
I did not sorry :grimacing: Thanks for the reminder, will open a new one when I get the chance.

PR #1575, opened 2025-01-15
ansi-render:0.8.0

8LWXpg:
Minor update with doc change.

Added doc about capturing ANSI output and the default font choice.


elegaanz:
Thank you!

PR #1581, opened 2025-01-16
classic-aau-report:0.2.0

Tinggaard:
I am submitting
- [ ] a new package
- [x] an update for a package

Added macros to import in sections, and a brief introduction to writing in Typst in general.


elegaanz:
Thanks!

PR #1589, opened 2025-01-18
modern-whs-thesis:0.1.0

alex289:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This template can be used for scientific papers and theses of the Westphalian University of Applied Sciences (Westfälische Hochschule). I created this for myself and some fellow students and we are using it a lot.


elegaanz:
Hello and thank you for submitting this package. Could you please change the name to something more unique? Unofficial packages should be clearly marked as such, and the current name may suggest that this is an official implementation. You can just add an adjective to the current name if you don't have any other ideas.

alex289:
> Hello and thank you for submitting this package. Could you please change the name to something more unique? Unofficial packages should be clearly marked as such, and the current name may suggest that this is an official implementation. You can just add an adjective to the current name if you don't have any other ideas.

@elegaanz, of course no problem.

Do you mean something like "modern-whs-thesis" or "whs-custom-thesis"?

packages/preview/modern-whs-thesis/0.1.0/example/main.typ
```diff
@@ -0,0 +1,30 @@
+#import "meta.typ" as meta
+#import "@preview/whs-thesis:0.1.0": *
```
elegaanz, line 2:
```suggestion
#import "@preview/modern-whs-thesis:0.1.0": *
```

elegaanz (commented):
The new name works, thanks for changing it. There is one occurrence of the previous one you forgot :)

packages/preview/modern-whs-thesis/0.1.0/example/main.typ
alex289: Fixed 😄

alex289:
Thanks a lot for the help and for considering this package

elegaanz:
Thanks!

PR #1591, opened 2025-01-18
modern-whs-assignment:0.1.0

alex289:
Waiting for #1589 to be completed

I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a template some students and I use for all kinds of assignments like project documentations, exercises, any kind of tasks or learning notes.


alex289:
Im not sure what the error in the workflow means 😅. Can somebody give me a hint

elegaanz:
Try to create a `0.1.0/` sub-directory and to move your code here, instead of putting it in `packages/preview/whs-assignment/` directly.

Can you also change the name of the package? It is a bit too canonical in my opinion.

alex289:
> Try to create a `0.1.0/` sub-directory and to move your code here, instead of putting it in `packages/preview/whs-assignment/` directly.
> 
> Can you also change the name of the package? It is a bit too canonical in my opinion.

@elegaanz, ahh thank you.

I added the same prefix as with the whs-thesis but I have problems with thinking about a name.

I use this template for:
- project documentations
- exercises
- any kind of tasks I have to submit
- summary of scripts of topics for a course

Assignment was the only name I could think of which describes this. Do you have an idea maybe?

elegaanz:
With the prefix, having assignment in the name is not a problem. :)

packages/preview/modern-whs-assignment/0.1.0/example/main.typ
```diff
@@ -0,0 +1,14 @@
+#import "meta.typ" as meta
+#import "@local/modern-whs-assignment:0.1.0": *
```
elegaanz, line 2:
```suggestion
#import "@preview/modern-whs-assignment:0.1.0": *
```

elegaanz (commented):
Just need to fix that import, and it should be good.

elegaanz:
Thank you!

PR #1595, opened 2025-01-18
iconic-salmon-svg:2.1.0

Bi0T1N:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
With this update the library is way more flexible in using/exchanging icons, even without generating a clickable link. It also allows to use local icons that are not (yet) part of the package. It also supports a few more websites now.


packages/preview/iconic-salmon-svg/2.1.0/examples/examples.pdf
elegaanz, line 1: Could you please delete this file? It can easily be reproduced by compiling the example if necessary, and will only make this repository heavier.

packages/preview/iconic-salmon-svg/2.1.0/examples/minimal_example.pdf
elegaanz, line 1: Same for this one.

packages/preview/iconic-salmon-svg/2.1.0/iconic-salmon-svg.typ
```diff
@@ -0,0 +1,268 @@
...
+/// In addition to this, the associated icon of the web page is inserted
+///
+/// - name (str): The visible name of the clickable link
+/// - url (str): The URL to the website (it is assumed to start with the http/https scheme)
+/// - icon_provider (function): The function that provides the associated icon
+/// -> content
+#let icon-link-generator(
+  name,
+  url,
+  icon_provider,
```
elegaanz, line 13: I would suggest to rename this parameter (here and in all other places you use it) to `icon-provider` to be more consistent with other parts of your package and more generally the Typst ecosystem.

packages/preview/iconic-salmon-svg/2.1.0/examples/examples.pdf
Bi0T1N: Is this new? The other two versions have a PDF file...
Would it be accepted if I add a `.gitignore` file to ignore PDF files for my package in the future?

packages/preview/iconic-salmon-svg/2.1.0/iconic-salmon-svg.typ
Bi0T1N: Indeed, thanks for the hint! :smiley:

packages/preview/iconic-salmon-svg/2.1.0/examples/examples.pdf
elegaanz: It is not really new, but we don't always very strictly enforce it to be honest, so it may have gone unnoticed in previous versions. And yes, having a .gitignore would be fine.

Bi0T1N: Added a `.gitignore` file

packages/preview/iconic-salmon-svg/2.1.0/iconic-salmon-svg.typ
Bi0T1N: Fixed. I also changed `icon_func` to `icon-func`.

elegaanz:
Thanks!

PR #1602, opened 2025-01-20
modern-cv:0.8.0

DeveloperPaul123:
I am submitting
- [ ] a new package
- [x] an update for a package

Lots of fixes and some new features/options.


packages/preview/modern-cv/0.8.0/lib.typ
```diff
@@ -0,0 +1,851 @@
...
+      ],
+    )
+  ]
+}
+
+/// ---- End of Resume Template ----
+
+/// ---- Coverletter ----
+
+#let default-closing(lang_data) = {
```
elegaanz, line 544: To be more consistent I would rename `lang_data` to `lang-data`.

elegaanz (commented):
Just one small thing, otherwise looks good. Thank you :)

packages/preview/modern-cv/0.8.0/lib.typ
DeveloperPaul123: Thanks, done!

elegaanz:
Thanks!

PR #1611, opened 2025-01-22
efilrst:0.3.1

jmigual:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- Fixed wrong numbering when using sub lists
- Added the option to choose the joiner between a reference name and the reference counter value


elegaanz:
Thanks!

PR #1621, opened 2025-01-24
flow:0.2.0

MultisampledNight:
I am submitting
- [ ] a new package
- [x] an update for a package

Set of templates for quick and painless notetaking, as well as a few utilities for drawing diagrams and the works.


elegaanz:
Thank you!

PR #1627, opened 2025-01-25
minimal-thesis-luebeck:0.2.0

fhemstra:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I noticed that I had a severe issue in the previous verison, as I included files from the template directory instead of the copies I gave to the user. With this update, I changed it so that everything the template uses is also available to the user. This includes leaving the entrypoint function empty.

While I see how restrictig the user can be helpful when making templates, my goal with this template is different: I want to give full control to the user so they can understand and change every piece of it to ultimately truly own their document.

I do not think this stands in contrast to the established typst template paradigm but is rather a different flavor of the same thing.


elegaanz:
This architecture would make updating your template to fix bugs (or add new features) much harder for users. It may also be a bit confusing to have so many files. I would argue that if they really want control over everything they are free to download a copy of the template's source code and import it. So I'm not really in favor of this, but I can maybe discuss it with the rest of the team to know what they think.

fhemstra:
Yeah okay, I can see that. I will rework it tomorrow to get the best of both an then push again.

fhemstra:
Okay so I think I got it now:
- functions for hiding configuration stuff and enabling updates
- not too many files
- decent configurability through parameters

Also, I think I got a lot better at testing locally, so I am fairly certain I don't have the same include issue as with the 0.1.0.

Thanks for your patience, looking forward to your opinion!

elegaanz:
Yeah, it looks better IMO. Thank you!

PR #1626, opened 2025-01-25
polylux:0.4.0

andreasKroepelin:
I am submitting
- [ ] a new package
- [x] an update for a package

Polylux is a package for creating slides. The new version is fit for Typst 0.12.0 and does not ship themes anymore, but relies on Typst templates from now on.

PR #1632, opened 2025-01-26
basic-polylux:0.1.0

andreasKroepelin:
I am submitting
- [x] a new package

A template to get started with the Polylux slides package.


andreasKroepelin:
Requires #1626 to be merged first.

packages/preview/basic-polylux/0.1.0/template/slides.typ
```diff
@@ -0,0 +1,50 @@
...
+  paper: "presentation-16-9",
+  footer: align(right, text(size: .8em, toolbox.slide-number)),
+  margin: (bottom: 2em, rest: 1em),
+)
+#set text(
+  font: "Lato",
+  size: 23pt,
+)
+#show math.equation: set text(font: "Lete Sans Math")
+#show heading: set block(below: 2em)
```
elegaanz, line 13: I think it would be more "idiomatic" to have these rules in a function that you `#show: …`.

andreasKroepelin: I thought about this and the purpose of this template is to only make a suggestion and be as configurable as possible, that's why I opted for this design.

elegaanz: Then I would argue that you could either provide parameters to the function, or let people override the rules they want to change, with something like:

```typst
#import "@preview/basic-polylux:0.1.0: basic-polylux"

#show: basic-polylux
#show math.equation: set text(font: "Cambria Math")
```

But if you're not convinced by this approach I won't force you to use it. I can merge the package as is :)

andreasKroepelin: I would prefer this template to communicate "you could have done this yourself". So I'd be happy to have this merged in its current state :)

PR #1642, opened 2025-01-28
equate:0.3.0

EpicEricEE:
I am submitting
- [ ] a new package
- [x] an update for a package

Changelog: 
- Added `share-align` function to align separate equations with each other (allowing something similar to LaTeX's `\intertext`)
- Added missing counting symbols when handling references
- Equations where only a single line is labeled now still include the sub-number
- Fix behavior with nested equations (i.e. disables numbering and the equate show rule)


elegaanz:
Thanks!

PR #1649, opened 2025-01-30
Ignore PDF files for packages

Bi0T1N:
This adds an entry to the `.gitignore` file to ignore all (new) PDF files for packages. Avoids that every single package maintainer may have it's own `.gitignore` file to achieve the same result... :smile: 

```
find packages/ -name .gitignore | wc -l 
154
```

Related to https://github.com/typst/packages/pull/1595#discussion_r1922043640

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ desktop.ini
 
 # Rust
 /bundler/target
+
+# PDF files for packages
+packages/**/*.pdf


elegaanz:
Thank you, but unfortunately there are a few exceptions to this rule: for instance we generally allow manual files as PDF, especially if they are linked in the README. This .gitignore rule would be too generic to allow for that, and I don't think that you can refine it to allow only for some specific PDF files (judging the contents of the PDF can probably only be done by a human). I'll close for this reason.

Bi0T1N:
Well, even if the `.gitignore` has a line to skip PDF files for packages you can add them to the commit by using `git add --force wanted.pdf`.
The other option I see to make it easier for maintainers is to allow the `.gitignore` file within the package directory. Currently the pipeline fails if you do so:
```
Error: packages/preview/iconic-salmon-svg/.gitignore: a package directory may only contain version sub-directories, not files.
```
The current state is that you would need to have a single `.gitignore` file in every version sub-directory which is cumbersome and prone to errors. And somehow useless, because you no longer touch the version directory after the merge to the `main` branch.

elegaanz:
Yes please, it will be more visible than this PR.

PR #1651, opened 2025-02-01
tableau-icons:0.1.0

joelvonrotz:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: `tableau-icons` is a package for inserting Tabler.io icons into a document. It provides two functions, one for 'block' icons and the other for inline icons, and supports coloring and size changing. To handle the large icon count better, I've created two svg spreadsheets, which Typst filters through via regex. These can be updated easily through a python script.

While it works, a heavy stress test has not been done. The PDF document inside the `docs` folder is the largest stress test I've done (background of the header image has been created with this package).

- As this is my first package, I would be grateful for some feedback.


joelvonrotz:
One thing I am concerned with is the 2MB file for the outlined icons. I do not wish to "motivate" others to also create such large files and therefore would like to hear your opinion on these.

joelvonrotz:
converted the render system from regexing to using dictionary structures. While it introduces a bit larger files, it improves rendering quite a lot. On my system rendering the stress test doc (which renders all the icons) takes ~1.2s with the json variant compared to the ~13s of the regex variant.

elegaanz:
Hello, and thank you for submitting this package. A preferable approach here would be to have the icons in a font file, and the package would only be responsible for translating human readable names to glyphs. The user of your package would have to download and install the font file by themselves, as they are not allowed in packages. Could you investigate to see if tabler.io provides an icon font, or if there are tools you could use to convert the SVG files to OTF, please?

joelvonrotz:
Thank you, so I've redid the thing and it should now work with the tabler icons font file!

elegaanz:
Thank you :)

PR #1654, opened 2025-02-01
abbr:0.2.1

jhalmen:
I am submitting
- [ ] a new package
- [x] an update for a package

- adds more configuration, convenience funcitons, and data loading from file.
- also cleans up exposed API
- some internal cleanup which results in less compiler warnings about non-convergence, plus a bugfix for the long-only forms.


elegaanz:
Thanks!

PR #1655, opened 2025-02-01
New Template "create-lattes-cv"

philkleer:
I am submitting
- [x] a new package/template
- [ ] an update for a package

Description: The template creates from the exported data from [LATTES](https://lattes.cnpq.br/) a CV in the format and with the information needed in regard to standards in Brazil (CNPq). Exports on the website create only mal-formatted RTFs, and, therefore based on a transformation from the XML-File to TOML, this template creates a nice looking curriculum. A vast majority of options is already included, however, [LATTES](https://lattes.cnpq.br/) offers a variety of entries and not all is covered yet. However, for standard curriculums the template should already fit. 


packages/preview/create-lattes-cv/0.1.0/lib.typ
```diff
@@ -0,0 +1,140 @@
...
+  database: "output.toml",
+  kind: "completo", 
+  me: str,
+  last_page: true,
+  date: datetime.today(),
+  subtitle: "Curriculum Vitae",
+  body,
+) = {
+    // define details:
+    let details = toml(database)
```
elegaanz, line 22: Because files are read relatively to the current package/project when using `toml` (and other functions that read from external files), this won't work as expected: you would only be able to access files in the package directory, not in the user's project. You should take the parsed data directly as a parameter, instead of a file path.

packages/preview/create-lattes-cv/0.1.0/lib.typ
```diff
@@ -0,0 +1,140 @@
...
+// - kind: o tipo de currículo Lattes (string)
+// - me: o nome para destacar nas citações (string)
+// - last_page: resumo de produção no final (boolean)
+// - data: a data de currículo (datetime)
+// - subtitle: para a página inicial 
+#let lattes-cv(
+  database: "output.toml",
+  kind: "completo", 
+  me: str,
+  last_page: true,
```
elegaanz, line 16: As suggested by the package review bot, it would be better if this parameter could use `kebab-case` instead of `snake_case`. And there are a few other occurrences of this in your package.

packages/preview/create-lattes-cv/0.1.0/template/example.typ
elegaanz, line 1: I think it would be nice to rename this file to `main.typ` or `cv.typ`, as it will be the name of the file when creating a project from your template.

elegaanz:
Also, can you please change the name of the package to something less canonical for what it does?

philkleer:
Many thanks for the comments and hints. I solved the problems and also corrected kebab-case and the entry for the toml-file. I changed the template-name to a less canonical one.

elegaanz:
Thank you :)

PR #1678, opened 2025-02-06
drafting:0.2.2

ntjess:
I am submitting
- [ ] a new package
- [x] an update for a package

- Support inside/outside margins + outline of all margin notes (Many thanks, @Tinggaard!)
- Preps 0.13 compiler support
- Uses new Tidy documentation

### DIFF
--- a/packages/preview/drafting/0.2.2/LICENSE
+++ b/packages/preview/drafting/0.2.2/LICENSE
@@ -1,0 +1,24 @@
+This is free and unencumbered software released into the public domain.
+
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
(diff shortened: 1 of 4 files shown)


xkevio:
Your README here and on the repo itself still uses `0.2.1` when importing. This should probably be changed as well.

elegaanz:
Just a small reminder that the README needs to be updated, if you have a bit of time :) It would be nice to have this new version merged.

ntjess:
Thanks for the ping. I need to also switch from `path` to `curve` since I see the deprecation in 0.13

ntjess:
@elegaanz the most recent changes:
- Update the readme as requested
- Use `curve` instead of `path`
- Indicate 0.13 compiler in `typst.toml`

Thanks for your patience!

elegaanz:
Great, thank you!

PR #1690, opened 2025-02-09
added ez-today-1.1.0 update

CarloSchafflik12:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: You can now specify if the day number should be zero-padded or not. The ISO 8601 format now uses zero-padded day numbers.


elegaanz:
Thanks!

PR #1695, opened 2025-02-10
vlna:0.1.0

ShinoYumi:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Creating official package for "vlna" behaviour in LaTeX. Primary useful for Czech language and academic typesetting.


packages/preview/vlna/0.1.0/README.md
```diff
@@ -0,0 +1,10 @@
+# typst-vlna
+
+First version of vlna. Main goal is to create similiar package to luavlna/vlna from LaTeX.
+
+## How to use:
+````typst
+#import "@preview/vlna:0.1.0":*
```
elegaanz, line 7: I'm not sure this would work: the `show` rule would only apply to the scope in which it is defined, which means the file in which it is defined. What you probably want to do instead is to wrap it in a function that takes a `body` parameter, applies the show rule and returns the body as it was (but with the rule applied).

elegaanz: Yes that should work.

elegaanz:
Wait, sorry, I misunderstood your previous message and I think it won't work in the end: this show rule is recursive and will be applied forever, preventing the document from being compiled within 5 iterations. And in all case, this show rule would only apply to the scope in which it is defined again.

packages/preview/vlna/0.1.0/vlna.typ
```diff
@@ -0,0 +1,6 @@
+// All characters with less than 3 letters are connected to the next word.
+#show text: body => {
```
elegaanz, line 2: This should become something like:

```suggestion
#let apply-vlna = body => {
```

(you could also name it `with-vlna` or anything you like, `apply-vlna` is only a suggestion)

And then people can use your package like this:

```typst
#import "@preview/vlna:0.1.0": apply-vlna
#show: apply-vlna

This text will automatically have the "vlna" rule applied to it.
```

PR #1697, opened 2025-02-10
elsearticle:0.4.2

maucejo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This update solves a bug when the variable journal of the template is set to none.


elegaanz:
Thanks!

PR #1733, opened 2025-02-16
Tryptophawa typsium 0.0.3 update

tryptophawa:
I am submitting
- [ ] a new package
- [x] an update for a package


Ants-Aare:
Do you know about whalogen? It already does a lot of chemistry related formatting, including isotopes, oxidation etc https://github.com/schang412/typst-whalogen/blob/master/manual.pdf
Maybe it makes sense to contribute to this project so it's all under the same roof and we don't have competing standards and package incompatibility

tryptophawa:
> Do you know about whalogen? It already does a lot of chemistry related formatting, including isotopes, oxidation etc https://github.com/schang412/typst-whalogen/blob/master/manual.pdf Maybe it makes sense to contribute to this project so it's all under the same roof and we don't have competing standards and package incompatibility

I know the project, however I decide to build one myself for kind of practice. Also, I hope my package could be easier to use for users. Additionally, I am planning to support smiles as a long term goal. 

About standard, I would probably follow walogen but maintain some difference in some labeling.

tryptophawa:
Sorry, but i put the package in wrone folder ;(

PR #1740, opened 2025-02-17
illc-mol-thesis:0.1.1

foxyseta:
I am submitting
- [ ] a new package
- [x] an update for a package

See the [changelog](https://codeberg.org/foxy/illc-mol-thesis/releases/tag/0.1.1). The original $\LaTeX$ template turned out not to be in line with the programme's official regulations, so staff updated it. I am therefore tweaking this port myself accordingly.


elegaanz:
Thanks!

PR #1743, opened 2025-02-17
zebraw:0.4.1

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

Sorry to bother! Just found that the svgs are not rendered as expected so only docs has changed.


elegaanz:
No problem, thanks for the fix. :)

PR #1747, opened 2025-02-18
basic-document-props:0.1.0

Notme112:
I am submitting
- [x] a new package

basic document with Methods to simply write header and footer, wrapper for Mail-Adress in Footer and Page Number, currentDate

I messed my fork up, so I started from scratch.


elegaanz:
I'm still not 100% convinced by the new name :sweat_smile: It still feels quite descriptive to me, even if there are enough words to make it unique. If you have a better idea and the will to rename it again, it would be nice to do so (we can also brainstrom ideas together if you need inspiration).

elegaanz:
I don't really have suggestions. If it was my package I would maybe have made a reference to a minimalist work of art, given the minimalist style of the template, which would give it a bit of personality while keeping a link with what the package does. But I think this is my personal way of naming things, not necessarily how you should proceed. Among the ones your proposed I think I like basic-document-props the best, so unless you had another idea in the meantime, renaming it for that would be fine.

Notme112:
The rename is done

elegaanz:
Thanks!

Notme112:
Thanks too :)

PR #1748, opened 2025-02-18
ape:0.3.1

gris-chat:
I am submitting
- [ ] a new package
- [x] an update for a package


gris-chat:
I wrote "string" and not "str" just out of habit...

elegaanz:
Sorry, but it is not possible to edit the code of a version that was already published, even for such a small bugfix. Can you please release v0.3.1 with this fix instead?

gris-chat:
Even if 0.3.0 doesn't work at all with this bug? (My fault, I forgot to test such a small change out of habit and overconfidence).
I feel a bit ashamed to leave a version that is completely unusable...

elegaanz:
I can understand, but we have to guarantee that the same document built with the same compiler version always gives the same output, even if this output is a bunch of compilation errors. For this reason it is not possible to update packages source code once they are published.
Most people using your package will probably not notice that the 0.3.0 was broken, since the fixed version will be the one that is shown by default on Typst Universe, so they will probably start from that / upgrade to that. And even if someone tried to use 0.3.0 they would probably notice the breakage and just upgrade. But I understand how unsatisfying it can be to know this broken version is technically still available even if nobody uses it. :smile: 

tl;dr: yes, you have to release a new version.

gris-chat:
Thank you very much. I appreciate your explanation.
Here's the new version without conflict.

elegaanz:
Thanks!

PR #1752, opened 2025-02-18
classic-aau-report:0.2.1

Tinggaard:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed a bug and updated dependencies for Typst v0.13


elegaanz:
Thanks!

PR #1755, opened 2025-02-19
shadowed:0.2.0

T1mVo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: shadowed 0.2.0 adds shadow displacement and works for both Typst 0.12.0 and 0.13.0.


T1mVo:
Whats the problem with the import being a relative path? The example source code is only visible for users that check their package cache.

elegaanz:
The idea behind this warning is to guarantee that examples can be copy/pasted as is in any project and just work. But it is no big deal if you prefer to keep it as is, I don't think a lot of people go in their package cache to copy examples anyways. Let me know if you want to change it or if I should merge.

T1mVo:
Alright, I will leave it as is. You can go ahead and merge it.

PR #1759, opened 2025-02-19
zebraw:0.4.2

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

sorry to bother again ;_;
this pr update docs about html variant and simplify line rendering logic. since there's no actual change so the version remains 0.4.1.


elegaanz:
This is a bit more than a docs update, even if the changes in your `mod.typ` file are not directly visible to the end user. Even without that, I would argue that the docs change seem big enough to warrant a new version. So I would at least ask to undo the changes to `mod.typ` if you want to get this merged as 0.4.1. Otherwise you would have to make a new release.

hongjr03:
> This is a bit more than a docs update, even if the changes in your `mod.typ` file are not directly visible to the end user. Even without that, I would argue that the docs change seem big enough to warrant a new version. So I would at least ask to undo the changes to `mod.typ` if you want to get this merged as 0.4.1. Otherwise you would have to make a new release.

okay! I'll update the version.

elegaanz:
Thank you :)

PR #1761, opened 2025-02-19
ctxjs:0.3.0

lublak:
I am submitting
- [ ] a new package
- [x] an update for a package

- implement a new api which works with the new typst 0.13
- add a new load option which allows to create a new context directly with javascript code


elegaanz:
Thank you!

PR #1766, opened 2025-02-19
harvard-gsas-thesis-oat:0.1.1

Moelf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Using new typst 0.13 features and fix various style guides.


packages/preview/harvard-gsas-thesis-oat/0.1.1/example_gsas.typ
elegaanz, line 1: Sorry, I didn't notice this last time I think, but I'm not sure what the point of this file (and the associated PDF) is. It seems to be the same content as the template, and it is not linked anywhere, so it is hard to discover for users.

elegaanz (commented):
A small remark, otherwise looks good to me :)

packages/preview/harvard-gsas-thesis-oat/0.1.1/example_gsas.typ
Moelf: indeed a good idea, I will demonstrate everything in the template going forward.

elegaanz:
Thanks!

PR #1769, opened 2025-02-20
modern-ecnu-thesis:0.2.0

jtchen2k:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This update makes the template compatible with `type' 0.13.0 and fixes multiple layout issues.


elegaanz:
Thank you :)

PR #1773, opened 2025-02-20
quill:0.6.1

Mc-Zen:
I am submitting
- [x] an update for a package

- Fixes braces in circuits which were broken in new Typst version 0.13. 
- Replaces usage of the now deprecated `path` element with the new `curve` element. 
- ⚠️ Requires Typst 0.13.


elegaanz:
Thank you :)

PR #1776, opened 2025-02-20
modern-whs-assignment:0.2.0

alex289:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update used package versions and make it compatible with Typst 0.13.0


alex289:
Mmh the warnings which make the check fail comes from the package codly. Is this a problem?

packages/preview/modern-whs-assignment/0.2.0/example/fonts/FiraCode-Bold.ttf
elegaanz, line 1: Font files should not be present in the template. What you should do instead is to add a note in your README and / or a comment in your template files to explain how to install them.

alex289:
@elegaanz it seems like the codly PR takes some time. I would like to complete this and create a fix version when codly fixed it if its possible

PR #1779, opened 2025-02-20
haw-hamburg-bachelor-thesis:0.4.0, haw-hamburg-master-thesis:0.4.0, haw-hamburg-report:0.4.0 and haw-hamburg:0.4.0

LasseRosenow:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Improve heading styles and support typst 0.13.0


elegaanz:
I don't know if this is related to the 0.13 update or to some change in your code, but I get this warning on all your templates:

```
warning: The following warning was reported by the Typst compiler: block may not occur inside of a paragraph and was ignored
    ┌─ @preview/glossarium:0.5.1/themes/default.typ:476:17
    │
476 │     align(start, it.caption)
    │                  ^^^^^^^^^^
```

Could you update glossarium to 0.5.3? It changed to use `figure.body` instead of `figure.caption`, which may solve the issue.

LasseRosenow:
@elegaanz Thank you for finding out! I have updated the dependencies now and also added a 0.13 compiler requirement to my typst.toml :)

Thank you!

elegaanz:
Looks good now, thank you.

Also, something I didn't realize when first merging these packages is that they don't really follow the guidelines for templates related to institutions or organizations: with the current names it feels like these packages are the canonical way of writing a thesis or report at HAW Hamburg. For that reason, it would be nice to either:
- rename them to something more unique. For instance you could change `haw-hamburg` for `hawburg` or `hawmburg`, or add a distinctive adjective to the current names (but they are already quite long, so adding more words may not be the best thing to do).
- or make these packages official. For that, you would have to ask the university staff to endorse them and guarantee that thesis written with it will be accepted. Ideally the option should also be presented to students along with templates for LaTeX / Word / etc. For us to be sure that the package is officially endorsed, a member of the staff will have to send an email to `hello@typst.app` stating so.

Sorry I didn't catch that in the first place, I hope that you will understand. I'll accept the current name for this time to let you decide what you want to do and eventually think about a good sounding new name.

LasseRosenow:
@elegaanz

I understand that problem, but we have talked about this before in the past and at that time you said me I could keep the name because the package is already published.

See: https://github.com/typst/packages/pull/1055#issuecomment-2402092688

I think that changing the name after so much time now would create some problems.

- The old package needs to be removed. Otherwise people will just stuck on an old version forever now knowing that that package is dead.
- People who used the old package need to migrate, but there is now way to really "inform" them about this problem
- Will the old import still work?

Regarding official support:
I have talked to the person who is responsible of the look of the HAW Hamburg thesis style and he said, that he is ok with me publishing it as long as I mark it as unofficial. That is why the description and the first line in the readme clearly states that. He was ok with that solution, but he does not intend to maintain or officially support it.

elegaanz:
My bad, I totally forgot we already had this discussion.

Renaming is not that bad:
- the old package can stay here for compatibility
- you can make a last release that shows a warning instead of the actual document (or `panic`s) to tell people to move to the new package. In the future we may have a proper mechanism to deprecate and/or rename a package.

But because I previously told you it was fine to keep the name, I will not backtrack on that, you can keep using it if you want. I will try to remember it to not ask you to rename again in three months. :sweat_smile:

PR #1802, opened 2025-02-21
simpleplot:0.1.0

Tobs2007:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: a simple package to quickly add a graph


packages/preview/simpleplot/0.1.0/README.md
elegaanz, line 1: This README is not very informative IMO, I think the contents of the help function should be moved here. If I was looking for a package to plot something, found this one and had to install it to see how it works and if it fits my need I would probably try an alternative.

Tobs2007: you are right, that is a lot better.

PR #1811, opened 2025-02-22
zebraw:0.4.3

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

this is a release mainly for fixes.

new feature: numbering-offset allows custom line number.

what's fixed:
1. fixed a special determination about empty code line that was removed by mistake in 0.4.2 (thanks for reminding me to update the version!)
2. fix the behavior when the lang parameter is true and the raw block does not carry a language attribute.

In addition, I've re-organized the file structure of the code to make it more maintainable.


elegaanz:
Thank you :)

PR #1818, opened 2025-02-23
ori:0.1.0 and theorion:0.2.0

OrangeX4:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:

[Theorion](https://github.com/OrangeX4/typst-theorion): out-of-the-box, customizable and multilingual theorem environment package for Typst.

[Ori](https://github.com/OrangeX4/typst-ori): simple enough but expressive template for notes, reports, and documents for Typst.


OrangeX4:
The ori depends on theorion, so the CI failed, but I think it should be fine.

elegaanz:
The following warning is reported when compiling `ori` on my machine:

```
`image.decode` is deprecated, directly pass bytes to `image` instead
    ┌─ @preview/octique:0.1.0/impl/octique.typ:335:8
    │
335 │   image.decode(
    │         ^^^^^^
```

Can you try to report the issue upstream and see if they can fix it? Otherwise you would have to use another icon set or maintain your own fork of octique.

PR #1834, opened 2025-02-24
flyingcircus:3.2.1

Tetragramm:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Do you want your homebrew to have the same fancy style as the Flying Circus book? Do you want a simple command to generate a whole aircraft stat page, vehicle, or even ship? I'll bet you do! Take a look at the Flying Circus Aircraft Catalog Template.


Tetragramm:
The typst checks that are failing are due to one warning inherited from cetz, and naming style differences.  This template is matching the syntax of a previous LaTeX template and so uses the naming style from that.

elegaanz:
Yes, no problem. For cetz, [it looks like a new version fixing the warning might be released soon](https://github.com/cetz-package/cetz/pull/820), so I think we should either wait for that before merging this PR, or you should publish another update before Typst 0.14 is released to update cetz (which gives you a bit of time, given that 0.13 was just released).

elegaanz:
To be honest, I saw the PR but I don't know if it will be merged soon, maybe it was opened long in advance. I will ask to not block your package for weeks if they indeed plan to release at the end of March.

Tetragramm:
Cetz's maintainer commented "Plan to make a release soon" a few hours ago, so I think it's fine to wait.
https://github.com/cetz-package/cetz/pull/820#issuecomment-2684419617

PR #1835, opened 2025-02-24
grape-suite:2.0.0

piepert:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
* add template options for fonts, date-formats and margins in page headers (detailed in README.md)
* add images to README.md


piepert:
> It would be nice if you could delete all your PDF files in the `example` directory, to avoid bloating the repository with binary files that can easily be reproduced. Other than that, everything looks good, thank you for this update!

Done, thanks for pointing this out!

elegaanz:
Thank you!

PR #1836, opened 2025-02-24
dvdtyp:1.0.1

DVDTSB:
I am submitting
- [ ] a new package
- [x] an update for a package

made it actually work lol

Description: wow colors


DVDTSB:
nvm i didnt make it actually work because it doesnt work

PR #1838, opened 2025-02-24
hei-synd-report:0.1.1 and hei-synd-thesis:0.1.1

tschinz:
I am submitting
- [x] an update for a package

Compatibility with typst 0.13.0


tschinz:
The hei-synd-thesis and hei-synd-report uses itself as a package. Since 0.1.1 is not yet published of course it can't be downloaded. I believe everything is ok here.-

PR #1845, opened 2025-02-25
mantys:1.0.1

jneug:
I am submitting
- [ ] a new package
- [x] an update for a package

### Version 1.0.1

- Compatibility with Typst 0.13


elegaanz:
There are warnings [because of codly](https://github.com/Dherse/codly/issues/65), but a fix will probably be released soon. I would advise to wait a bit and update to the new codly version. But if you'd rather have a first 0.13-compatible version with warnings and publish another patch once codly is fixed, it is possible too. Let me know what you prefer.

jneug:
> There are warnings [because of codly](https://github.com/Dherse/codly/issues/65), but a fix will probably be released soon. I would advise to wait a bit and update to the new codly version. But if you'd rather have a first 0.13-compatible version with warnings and publish another patch once codly is fixed, it is possible too. Let me know what you prefer.

Since version 1.0.0 does not compile at all, I would like this as a temporary fix and will update again when codly has a new version. Thanks.

elegaanz:
Sure :)

elegaanz:
For info, a new codly version with a fix for deprecation warnings was just released.

PR #1846, opened 2025-02-25
definitely-not-isec-thesis:2.0.0

ecomaikgolf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Unofficial ISEC TUGraz thesis


ecomaikgolf:
As there is a lot of overlap between `1.0.0/` and `2.0.0/` files I'm not sure if there's some optimization guidelines like symlinking `2.0.0` files that are the same ones as `1.0.0`, in order to reduce bundle size.

I could be symlinking `packages/preview/definitely-not-isec-thesis/2.0.0/src/assets/tuglogo.svg` to `packages/preview/definitely-not-isec-thesis/1.0.0/src/assets/tuglogo.svg` for example.

But now that I think about, I remember symlinks are not good for typst packages due to https://github.com/typst/typst/issues/5454

elegaanz:
Thank you for this update :) Duplicating files is the correct way to do it, don't worry. Git is normally smart enough to de-duplicate files with the same contents.

PR #1850, opened 2025-02-25
decasify:0.9.0

alerque:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: [Decasify](https://github.com/alerque/decasify) is a utility implemented as a Rust crate for changing the case of prose in a locale and style guide aware manor. Notably titlecasing does not just blindly change the Unicode case of the first character, but does various checks depending on the style guide. A similar package [titleize](https://typst.app/universe/package/titleize) from @SolidTux has been submitted before, but it only covers English and only uses the Gruber style. Decasify also handles Turkish robustly, and is open to supporting more languages. It also implements multiple style guides per locale. Notably it also covers other text cases (lower, upper, sentence, etc.) something that requires locale awareness to properly handle for some languages, notably Turkish and Azeri.

This package is [a thin wrapper](https://github.com/alerque/decasify/blob/master/typst/src/lib.rs) around the Rust library that [provides basic Typst functions](https://github.com/alerque/decasify/blob/master/typst/decasify.typ) and is directly managed and released in sync with the upstream repository with full feature parity with other modules including language bindings for Python, Lua, and JavaScript, a SILE typesetter module, a NeoVIM plugin, etc.

This build is generated from a new tagged release of decasify, but I haven't pushed the tag and released it upstream yet in case there are changes that need to be made here to correctly package it.

~~Note there seems to be mixed precedent for whether to exclude the package README.md file from packages. I've excluded it as many other packages to but clearly this is not universal either, so let me know if it is incorrect.~~ A CI lint here as suggested *not* excluding it. I guess that doesn't apply to legacy packages before the rule was added, hence why I saw inconsistencies in existing packages. I removed the exclude rule for this package.


packages/preview/decasify/0.9.0/example.pdf
elegaanz, line 1: Can you please delete this file? Since it is a compilation artifact it is better to not commit it.

alerque: I thought (from some docs I was following) the idea was to have this available for linking online documentation purposes. If it doesn't serve that purpose, is there even any point in keeping the `example.typ` file in the published package? Or would it make more sense to just move that example code into the README as a second example block and track a second SVG that gets embedded into the README?

alerque: Ya I have no use for generated PDFs being tracked in Git on my end upstream! I only generated because I was under the impression from somewhere that it was wanted/required. Unfortunately it was back in November when I set that up and I don't know what guide I was following. When I came along the other day to finish this up and publish it I just finished missing items, I didn't review the things I'd already speced as needing to be in the generated package.

I'll submit an update in just a minute with the example moved into the readme as a code block and the output as an SVG like the existing one.

elegaanz: No problem, thank you.

alerque: How's that now?

elegaanz: Looks good, thanks!

PR #1852, opened 2025-02-26
cineca:0.5.0

HPDell:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

- Rename parameter `stroke` to `line-style` to avoid conflict with type `stroke`
- Fix comparison between type and string


elegaanz:
Thank you.

PR #1855, opened 2025-02-26
flagada:0.1.0

samrenault:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: The flagada package display flags from countries, at a scalable height.


packages/preview/flagada/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "flagada"
+version = "0.1.0"
+entrypoint = "flags.typ"
+authors = ["renault.samuel@free.fr"]
```
elegaanz, line 5: You should put your name or username here, and have your email between braces, i.e `["samrenault <renault.samuel@free.fr>"]`.

elegaanz (commented):
It would be nice if you could take the warnings reported by CI into account. For the ones about kebab-case, you can ignore them and decide to keep snake_case instead if you have a real preference, the warning exists to make the ecosystem more consistent but it is not a hard requirement to follow that convention.

Also, I love the package name :smile:

packages/preview/flagada/0.1.0/typst.toml
samrenault: Thanks for the feedback and the actions to do. 
I followed the comments and committed an update (now in 0.1.1)

(I hope this was the way to do from my side...)

elegaanz: Since 0.1.0 was not released yet, you could have edited the files of this version directly. If you want you can move the changes here and deleted v0.1.1. But if you prefer to keep two separate versions, it's possible as well.

There are still a few warnings to fix, it would be nice to address them (the SVG file being too large is fine, I don't think there is a good way to reduce its size, the only solution I can think of would be to find an alternative version with less details).

samrenault:
Hello,
I understood the guideline and the recommendations. 
I submitted (back) only version 0.1.0. I still have 1 warning about a SVG coat of arms file, but indeed, I don't see how to reduce this version easily.
Tell me if there is still some changes to do in the proposition.

elegaanz:
Looks good now, don't worry about the large SVG file, it's not that bad. Thank you.

PR #1856, opened 2025-02-26
obsidius:0.1.0

l0drex:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Adds a new template package called _obsidius_. It features a modern style and callouts for warnings and solutions. The style and callouts are inspired by [Obsidian](https://obsidian.md). Its main purpose is the create notes of lectures, but it also fits other use cases. 


packages/preview/obsidius/0.1.0/sample/sample.pdf
elegaanz, line 1: Ideally, it would be better to not commit this file, as it can easily be reproduced by compiling the document. It would only add its weight to the git repository, which we try to keep as small as possible. The PNG version can stay as it is necessary for your README.

packages/preview/obsidius/0.1.0/typst.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "obsidius"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["@l0drex"]
```
elegaanz, line 5: If you want to have a link to your GitHub profile you should do that instead:

```suggestion
authors = ["l0drex <@l0drex>"]
```

elegaanz (commented):
Nice looking template :) Just two small remarks and it should be good to merge. Thank you!

packages/preview/obsidius/0.1.0/sample/sample.pdf
l0drex: Ok, makes sense. Can I simply push a commit removing the file, and you squash on merge? Or do I have to edit my initial commit somehow?

elegaanz: Yes, I will squash merge.

packages/preview/obsidius/0.1.0/typst.toml
l0drex: Ah, I misread the readme. Would be good to have an example there.

elegaanz:
Thank you!

PR #1871, opened 2025-02-28
quick-maths:0.2.1

EpicEricEE:
I am submitting
- [ ] a new package
- [x] an update for a package

Changelog:
- Added differentiation between italic math symbols (`$x$`) and math text (`$"x"$`)
- Fixed error with non-content shorthand replacements (e.g. symbols as in `math.plus.minus`)
- Fixed warning regarding discarding content (introduced with 0.13)


elegaanz:
Thank you :)

PR #1877, opened 2025-02-28
grayness:0.3.0

nineff:
I am submitting
- [ ] a new package
- [x] an update for a package

### Added

- Updated documentation with extensive examples
- All functions now also work with SVG images.
- `image-invert()` function
- `image-huerotate()` function
- `image-brighten()` function
- `image-darken()` function

### Changed

- Changed function-name ordering: now "image" always comes first, e.g. `image-grayscale()` instead of `grayscale-image()`. *breaking change*
- All arguments besides the imagebytes are now named instead of positional. *breakin change*
- Flipping is now done via the `scale()`-function and no longer uses the wasm plugin.
- Updated to work with Typst 0.13 without warnings.


packages/preview/grayness/0.3.0/example.pdf
elegaanz, line 1: Can you please delete this file? We try to not bloat the repository with heavy binary files that can easily be reproduced.

The PNG version can be kept as it is linked in your README.

packages/preview/grayness/0.3.0/grayness.wasm
elegaanz, line 1: Could you run `wasm-opt -Os` on this file please? It should reduce its size by more than 10%, which is worth it I think.

nineff: wow, i did not know about wasm-opt, this is indeed noticeable

packages/preview/grayness/0.3.0/example.pdf
nineff: sure.

elegaanz:
Great, thank you!

PR #1881, opened 2025-03-01
remove outdated template code from touying 0.3.2

OrangeX4:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Touying have a issue: https://github.com/touying-typ/touying/issues/159

In short, users will browse the old touying 0.3.2 template in webapp, which will cause misunderstanding, so I hope to delete this old version code of template. In my opinion, deleting these fields should not affect the normal use of users, and even if there is, it should not be used like this. So can I remove outdated template from touying 0.3.2? Anyway, thanks!

### DIFF
--- a/packages/preview/touying/0.3.2/typst.toml
+++ b/packages/preview/touying/0.3.2/typst.toml
@@ -9,8 +9,3 @@ repository = "https://github.com/touying-typ/touying"
 keywords = ["presentation", "slides", "lecture", "touying"]
 categories = ["presentation"]
 exclude = ["examples"]
-
-[template]
-path = "template"
-entrypoint = "main.typ"
-thumbnail = "thumbnail.png"


OrangeX4:
Well, actually the new version after 0.3.2 doesn't have these fields, so creating a new version doesn't solve this problem either. Maybe just let this problem exist, and it shouldn't hurt so much. So I'm going to close this PR?

elegaanz:
I'm not sure I understand correctly, but if you release a new version without these fields it should stop appearing in the template selector of the web app. Anyways, it's up to you, if you decide it is not that bad and want to leave it as is for now, you can.

OrangeX4:
touying have removed template field in versions after 0.3.2, only 0.3.2 does. Then only touying 0.3.2 appeared in the web app, resulting outdated template. So I think it's due to the logic of template code of the web app.

PR #1890, opened 2025-03-02
zebraw:0.4.4

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

Zebraw 0.4.4 fix the wrong display of comments' numbering which is caused by the numbering offset feature in 0.4.3, and adjust the numbering width to auto to contain large numbers. 🎉


elegaanz:
Thank you.

PR #1891, opened 2025-03-02
universal-hit-thesis:0.3.0

vonbrank:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

## Added
- Shenzhen campus master thesis support
- universal doctor thesis support for all three campuses
- bilingual outline
- bilingual figure
- digital signature support

## Fixed
- adjust the space between lines and paragraphs


elegaanz:
Your dependency on `algorithmic` reports deprecation warnings with Typst 0.13. It would be nice if you could fix that (ideally in this version, but if you want to get this merged quickly it can come with the next update). Unfortunately, the package seem unmaintained. You would either have to maintain a fork yourself or switch to an alternative (there a few of them I think, so this is probably the best option).

vonbrank:
> Your dependency on `algorithmic` reports deprecation warnings with Typst 0.13. It would be nice if you could fix that (ideally in this version, but if you want to get this merged quickly it can come with the next update). Unfortunately, the package seem unmaintained. You would either have to maintain a fork yourself or switch to an alternative (there a few of them I think, so this is probably the best option).

I updated our template and introduce the `algo` package instead of `algorithmic` package. Although the former has the same issue, it appears that the author is still maintaining it, and there is already a [PR](https://github.com/platformer/typst-algorithms/pull/26) addressing this problem. Once the issue is resolved, we will promptly update the version of the package referenced in our template. However, at least for now, I believe our new version of template is ready to be merged.

elegaanz:
Okay, thank you!

elegaanz:
A new version of algo with the fix was just published, if you want to start working on a patch release.

PR #1896, opened 2025-03-02
ilm:1.4.1

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog

Changed:
- Minimum required compiler version to 0.13.0.

Fixed: 
- Superfluous set rule.


elegaanz:
Thanks!

PR #1898, opened 2025-03-03
illc-mol-thesis:0.1.2

foxyseta:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
- upgraded Typst compiler to 0.13.0;
- taking advantage of Typst 0.13 to make everything look more similar to the original template. 


elegaanz:
Thanks!

PR #1901, opened 2025-03-03
tidy:0.4.2

Mc-Zen:
I am submitting
- [x] an update for a package

- Code examples can now also show code that is _not_ executed on lines starting with `<<<`. 
- The type `tiling` is now supported and it is shown with the new gradient. 
- Fixes formatting of multiline default arguments.


elegaanz:
Thanks!

PR #1916, opened 2025-03-04
flow:0.3.0

MultisampledNight:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Made compatible with Typst 0.13 (with a fallback mechanism for staying compatible with Typst 0.12) and added a few more utils (new keyword conjugations, shortcuts, styling updates, 7 segment display).


elegaanz:
Thank you.

PR #1924, opened 2025-03-05
classic-aau-report:0.3.0

Tinggaard:
I am submitting
- [ ] a new package
- [x] an update for a package

Removed warnings introduced by the 0.13 compiler and bumped minimum supported version to Typst 0.13.0

Diff: https://github.com/Tinggaard/classic-aau-report/compare/v0.2.1...v0.3.0


elegaanz:
Thanks!

PR #1935, opened 2025-03-06
student-report-aau:0.1.2

krestenlaust:
Third times the charm

I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a report template which can be used by all students at Aalborg University for their reports.

- [ ] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [?] tested my package locally on my system and it worked


elegaanz:
Hello, I'm closing this PR for now, to only keep active submissions. Feel free to open a new one when you have solved the points I raised in my previous message :)

krestenlaust:
The other project is more general and has a different philosophy, this is exactly what it is called. A template for students at Aalborg University. I don't understand the guideline of changing the title away from the purpose of the template?

Will default content still be included when people use the template? The idea is to have a template which is fully customizable (including the actual template file), not only a few parts

elegaanz:
The rule about having a non-canonical name exists to not give an unfair SEO advantage to one package or the other, and to not make an unofficial package look official. You don't have to change much of the current name to respect it: for instance the other template added `classic` as a prefix, which makes it unique and distinguishable enough to meet our requirements.

Regarding the template structure, it can be overwhelming for newcomers to have all this code in their projects, the goal of templates should be to make it possible to focus on content without worrying about the layout and esthetics of your document. Still, I understand that in some case you may want to customize some parts of the provided template. What I would advise you to do, is to have a package that exports the default pages and elements, as well as a template configuration function that takes parameters for each component as inputs (with your implementation as the default value). This way, you provide a working template without too much complexity for those who don't need it, while allowing some parts to be overridden if needed.

```typst
// in your package:
#let default-front-page(title) = align(center + horizon)[
  = #title
  AAU student report
]

#let template(title, body, front-page: default-front-page) = {
  front-page(title)
  pagebreak()
  body
}
```

```typst
// in the user's project
#show: template.with(
  front-page: (title) => {
    set text(fill: red)
    strong(title)
  }
)

= Introduction

#lorem()
```

There are other solutions, that have been discussed [in this forum thread](https://forum.typst.app/t/overriding-template-parameters-missing-social-convention-or-typst-design-flaw/2792).

When custom types and elements will be available the situation will be better, but for now we have to rely on these kind of solutions.

PR #1938, opened 2025-03-06
frame-it:1.1.0

marc-thieme:
I am submitting
- [ ] a new package
- [x] an update for a package

### Description
[The package](https://github.com/marc-thieme/frame-it/tree/main) provides the possibility to create and customize "frames" to visually emphazize content. These frames can be used to become theorems, definitions, corollaries, etc. The style is completely customizable and even for the same style, different frame types like theorems, definitions, corollaries will still be recognizable and distinct.

Looking like this:
image
### Changes
- Add new styling
- In the `(make-)frames`–function, allow supplement to be supplied as single value
  instead of array
- Pass additional arguments in a frame function onto the figure function
  when placing it in the document
- Change API to declare the styling function to use using a show rule
- Refactor layouting system to be simpler and more robust
- Make frames breakable across pages
- Add version of readme for dark mode on GitHub
- Influence the auto–generated colors for the frames using `base-color` parameter
- Improve Readme


packages/preview/frame-it/1.1.0/README.md
elegaanz, line 1: I think it is not so great from an accessibility (and SEO) point of view to have your README as SVG files. If you want to write it in Typst, maybe a better solution would be to use the experimental HTML export and save it as Markdown (using `html.framed` for the examples, so that it renders faithfully).

packages/preview/frame-it/1.1.0/.github/ISSUE_TEMPLATE/bug_report.md
elegaanz, line 1: This file and the `feature_request` one should not be committed to this repository. Can you please delete them?

packages/preview/frame-it/1.1.0/README.md
marc-thieme: I managed to get it to compile to html. However, when I paste it into the markdown README, it doesnt show any of my frames. To me, it looks like it does not render the inline-svg.
Do you have a solution for that? Would a compromise be possible releasing the README as is currently and I prepare a fix for a more accessible README as soon as possible?

elegaanz: Yes if you want we can do that. Maybe my solution actually doesn't work, it is possible that GitHub and Typst Universe remove `<svg>` tags from markdown files as they can maybe be abused.

marc-thieme: If we can do that that would me nice. I don't have time to fix it this week so it would be a blocker.
I'm pretty sure the embedded svg doesn't work. However, one could write a script to place them into external files. But that would be very fiddly of course.
What I'm going to do is add `elem` tags to try to support html natively.

PR #1944, opened 2025-03-07
zebraw:0.4.6

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

new feature: 
- adds `numbering-font-args` to customize numbering font!

bug fixed: 
- use `"\n"` instead of `linebreak()` to prevent `<br>` being selected first in html export 

THanks Ana!


elegaanz:
Thank you :)

PR #1945, opened 2025-03-07
arborly:0.1.1

pearcebasmanm:
I am submitting
- [ ] a new package
- [x] an update for a package

This update does two things:
- Fix labels with non-string content (ie. content, math, none, etc.), which previously failed to compile.
- Allow using an empty children list to stand in for not having children, which is necessary because of a current design limitation in the package where all nodes with siblings require something in the children position. Prior to this update putting an empty list there caused a number of compiler errors. 


elegaanz:
Thanks!

PR #1951, opened 2025-03-07
slydst:0.1.4

glambrechts:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Mostly bug fixes: fix typos in README, advertised template in README, fix `default-color` import, improved template when both `subtitle` and `authors` are present.


elegaanz:
Thanks!

PR #1961, opened 2025-03-08
versatile-apa:7.1.1

jassielof:
I am submitting
- [ ] a new package
- [x] an update for a package

- Typst 0.13.1
- Abstract error fix
- ORCID RegEx fix


jassielof:
Regarding the typst.toml manifest, are there plans to implement a JSON schema for it? I have been working on one for it, but where would I submit a PR for it? Here, or the typst repo?

PR #1972, opened 2025-03-10
tiptoe:0.3.0

Mc-Zen:
I am submitting
- [x] an update for a package

This release adds a `curve` function in analogy to `std.curve`. Also, all usage of the now deprecated `std.path` is replaced with `std.curve`.


elegaanz:
Thank you!

PR #1978, opened 2025-03-11
statementsp:0.1.0

Riley719:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package is useful to make mathematic documents because we can make a definition, proposition, theorem box easily. And users can make a custome box freely. For example, we can make a box with a favorite color using my function, newstatementsp. This enables beginners to make a custome box easily without being confused. Moreover, these custome box is given a counter which inherits a heading counter automatically. i.e. the statement  counter increases like "Def 1.1", "Prop 1.2", "Thm 1.3", "Def 2.1", "Lem 2.2",... Of course we can set a box without a counter. In this case, the statament counter, of course, does not increase. i.e. in this case, for example the counter is like "Def 1.1", "Remark", "Prop 1.2",... Finally, we can use a linksp function. This is for cross refferences. If you type linksp(<def:labelofdef>), you can see "Def 1.1" in your document. If you want to know more precise way to use, you should read my README.md.


elegaanz (commented):
Thank you for submitting your package. I have small remarks, but otherwise it looks good to me.

Riley719:
Thank you for your review! I decided to use ":" only in a box title. And I replaced images.

elegaanz:
Thank you!

Riley719:
Thank you too!

PR #1980, opened 2025-03-11
chordx:0.6.0

ljgago:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes in chordx 0.6.0:

- Updated deprecated syntax introduced in typst 0.13.0.
- Updated string types with the native types.
- Replaced the deprecated path function in favor of the curve function.


elegaanz:
Thank you!

PR #1990, opened 2025-03-12
parcio-thesis:0.2.0

xkevio:
I am submitting
- [ ] a new package
- [x] an update for a package

Makes the thesis template compatible with Typst 0.13.x and uses the new outline API as well.


elegaanz:
Thanks!

PR #2003, opened 2025-03-14
unofficial-fhict-document-template:1.1.3

TomVer99:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 0.13 support and slight change to TOC behaviour.


TomVer99:
I think the error comes from the `codly` package. I have imported the latest version of `codly` (1.2.0). So I am not sure if I can fix these warnings.

TomVer99:
Ah I see it is fixed in https://github.com/Dherse/codly/issues/65, but it has not been merged yet. I do not mind waiting for this to be merged and then updating the codly version in my template.

elegaanz:
The codly update is now available.

elegaanz:
Thanks!

PR #2004, opened 2025-03-14
alchemist:0.1.5

Ants-Aare:
I am submitting
- [ ] a new package
- [x] an update for a package

compiler 0.13.1 compatibility and cetz version bump to 0.3.4


packages/preview/alchemist/0.1.5/typst.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "alchemist"
+version = "0.1.5"
+entrypoint = "lib.typ"
+authors = ["Robotechnic <@Robotechnic>"]
+license = "MIT"
+compiler = "0.11.0"
```
Robotechnic, line 7: you should update the compiler version here because the module is no longer compatible with 0.11.0

Ants-Aare: good catch!
I don't think we have the compiler attribute on our other packages, which is probably why I didn't have it on my mind. maybe I should add it...

elegaanz:
Thanks!

PR #2007, opened 2025-03-14
README.md: document a command-line script to produce thumbnails

gasche:
The packaging guidelines ask for a thumbnail, but without suggesting how to produce it. It's a bit annoying to reconstruct this knowledge, and it is something that should be easily scriptable so that template authors can easily update their thumbnails as they update their templates.

(Note: I opened a forum thread with these instructions, which may suggest alternative approaches: https://forum.typst.app/t/a-command-line-and-a-makefile-rule-to-build-thumbnails/3262 )

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,19 @@ Required for submissions to this repository:
   will automatically be excluded from the package files and must not be
   referenced anywhere in the package.
 
+   If you have ImageMagick and [oxipng] installed, you could produce
+   a thumbnail file `thumbnail.png` from `foo.typ` as follows:
+   
+   ```shell
+   typst compile foo.typ
+   magick -density 250 foo.pdf[0] -flatten thumbnail.png
+   oxipng -o 2 thumbnail.png
+   ```
+   
+   (`foo.pdf[0]` selects the first page, `-flatten` uses white
+   rather than transparent background, and `-o 2` is a good
+   optimisation level for `oxipng`.)
+
 Template packages must specify at least one category in `package.categories`.
 
 If you're submitting a template, please test that it works locally on your


Ants-Aare:
I'm pretty sure you can natively export to png in typst:
`typst c file.typ file.png`
It was implemented in this pull request. https://github.com/typst/typst/pull/1159
So I would recommend to remove the imagemagick dependency from this. the png optimisation with oxiimage is a nice to have, but not really neccessary imo. As far as I know you can even export to svg in typst, so if file size is an issue, maybe use svg instead.

gasche:
How would you then select the first page only?

gasche:
Thanks for gracefully pointing to the `--pages` option, I didn't know about it.

I realize that my comparison is flawed because typst produces png with a default dpi/ppi setting of 144. If I set the `--ppi` setting to 250 as recommended by the documentation, then the output takes 2Mio, and the oxipng-optimized version takes 592Kio, which is more than twice bigger than the ImageMagick version at 264Kio.

Another way the comparison is flawed is that `typst` produces nicer-looking output than `magick` when you zoom a lot. They seem to use different rasterization algorithms and `typst` is better at reproducing the same look as with the pdf.

My impression remains that optimizing for file size is useful, so given the guidelines to target 250dpi and not below, I'll stick to `magick` for now.

> I'm not sure as to why there's a difference in file size, after optimizing both, they should have a pretty similar file size

A guess would be that typst rasterization algorithm produces smoother images that end up requiring more space.

> generally speaking you can get svgs a lot smaller than pngs though. you can read more here: https://typst.app/docs/reference/svg/

I didn't find any options in this page that would let us reduce the size of the produced svg file. Do you have a specific command to try in mind?

elegaanz:
I think I will not merge this PR because the README is getting too large, and these instructions seem a bit too specific to me (most people don't have imagemagick installed). And as said above, there is already a simple solution to generate a thumbnail with Typst, even if the size of the resulting file is not optimal. The fact that you didn't know about it makes me think that it would still be worth it to better document that, but the forum thread you opened should help with discovery of both your solution and the "Typst native" way of doing it.

I plan to rework the packaging guidelines at some point, to split it in multiple files so that it is better structured and easier to navigate and better explain how to solve some common problems. I will probably include theses instructions here (and credit you as a commit co-author), if that suits you.

Still, thank you for the suggestion!

gasche:
Fine with me! (Note that installing a common dependency is just one `<package-manager> install` away, it's much less work than deciding how to perform the conversion.)

A bit more data on possible conversion commands:

1. It is also possible to convert a PDF into a PNG using `pdftoppm`, `pdftoppm -r 250 -png -f 1 -l 1 foo.pdf thumbnail` will produce `thumbnail-1.png` for the first page at 250dpi. The result looks visually better than the `magick` PDF->PNG output, yet much smaller than the `typst` PNG output (456 Kio vs. 592 Kio).
2. Jpeg compression gives good results, using very low quality settings such as `-quality 10` or `-quality 25` will produce fairly small files; there are some visual artifacts that can be seen at higher zoom levels. My example file takes 276Kio at quality 25, and 176Kio at quality 10, compared to . (But: if we want smaller thumbnails, it is probably better to just use lower dpi settings and let clients upscale the image, the result is probably going to look better than looking at original-scale images at low jpeg quality.)

elegaanz:
Okay, closing this PR then.

Thanks for the extra information. Can PNG or WebP use JPEG compression? Thumbnails can only be in one of these two formats.

gasche:
Ah, looking at WebP is a good suggestion, I get excellent results with

```
typst compile foo.typ thumbnail.png --pages 1 --ppi 250
magick thumbnail.png -define:webp:target-size=200000 thumbnail.webp
```

The results takes 192KB and is visually undistinguishable from thumbnail.png. This is the best approach I have seen so far. (One can set target-size to a lower value, but for example a 100KB target has some small visible artifacts.)

PR #2015, opened 2025-03-15
decasify:0.10.1

alerque:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Upstream decasify crate added the option to override casing for specific words. This package bump exposes that feature as an optional named parameter to the `#decasify` and `#titlecase` functions.

PR #2016, opened 2025-03-16
vienna-tech:1.0.0

npikall:
I am submitting
- [x] an update for a package

Description:
- Separate Functions for the styling of the document and the placing of content
- Typst 0.13.0 compatible
- removed unnecessary features and simplified the codebase
- new License


npikall:
The failed test from the bot seems to originate from the "codly" package.

alerque:
That doesn't sound right. My PR #2015 passes, which has the same codly package. Besides your checks show this:

> packages/preview/vienna-tech/1.0.0/src/args.typ#L89
>
> The following warning was reported by the Typst compiler: comparing strings with types is deprecated

That sounds like *this* PR, not something in the underlying package set.

npikall:
@alerque thanks for the feedback. 
As much as I would like this warning/error to be on my side I did not find any evidence supporting that.
The Bot locates the error in the file `src/args.typ#L89`. This file does not exist in my codebase and seems to be referring to the "Codly" Package as shown in the warning below. 
I am also not sure what you mean, when writing that you use that package as well. I was looking through the PR #2015 and did not find the "codly" package anywhere. What did catch my eye was that in the [toml](packages/preview/decasify/0.10.1/typst.toml) you where setting the compiler to `typst v0.12.0`.

When compiling the [main.typ](packages/preview/vienna-tech/1.0.0/template/main.typ) file I get the following warning:
```bash
warning: comparing strings with types is deprecated
   ┌─ @preview/codly:1.2.0/src/args.typ:89:24
   │
89 │     if arg.function and function in arg.ty {
   │                         ^^^^^^^^^^^^^^^^^^
   │
   = hint: compare with the literal type instead
   = hint: this comparison will always return `false` in future Typst releases
```

This warning was raised when running:
- `typst 0.13.1` locally and
- `typst 0.13.0` on the [web app](https://typst.app)

Here is an minimal example which recreates this warning:
````typ
#import "@preview/codly:1.2.0": *
#import "@preview/codly-languages:0.1.1": *
#show: codly-init.with()

#codly(languages: codly-languages)

```python
print("Hello World")
```
````
If I overlooked something, and the warning does indeed come from my template, I would be happy to fix it (if I can locate it).

alerque:
Never mind I though you were saying something else altogether.

elegaanz:
Do you prefer to wait [for this patch](https://github.com/Dherse/codly/issues/65) to be released, or do you want to merge this now, and publish another update once codly has been fixed?

npikall:
@elegaanz  I would prefer it to be merged as soon as possible, since codly should have no effect on the template itself and is only there to demonstrate how code can be shown in the thesis. Thanks for asking though.

elegaanz:
Okay, let's merge then. Thanks.

PR #2021, opened 2025-03-16
academicv:1.0.0

roaldarbol:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: academicv is a flexible, customizable CV template. 


elegaanz:
Thank you for submitting this template. It would be nice if you could fix the reported warnings about kebab-case. If you have a real preference for snake_case, your package can also be merged as is, but kebab-case is a much more common convention in the Typst ecosystem.

roaldarbol:
Ah, fair enough - I guess its so deeply rooted I didn't even think about it. Is it recommended just for function names or for variable names too?

elegaanz:
Ahah, that works, thanks!

PR #2027, opened 2025-03-17
Always use the filesystem timestamp if Git did not succeed

reknih:
Typst's containers can now include Git. This means that package indexing fails if the mounted folder is not a repository. This PR fixes that by forcing the bundler to always try using the file system to determine the package creation time before erroring out.

### DIFF
--- a/bundler/src/timestamp.rs
+++ b/bundler/src/timestamp.rs
@@ -23,7 +23,7 @@ pub fn determine_timestamps(
         .par_iter()
         .map(|p| {
             if has_git {
-                timestamp_for_path_with_git(p)
+                timestamp_for_path_with_git(p).or_else(|_| timestamp_for_path_with_fs(p))
             } else {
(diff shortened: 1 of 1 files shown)


bundler/src/timestamp.rs
```diff
@@ -19,16 +19,24 @@ pub fn determine_timestamps(
 
     // Determine timestamp via Git. Do this in parallel because it is pretty
     // slow.
-    let timestamps = paths
-        .par_iter()
-        .map(|p| {
-            if has_git {
-                timestamp_for_path_with_git(p)
```
elegaanz, line 26: Wouldn't it be simpler and more accurate (in the case only a single call to git fails for some other reason, even if that seems unlikely) to call `.or_else(|_| timestamp_for_path_with_fs(p))` here?

reknih: Yeah, thanks!

PR #2041, opened 2025-03-19
tuhi-presentation-vuw:0.1.0

baptiste:
I am submitting
- [x] a new package
- [ ] an update for a package

Presentation slides with custom aesthetics and matching colour boxes, fonts, defaults.


packages/preview/tuhi-presentation-vuw/0.1.0/template/Virgil.otf
elegaanz, line 1: Font files should not ship with templates. As far as I can tell, it is only necessary for a slide of your template, that most people will replace with their own content anyways, so I would advise to just remove the font file, and maybe add a few other similar looking fonts to the fallback list on that slide (so that there are more chances that people have one of them already installed on their computer and that the default contents look more like what you intended).

elegaanz:
Also, maybe I don't remember correctly but wasn't the "vuw" suffix linked to your university? If so, and if this package is not related to that, maybe you should drop it.

baptiste:
> Also, maybe I don't remember correctly but wasn't the "vuw" suffix linked to your university? If so, and if this package is not related to that, maybe you should drop it.

Yes, I'd prefer to stick to the naming convention for this set of packages, so they're easier to find and recognise for my colleagues at work. The styling choices etc. are meant to be consistent across these 9 templates (but of course the uni logo etc. aren't in the template files).

packages/preview/tuhi-presentation-vuw/0.1.0/template/Virgil.otf
baptiste: OK, removed Virgil. It's a shame it's not included on the webapp by the way, I find it quite useful for the hand-drawn feel of annotations (which is why I was hoping to have it bundled in the template so users would have it on hand)

elegaanz:
Okay, no problem. Thank you!

PR #2048, opened 2025-03-19
unofficial-sdu-thesis:0.2.0

GrooveWJH:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: unofficial-sdu-thesis 0.2.0 publish !


GrooveWJH:
I need help please, the bot does not specify what the exact error is, I cannot trace the source of the error

packages/preview/unofficial-sdu-thesis/0.2.0/typst.toml
```diff
@@ -0,0 +1,16 @@
...
+name = "unofficial-sdu-thesis"
+version = "0.2.0"
+entrypoint = "lib.typ"
+authors = ["gr00ve"]
+license = "MIT"
+description = "山东大学本科毕业论文（设计）模板"
+repository = "https://github.com/GrooveWJH/typst-sdu-thesis"
+keywords = ["Shandong University", "Thesis"]
+categories = ["thesis"]
+exclude = ["fonts","template/img","template/thesis.pdf"]
```
elegaanz, line 11: You should not exclude the `template/img` directory.

elegaanz: And the `template/thesis.pdf` is unnecessary as well, since the file does not exist.

elegaanz: Actually, you can remove the whole line, since you don't have a `fonts` directory either.

elegaanz (commented):
Can you also undo the changes to the 0.1.0 version please?

packages/preview/unofficial-sdu-thesis/0.2.0/typst.toml
GrooveWJH: Very Sorry!
When you mentioned these things, I realized how ridiculous what I wrote was.😂 I manually deleted those files that I hoped to exclude before the push, which was actually a form of self-protection.

At the same time, I restored version 0.1.0.
Thank you very much for your guidance.

packages/preview/unofficial-sdu-thesis/0.2.0/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2011-2025 The Bootstrap Authors
```
elegaanz, line 3: Sorry, I just noticed, but I think you forgot to update that line after copy/pasting.

packages/preview/unofficial-sdu-thesis/0.2.0/typst.toml
elegaanz: Thank you for fixing it.

PR #2049, opened 2025-03-19
nifty-ntnu-thesis:0.1.2

saimnaveediqbal:
I am submitting
- [ ] a new package
- [x] an update for a package

Changed outline styling for 0.13. Additionally some small tweaks in styling.


elegaanz:
There are warnings reported by t4t (on which subpar depends), but [updating subpar woudln't fix the issue](https://github.com/tingerrr/subpar/issues/32) so let's merge this as it is for now. I'll let you submit a patch once subpar will have been fixed.

elegaanz:
For info, subpar 0.2.2 was just released, and should fix the warnings you had. It would be nice if you could update it for the next release.

PR #2035, opened 2025-03-18
frame-it:1.1.1

marc-thieme:
I am submitting
- [ ] a new package
- [x] an update for a package

### Description
[The package](https://github.com/marc-thieme/frame-it/tree/main) provides the possibility to create and customize "frames" to visually emphazize content. These frames can be used to become theorems, definitions, corollaries, etc. The style is completely customizable and even for the same style, different frame types like theorems, definitions, corollaries will still be recognizable and distinct.

Looking like this:
image
### Changes
- Rework README building
- Add abitlity to export to HTML
- Support dark theme for thmbox styling

- Fixing the issues brought up [here](https://github.com/typst/packages/pull/1938#discussion_r1984678521)

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name that isn't the most obvious or canonical name for what the package does
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] `exclude`d PDFs or README images, if any, but not the LICENSE

- [ ] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #2057, opened 2025-03-21
frame-it:1.1.2

marc-thieme:
I am submitting
- [ ] a new package
- [x] an update for a package

### Description
[The package](https://github.com/marc-thieme/frame-it/tree/main) provides the possibility to create and customize "frames" to visually emphazize content. These frames can be used to become theorems, definitions, corollaries, etc. The style is completely customizable and even for the same style, different frame types like theorems, definitions, corollaries will still be recognizable and distinct.

Looking like this:
image
### Changes
- Add inspection functions `lookup-frame-info` and `is-frame`
- Add suggestions for abbreviations to readme
- Fix compilation without feature flag html

In the last pullrequest ( #2035 ), I added html support to my packages.
Unfortnately, unbeknowst to me, this completely broke any compilation if `--features html` is not passed.
Therefore, this update is very important. At the moment, using my package breaks all compilation once one of my functions i used.

The reason for the breakage was that not only `html.elem` gives an error but also just using the `target()` function gives an error. To fix it, I added a further check for `--input html-frames=true` before ever calling `target()`.
Now, I also added a test case (in my [justfile](https://github.com/marc-thieme/frame-it/blob/main/justfile) ) which tests for compilation when the feature flag is not set.

### DIFF
--- a/packages/preview/frame-it/1.1.2/LICENSE
+++ b/packages/preview/frame-it/1.1.2/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Marc Thieme
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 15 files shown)


elegaanz:
Thank you, sorry I didn't catch this bug during the last review.

PR #2090, opened 2025-03-28
derive-it:0.1.3

0rphee:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- The behavior of the exported code should not change, however, additional checks are done to the input to give more informative panics (because of the lack of better stack traces, errors in the input are hard to detect: https://github.com/0rphee/derive-it/issues/5)
- The diff can be seen here: https://github.com/0rphee/derive-it/commit/7ab056c428301f227211fa47c35fe78d90c125b5


packages/preview/derive-it/0.1.3/examples/example.pdf
elegaanz, line 1: Can you please delete this file? It can be reproduced easily by compiling the Typst file and would only add its weight to the repository.

elegaanz (commented):
Just a small remark. It would also be nice if you could fix the warning about kebab-case reported in CI. After that it should be good to merge.

0rphee:
I've removed the `example.pdf` file, and the kebab-case is also fixed. As for the other warning regarding the relative import, I don't really know how to silence it... I've seen some other packages with `example.typ`'s and some use the relative import and others the package specification (though I think for this case the relative path is more appropriate).

PR #2095, opened 2025-03-29
show-the-frame:0.1.1

wisp3rwind:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:

> Show borders around the main text area, header and footer in Typst documents.
>
> This may be useful to understand and debug layout issues. The idea is similar to the LaTeX [showframe](https://ctan.org/pkg/showframe) package.

### DIFF
--- a/packages/preview/show-the-frame/0.1.1/CHANGELOG.md
+++ b/packages/preview/show-the-frame/0.1.1/CHANGELOG.md
@@ -1,0 +1,5 @@
+# [v0.1.1](https://github.com/wisp3rwind/typst-show-the-frame/releases/tag/v0.1.1)
+- Support all possible `page.margin` specifications
+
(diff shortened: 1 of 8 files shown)


elegaanz:
I feel like the package name is too descriptive, I would like you to change it please. Otherwise, everything looks good.

wisp3rwind:
Would you think that something like `scaffolder` or ` scffldr` is acceptable?

I would appreciate if the "not too descriptive" requirement could be mentioned in the README, possibly with an example that uses a pattern like the one I chose here. It used to be stated explicitly, [but it is not anymore](https://github.com/typst/packages/commit/aa4f90fba314d2cb45b05443d814561274c26f11).

elegaanz:
Yes, both would work.

I started working on a README revamp, I'll include that in the new version.

PR #2101, opened 2025-03-30
aio-studi-and-thesis:0.1.1

fuchs-fabian:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Ensures that the template works with the typst compiler version 0.13.x. The versions of the packages used have also been updated. Fixed the todo logic and the footer when thesis-compliant is set to true


fuchs-fabian:
I forgot to adjust the version in the manuals

PR #2104, opened 2025-03-30
wenyuan-campaign:0.1.1

yanwenywan:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Give more customisation options to the simple title page and fix some bugs around gradients and transparency


yanwenywan:
@elegaanz I'm not actually sure why this useless imports warning is here -- As far as I can see it's referencing calc, I'm definitely using calc and require it. Or did 0.13 make calc a part of the default namespace or something? 

Are there any other imports which are not needed since I've probably missed some subtlety in the 0.13 update, as far as I'm aware this was fine in .12 so if there's anything I should know about this would be very grateful thanks

yanwenywan:
In any case it is fixed, guess calc is now included by default

elegaanz:
As far as I remember, `calc` was always available globally. Anyways, thank you for fixing it.

PR #2110, opened 2025-04-01
quick-sip:0.1.2

artomweb:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added support for Typst tags like `<tag>` when referencing steps instead of using the step number. 
- Removed the need to pass an empty string to the step function when not using the action parameter.


elegaanz:
Thanks!

PR #2112, opened 2025-04-01
zebraw:0.5.0

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

Zebraw 0.5.0 brings a number of new features: [indentation line rendering](https://github.com/hongjr03/typst-zebraw/issues/9), line numbering separator line rendering, [slicing lines of code](https://github.com/hongjr03/typst-zebraw/issues/8). 

In addition, the code has been refactored a bit to make it easier for future development. The `zebraw-html` function has now been deprecated and combined into `zebraw` function.


packages/preview/zebraw/0.5.0/README.typ
elegaanz, line 1: I'm not sure this file and the corresponding PDF (especially the later actually) are relevant, they are not easily discoverable and only the Markdown version will be displayed on Typst Universe. I would suggest to delete them (or at least the PDF, which is redundant given that one can generate it by compiling the Typst file).

hongjr03: okay I'll delete the pdf. thanks!

elegaanz:
Thanks!

PR #2114, opened 2025-04-02
lilaq:0.2.0

Mc-Zen:
I am submitting
- [x] an update for a package

With this release, important improvements are made to the defaults, prominently including a new default color cycle.  

Instead of a somewhat arbitrary color cycle, Lilaq now uses a thoroughly researched, highly optimized color sequence that is friendly to people with color deficiencies. This map is a result of the work of M. A. Petroff (https://arxiv.org/abs/2107.02270). 
On top, Lilaq ships a small set of color sequences, available under `lq.color.map`. 

Moreover, the default axis thickness and tick length have been reduced, resulting in a better harmony with text surrounding a diagram figure. 

The mark shapes of Lilaq are carefully designed to match in optical size, so they look good together in one plot without further adaptation. This release makes some fine adjustments to the mark sizes. Furthermore, some marks have been added and a few ones renamed (see below). 

The most important fixes in this release concern a bug that made it impossible to replace the default style cycle and an issue when first-line-indent was applied to all paragraphs. 

The full changelog can be viewed at https://lilaq.org/blog/release-0.2.0.

### DIFF
--- a/packages/preview/lilaq/0.2.0/LICENSE
+++ b/packages/preview/lilaq/0.2.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Mc-Zen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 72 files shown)


elegaanz:
Thanks!

PR #2117, opened 2025-04-03
modern-xmu-thesis:0.2.0

HPCesia:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: The new version only has some minor bug fixes. But because some unused template configuration items have been removed, the version number is 0.2.0.


elegaanz:
Thank you!

PR #2118, opened 2025-04-03
IJIMAI v0.0.2

pammacdotnet:
I am submitting
- [ ] a new package
- [x] an update for a package

The dropping cap auto-detection for the first paragraph has been changed to a more conservative version.


elegaanz:
Thank you!

PR #2119, opened 2025-04-03
amsterdammetje-article:0.1.1

qu1ncyk:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- Added support for `authors` and `ids` as strings, in addition to arrays.
- Added documentation.


elegaanz:
Thank you.

PR #2129, opened 2025-04-04
parcio-thesis:0.2.1

xkevio:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixes a page numbering bug regarding the outline and the switch between roman and arabic numbering.


elegaanz:
Thank you!

PR #2133, opened 2025-04-04
callisto:0.2.2

knuesel:
I am submitting
- [ ] a new package
- [x] an update for a package

Several bug fixes, a breaking change (one function removed) and better documentation


knuesel:
This is now ready :)

(I think the first CI warning is a false positive: it says that `docs/example.ipynb` seems to be an example and should probably be added to `exclude` but the whole `docs` directory is already in `exclude`)

PR #2137, opened 2025-04-06
IJIMAI v0.0.3

pammacdotnet:
I am submitting
- [ ] a new package
- [x] an update for a package

Now Tables and Figs. have different supplements. 


elegaanz:
Thank you.

PR #2145, opened 2025-04-09
qooklet:0.1.0

ivaquero:
I am submitting
- [x] a new package

Description: This is a quick start template for scientific booklets


packages/preview/qooklet/0.1.0/typst.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+    name = "qooklet"
+    version = "0.1.0"
+    compiler = "0.13.0"
+    repository = "https://github.com/ivaquero/qooklet.git"
+    entrypoint = "src/lib.typ"
+    authors = ["ivaquero"]
+    license = "MIT"
+    description = "A quick start template for scientific booklets"
```
elegaanz, line 9: If this package is intended to be used a template, maybe it should have a `[template]` section in the manifest and a default file from which to start using it? If not, I think it would be better to change the description to something like "Utilities to typeset scientific booklets".

packages/preview/qooklet/0.1.0/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 ivaquero
```
elegaanz, line 3: The year here is outdated I think (but this is not that important).

packages/preview/qooklet/0.1.0/README.md
```diff
@@ -0,0 +1,51 @@
...
+  title: "Bellman Eqation",
+  author: "ivaquero",
+  header-cap: "Reinforcement Learning",
+  footer-cap: "ivaquero",
+  outline-on: false,
+  doc,
+)
+```
+
+## Get Started
```
elegaanz, line 37: It would be better to distinguish between this section and the previous one by giving them different titles. I also think this section should have some instructions on how to make the package available in the `@local` namespace.

ivaquero: Yes, you are right, I copied wrong section here.

packages/preview/qooklet/0.1.0/LICENSE
ivaquero: This is the start date of the repo, I'll update it

packages/preview/qooklet/0.1.0/typst.toml
ivaquero: I agree, I'll use utility instead of template.

elegaanz:
Thank you!

ivaquero:
Thank you for your detailed review and suggestions!

PR #2166, opened 2025-04-11
ijimai:0.0.4

pammacdotnet:
I am submitting
- [ ] a new package
- [x] an update for a package

Added "cite this paper as" section at the bottom of page 1. Bib style has also been changed to IEEE, and a small extra space has been added before the title of tables. 


elegaanz:
Thank you.

PR #2174, opened 2025-04-12
tapestry:0.0.2

BrandenXia:
I am submitting
- [ ] a new package
- [x] an update for a package

description: update
- add abbreviation for common math and physics symbols
- adjust header position
- enhance equation environment


packages/preview/tapestry/0.0.2/lib.typ
```diff
@@ -0,0 +1,55 @@
+#import "@preview/equate:0.3.1": equate
+
+#let tapestry(
+  title: [],
+  year: [],
+  author: str,
```
elegaanz, line 6: What comes after the semicolon is generally not a type, but the default value for this argument. I think you wanted to write something like `""` instead of `str`.

BrandenXia: yeah, thanks, i'll fix that

elegaanz:
Thank you.

PR #2181, opened 2025-04-13
leonux:1.1.0

LordBlacky:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added an option for setting the primary-color manually. 
Added section slides and changed register-method to only register section-slides instead of regular slides.
Added show: later functionality for dynamic slides.


elegaanz:
Thanks!

PR #2205, opened 2025-04-15
nup:0.1.1

baptiste:
I am submitting
- [ ] a new package
- [x] an update for a package

I stupidly tested version 0.1.0 with local images but as soon as I tried with the published `@preview` package it failed to find image files, as I was passing file paths instead of images. This update fixes it by requiring that images be passed to the function, rather than file paths.


elegaanz:
Sorry I didn't catch it during my previous review. Could you just rebase your branch so that only v0.1.1 is included in the diff please? After that I'll merge.

baptiste:
> Sorry I didn't catch it during my previous review. Could you just rebase your branch so that only v0.1.1 is included in the diff please? After that I'll merge.

Totally my fault! I should've known better, but I admit I still don't really understand why one cannot point to images via their file path and must pass images instead.

I've attempted a rebase, though I've never done that before so might have done it wrong.. It looks as though the 0.1.0 is now gone from history, which seems right.

Thanks!

elegaanz:
> Totally my fault! I should've known better, but I admit I still don't really understand why one cannot point to images via their file path and must pass images instead.

This is because Typst doesn't track in which context strings (which file paths are) are created, and thus resolves them when it reads the file. In the future, there will probably be a dedicated "path" type to address this issue.

> I've attempted a rebase, though I've never done that before so might have done it wrong.. It looks as though the 0.1.0 is now gone from history, which seems right.

Looks good, thanks!

PR #2209, opened 2025-04-16
rubber-article:0.4.0

npikall:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
  - fixed spacings for paragraphs, figures and equations
  - removed the appendix and bibliography features from the main function because of redundancy
  - fixed table cell spacings
  - changed the template and the thumbnail
  - added tests (in repo)


elegaanz:
Thanks!

PR #2217, opened 2025-04-17
messeji:0.2.0

Tanikai:
I am submitting
- [ ] a new package
- [x] an update for a package

Changelog:
- Add theme support for chat messages
- Extend user manual document
- Fix: Removed `parse-json` function that calls `json()` from the messeji.typ file, which caused problems with the path. Now, the user has to call `json()` from their own document and pass the message list to `messeji()`.


elegaanz:
Thanks :)

PR #2222, opened 2025-04-18
curryst:0.5.1

MDLC01:
I am submitting
- [ ] a new package
- [x] an update for a package

When laying out a proof tree in a block with exactly the necessary space, it was possible due to floating point approximations that we would think there is not enough space, and thus lay out the premises on multiple lines. The previous fix (https://github.com/pauladam94/curryst/pull/16) did not work in all cases. With this new approach, we test the available space against exactly what we return when there is enough space, so there should no longer be any issue.


elegaanz:
Thanks :)

PR #2223, opened 2025-04-19
tapestry:0.0.4

BrandenXia:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Add table style function, simplify codes


elegaanz:
Thank you.

PR #2225, opened 2025-04-19
modern-shu-thesis:0.3.3

XY-cpp:
I am submitting
- [ ] a new package
- [x] an update for a package


packages/preview/modern-shu-thesis/0.3.3/fonts/FangSong.ttf
elegaanz, line 1: It is not possible to distribute fonts with packages. Can you please delete this file (and other TTF files) and have instructions on how to download and setup the required fonts in the README instead?

XY-cpp: I have wrote _exclude files_ in typst.toml. I thought it will remove fonts file when merging. 

Could you please tell me what's the usage of _exclude_ ?

elegaanz: `exclude` will prevent the files from being distributed in the package archive that the compiler downloads when importing the package: these files will never reach the user's computer. However, by still copying them here, they will be added to this repository when merging, which will add a lot of weight (fonts files being relatively large) to an already heavy repository, while not being useful. For this reasons, it is better to just not commit them.

XY-cpp: Thank you. I will remove these files ASAP.

PR #2227, opened 2025-04-20
versatile-apa:7.1.2

jassielof:
I am submitting
- [ ] a new package
- [x] an update for a package

- Quotes refactor
- Abstract keywords and figure notes emph fix
- Computer code typo
- Paragraph indent on all
- Tables hline on headers
- link underline fix
- ORCID icon from png to svg
- Figure placement set to auto instead of none as default


elegaanz:
Thank you!

PR #2233, opened 2025-04-21
silky-report-insa:0.5.0

SkytAsul:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: added function to layout a new kind of sub-template (end-of-study project report)


elegaanz:
Thank you!

PR #2238, opened 2025-04-22
sdu-touying-simpl:0.3.0

Dregen-Yor:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: update a new title page.


Dregen-Yor:
It seem that these warning are caused by some dependency in the demo file.

elegaanz:
Could you update the two dependencies that are reported to not be at the latest version to see if that helps?

Dregen-Yor:
It seem that those dependencies have already be the latest

elegaanz:
`codly-languages` can still be updated to 0.1.8.

Dregen-Yor:
got it,I‘ll fix it soon

elegaanz:
Thanks!

PR #2243, opened 2025-04-23
starter-journal-article:0.3.3

HPDell:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Hot fix an error in the footnote numbering. Before, the footnote counter continuous after the affiliations. Now it starts from 1.


elegaanz:
Thank you.

PR #2258, opened 2025-04-25
machiatto:0.2.0

Mustafif:
I am submitting
- [ ] a new package
- [x] an update for a package

The next version 0.2.0 brings some better functionality to the package, the main focus bringing template support. 
- Info box added
- Code snippet added
- Template shows example on how to set up license and preface page 


packages/preview/machiatto/0.2.0/typst.toml
```diff
@@ -0,0 +1,16 @@
...
+authors = ["Mustafif", "Moka-Reads"]
+license = "MIT"
+description = "A package to help develop publications under MoKa Reads Specification"
+repository = "https://github.com/Moka-Reads/machiatto"
+exclude = []
+compiler = "0.13.1"
+categories = ["book"]
+
+[template]
+path = "."
```
elegaanz, line 14: Your template should be in its own subdirectory.

Mustafif: Okay I will make that change, thanks!

packages/preview/machiatto/0.2.0/lib.typ
```diff
@@ -0,0 +1,333 @@
...
+  ]
+}
+
+#let code-file(file-path, lang, fill, line-numbers: true, line-number-color: gray) = {
+  frame(
+    title: file-path,
+    fill: fill,
+    line-numbers: line-numbers,
+    line-number-color: line-number-color,
+    [#raw(read(file-path), lang: lang)],
```
elegaanz, line 245: `read` is relative to the file in which it is called, not to where the value representing the path is created. You should accept a string with file contents and call `read` from the template instead of taking a file path as input.

packages/preview/machiatto/0.2.0/lib.typ
```diff
@@ -0,0 +1,332 @@
...
+          )
+        } else {
+          body
+        }
+      },
+    )
+  ]
+}
+
+#let code-file(title: str, file-contents: str, lang, fill, line-numbers: true, line-number-color: gray) = {
```
elegaanz, line 239: If you want to pass the file content as an unnamed argument (making it positional), as you seem to by looking at how you call this function in your template, you should not provide it with a default value. It is then the order in which they will appear in the function call that will determine which value correspond to which argument.

By the way, in Typst, the value coming after `:` in arguments is never a type but a default value, and I don't think having `str` as a default value makes sense, you probably want `""` or `none` instead.

elegaanz:
Great, thank you!

PR #2260, opened 2025-04-25
starter-journal-article:0.4.0

HPDell:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

Breaking changes:

- Affiliation labels are now shown with alphabetic characters instead of numbers.
- Parameter `bib` is removed. Users should use the built-in `bibliography` function to create a bibliography in the right place.

New features:

- A `appendix` function is added to show contents as appendices.
- A `suffix` function is added to show additional contents before the appendix. This is useful when heading numbers are shown but need to be hide for the "suffix" part.
- A `booktab` function is added to create three-line tables.

Improvements:

- A 1em space is added beneath the abstract and keywords.
- The `placement` argument of the `figure` function is set to `top` by default.
- The separator between figure labels and captions is changed to a period followed by a space.


elegaanz:
Thank you :)

PR #2262, opened 2025-04-25
htlwienwest-da:0.3.2

peterw16:
I am submitting
- [ ] a new package
- [x] an update for a package

New:
- Adds a new listing outline, 
- omits outlines if there are any entries, 
- updates outlines to be compatible with types 0.13
- fix deprecation warnings


packages/preview/htlwienwest-da/0.3.2/lib/pages/dokumentation.typ
```diff
@@ -0,0 +1,46 @@
+#import "@preview/tablex:0.0.8": tablex, colspanx, rowspanx, cellx
```
elegaanz, line 1: From what I can tell you can remove this import, you don't use tablex anymore in this file.

peterw16: Ah yes, you’re right, thanks!

elegaanz:
Thank you!

PR #2269, opened 2025-04-27
rexllent:0.3.1

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

ReXLlenT 0.3.1 adds `formatted-cell` param for formatted number parsing, and it's still under development so only patch version is updated.


elegaanz:
Thank you.

PR #2282, opened 2025-04-28
illc-mol-thesis:0.2.0

foxyseta:
I am submitting
- [ ] a new package
- [x] an update for a package

Changelog: https://codeberg.org/foxy/illc-mol-thesis/releases/tag/0.2.0


elegaanz:
Thank you!

PR #2283, opened 2025-04-29
adaptable-pset:0.2.0

stuxf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Set A4 to the default paper size.


packages/preview/adaptable-pset/0.2.0/src/lib.typ
```diff
@@ -0,0 +1,70 @@
...
+#let homework(
+  title: "Homework Assignment",
+  author: "John Doe",
+  collaborators: [],
+  course-id: "ILY143",
+  instructor: "Prof. Smith",
+  semester: "Summer 1970",
+  due-time: "Feb 29, 23:59",
+  accent-color: rgb("#000000"),
+  paper-size: "A4",
```
SillyFreak, line 31: should be "a4" instead of "A4"

stuxf: how embarrassing! oops should be fixed now

elegaanz:
Thank you.

PR #2298, opened 2025-05-01
minimal-thesis-luebeck:0.4.0

fhemstra:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I simplified some internals, added a confidentiality notice and equation numbering and unified some fonts.


elegaanz:
Thank you.

PR #2308, opened 2025-05-04
quill:0.7.0

Mc-Zen:
I am submitting
- [x] an update for a package

Changes:

- Improvements to Tequila: 
  - Exposed a `tequila.ca` gate for arbitrary single-qubit controlled gates. 
  - Added `tequila.measure` to replace `tequila.meter`. The new gate can also receive an index of a wire to send the result to via a classical wire. 
  - Addd optional `start` and `end` parameters to `tequila.barrier` that allow local barriers. 
  - Fixed problems with `multi-controlled-gate`. 
- `targ` can now take a "target" qubit as in `targ(2)` to produce a vertical wire, just like `ctrl` and `swap`. 
- Both `ctrl` and `swap` can now be used without target argument like `ctrl()`. This can replace usage like `ctrl(0)` for gates without a control wire. 
- Added position parameters `x` and `y` to the `phantom` gate.


packages/preview/quill/0.7.0/src/Untitled-1.ipynb
elegaanz, line 1: Is there a reason for this file to be here, or was it committed by mistake?

Mc-Zen: ooops

elegaanz:
Thank you :)

PR #2312, opened 2025-05-04
grotesk-cv:1.0.3

AsiSkarp:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added the option to leave "Location" field empty for education entries
- Remove unused attributes
- Cleaned import statements
- Added instructions on how to use FontAwesome fonts with this package


packages/preview/grotesk-cv/1.0.3/src/template/info.toml
```diff
@@ -0,0 +1,69 @@
...
+[language.en]
+subtitle = "Software Engineer with a knack for human-friendly AI solutions"
+
+[language.es]
+subtitle = "Ingeniero de software con un talento para soluciones de IA amigables para humanos"
+
+[import]
+path = "../../lib.typ"
+lib-path = "../lib.typ"
+fontawesome = "@preview/fontawesome:0.5.0"
```
elegaanz, line 69: I'm really not convinced by having this in the TOML file, in my opinion it should be directly in the code. In any case, the `lib-path` needs to be fixed to use a `@preview/...` import.

PR #2313, opened 2025-05-04
casson-uom-thesis:0.1.1

ALEX-CASSON-LAB:
I am submitting
- [ ] a new package
- [x] an update for a package

There was a breaking change in 0.13 which meant the template wasn't working. This is now fixed.

- [ X ] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.


elegaanz:
Thank you.

PR #2316, opened 2025-05-05
glossarium:0.5.6

quachpas:
This is a hotfix release as I somehow re-introduced a bug in-between fixing it and releasing the new version... I really need to setup that tytanic test suite!


elegaanz:
Thanks!

PR #2320, opened 2025-05-05
basic-resume:0.2.8

stuxf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Add font configuration options


stuxf:
Do you happen to know why this is failing with "This import seems to use an older version of the package"?

PR #2323, opened 2025-05-06
unilab:0.0.3

sjfhsjfh:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update deps for newer compilers.


elegaanz:
Can you replace the `@local` import with `@preview`, and optionally update your dependencies to the latest version, please?

Issue #2024, opened 2025-03-17
Size of the typst/packages repository


gasche:
The typst/packages repository takes 1.9Gio on my machine with a current clone. As Typst gets more popular, its size will increase at a higher-than-linear speed, and there is a risk that it becomes painful in practice to operate with the package repository: at some point, people with low bandwidth will have trouble cloning the repository to contribute their own package.

The size of the repository is currently roughly:
- 500MiB of git metadata
- 1.4GiB of package data

(in particular, doing a shallow clone will not help much)

On my current checkout of the repository, there are
- 585 packages in total
- 343 packages which take less than 1Mio of disk space, they consume 117Mio in total
- 219 that take between 1Mio and 10Mio, they consume 724Mio in total
- 23 which take more than 10Mio, they consume 567Mio in total

In the short term, the following could work:
- in packaging guidelines, encourage people to stick to small packages below 1Mio (maybe templates need different recommendations)
- replace identical asset files by symbolic links, to avoid duplication of assets across different versions

Replacing identical asset files by symbolic links can be done by package authors if they are told how to do it, or by repository maintainers after the fact. (git already deduplicates its internal data, so it is not strictly necessary to do it at package-submission time.) A quick experiment suggests that doing this with the current repository should shrink its size from 1.4Gio to 947Mio, which is a sizeable win.

In the long term, I think that repository maintainers should maybe consider git-lfs or other options. The end goal would be that package authors do not need to download all other packages to submit theirs.

gasche:
>  I plan to rework the packaging guidelines at some point, and explanations on how to do that (as well as general tips on how to keep a package small) will definitely be part of it.

Would you be able to suggest a way to do this incrementally, so that we can get this useful documentation out right now, without being in conflict with your medium-plans for the README? It's a shame to find out how to generate thumbnails or cloning the repository later on, after having already done both in a non-optimal way, because the information is hard to find.

For example, an immediate plan would be to create a doc/ repository, and move the content of the current README in subfiles there (with links from the main README). I would think of the following separate topics:

- submitting-packages.md: the process of how to submit a new package
- packaging-guidelines.md: general packaging rules and recommendations
- package-content.md: recommended content of a package (README, thumbnails... including the recommended filesystem layout)
- typst.toml.md: how to write typst.toml
- local-packages.md: local packages / how to test packages locally

[...]

PR #2164, opened 2025-04-10
Reorganize the docs

elegaanz:
The main README is now much shorter and redirects to the `docs/` folder, in which multiple files guide package authors through the process of creating and publishing a package. I tried to document all common mistakes and questions I have encountered while reviewing packages, while making it easy to find a specific piece of information if you know it exists in the docs, and trying to not overwhelm the reader with too much unnecessary information. I'm not entirely sure I met these goals, feedback on the structure is welcome (and in general, feedback is welcome).

This includes #1036 

[Rendered](https://github.com/typst/packages/blob/da930735a6d215c808efdebfdebefd0dea5b61ca/README.md)

PR #2340, opened 2025-05-09
run `fclones` to replace duplicated files by symbolic links

gasche:
(See #2024 for my preliminary issue discussing the size of the typst/packages repository.)

[fclones](https://github.com/pkolaczk/fclones) is a fast file-deduplication utility. It has enough options to support exactly the needs of reducing file duplications in the typst/packages repository: every once in a while, repository maintainers can run this, and this will replace certain files by symbolic links.

This is not a satisfying fix to the issue of ever-growing package repository by any metric, but it is a very simple change with no cost/downside that I can see, which makes a significant difference in total file size.

When run on the current `main` branch, the command:
- takes less than a second on my machine (SSD disk)
- reclaims 590MB of disk space in the working copy (out of 1.7G; in addition, the .git data takes 570MB)

```
# before the PR
[gasche@framawork typst-packages (main)]$ du -sch .
2.3G	.

#after the PR
[gasche@framawork typst-packages (fclones)]$ du -sch .
1.7G	.
```

Command:

```
fclones group packages | fclones link --soft --priority bottom
```

Explanations:

- running on the `packages/` subdirectory is important, otherwise the toplevel file LICENSE gets replaced by a symbolic link!

- `--priority bottom` means that files lower in the `fclones group` output will be replaced by links to files higher in the output. The sorting method for the `flcones` output appears to perform a natural numeric sort on version numbers, so this gives the important property that we turn files in newer versions into links to older versions, and not the other way around. This is important if we run this deduplication script several times, after adding newer versions.

#### is it enough to deduplicate files in the working copy only?

My understanding is that git already performs some deduplication in its internal data, and this is why the `.git/` subdirectory of a typst/packages clone is much smaller than the working copy.

This means that:
- the present PR does not reduce the data that needs to be downloaded when cloning the repository
- ... but also that it is effective (at reducing the total filesystem size) even when run after the fact, we don't need to ask each PR author to run it, it is fine if the non-deduplicated version existed in the past git history

#### What happens on systems that don't support symbolic links well?

Git recognizes symbolic links and stores them in a special way in its internal data. When cloning the repository on systems that don't support symbolic links well (eg. some Windows configurations), it will follow the link itself and just put a copy in the filesystem. This means that those users will see the pre-deduplication filesystem usage, but everything should still work for them.

#### We need to run this command again from time to time?

Yes. There is nothing automatic in this PR, just run the same command again in six months. Nothing bad happens if you forget.


gasche:
(When I said that "nothing broke", it looks like the CI has trouble with the resulting repository, and I need to investigate what is going on. Maybe the `fclones` script did something wrong, maybe the CI tools should be better at following links. Please feel free to give you opinion on the overall approach in the meantime.)

gasche:
The reason the CI fails is that the "bundler" test is running in `git sparse-checkout` mode, with a sparse-checkout cone populated by the directories that have been modified in the pull request being checked. This fails if the directories that have been modified contain symlinks to other directories that have not been modified: after `sparse-checkout` those symlinks become dead links.

It would be nice if sparse-checkout would follow symlinks so that this just works. I wrote to the Git mailing-list to suggest this, but this will not solve the short-term problem that my proposal breaks the CI.

I think that one of the two following fixes would be reasonable:

- 1. could write a more clever command that would guarantee that symlinks are only created between versions of the same package. (Currently there are a few cases where files are copied from a package to another.) Then I would widen the sparse-checkout logic to always include all versions of a modified package, instead of just the modified versions(s).

2. The reason `sparse-checkout` is used, I believe, is that the `bundler` script run by the CI does not take a list of packages to process as input, it blindly processes all packages present in the `packages/` directory. (This bundling process takes ages due to the fact that it re-encodes everything at high WebP quality.) After `sparse-checkout`, the bundler script sees only a few packages and it processes only those.

   We could instead fix the `bundler` script to take a list of packages to process as input, and then the script would run from a normal (non-sparse) checkout, with the list of packages to process being passed explicitly by the CI script.

(2) sounds like the better option by far, and I volunteer to implement it.

But ideally I would like to hear from repository caretakers before doing this work: are you interested in this symlink-creation approach to reduce the total repository size?

elegaanz:
Hello, and thank you for putting effort into this. I would like to discuss with the rest of the team what their opinion on the matter is, but I'm personally worried that this may cause issues that we may not yet think of. Still, if we decide to go in that direction, I think implementing both solutions you offered to fix the CI would be nice: some people use a sparse checkout locally, and implementing (1) would allow them to continue using this workflow, while (2) has the benefit of making it easier to repackage only a specific package, which is also useful when testing the bundler locally (and that we may want in the future to make package index deployment faster, as it currently re-bundles and deploy everything after each package update).

Another issue that I noticed and that would need fixing is that the symlinks are absolutes and include directories that only exist on your machine. They should be relative to the root of the repository or of the `packages/` directory.

I will let you know what we decide with other members of the team.

elegaanz:
A question that we have after discussing this a bit is whether there is anything unsatisfying to you with using a sparse checkout? We think that it solves the size problem in a nicer way than adding symlinks everywhere, and we are sure that it won't break anything in our current setup, which is a concern we have with symlinks.

gasche:
sparse-checkout is an arcane feature of git whose usage is not documented anywhere that I could find in Typst documentations. One advantage of my approach is that the space-savings apply to everyone, including all the package authors that haven't heard of sparse-checkout.

This being said, with good/clear documentation on how to use sparse-checkout to work with the Typst packages repository, I think that this would also be a reasonable approach.

elegaanz:
It will be documented with https://github.com/typst/packages/pull/2164, feel free to make suggestions there if you would have done it differently or see things that could be improved. I will close this PR for now, as we would really prefer to encourage sparse checkouts.

PR #2347, opened 2025-05-10
Deprecate `prooftrees` package, and maybe the name?

david-davies:
I am submitting
- [ ] a new package
- [x] an update for a package

`prooftrees` is no longer maintained, and users should be forwarded to `curryst` instead.

Another point:
Since I first submitted the package, a new naming guideline for packages asks that packages not have the 'obvious' or 'canonical' names for their functionality.
I believe this package breaks that guideline, as new users will likely unwittingly find `prooftrees` before the other more useful (and maintained) libraries.

I'm not sure what the best solution to this would be.

Thanks,
David

### DIFF
--- a/packages/preview/prooftrees/0.1.0/README.md
+++ b/packages/preview/prooftrees/0.1.0/README.md
@@ -1,3 +1,13 @@
+# Prooftrees [DEPRECATED]
+
+Deprecated in favour of [curryst](https://github.com/pauladam94/curryst). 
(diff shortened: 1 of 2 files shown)


elegaanz:
I think the package can just stay there with this deprecation notice, we don't really have a dedicated mechanism for this situation yet, users will just have to click one more time before finding `curryst`. In the future, if we have a proper way of marking a package as deprecated, we will use that for this package.

PR #2351, opened 2025-05-11
upd 0.3.1

Dregen-Yor:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: upd the outline and footer.


elegaanz:
Thanks! For info the `cetz` warning is coming from `timeliney` which isn't yet using the latest version, so not much you can do on your side for now. I'll merge as is.

PR #2361, opened 2025-05-12
callisto:0.2.4

knuesel:
I am submitting
- [ ] a new package
- [x] an update for a package

More flexible handling of code cell metadata:

* A space is now allowed between `#` and `|` so one can write `# | label: x` instead of `#| label: x` (useful for code formatters and linters).
* The header pattern can be configured with the `cell-header-pattern` keyword argument.
* The header can be preserved in the source and rendering using `keep-cell-header: true`.


elegaanz:
Thank you.

PR #2365, opened 2025-05-12
ulb-phd-cover:0.1.0

mo-nonym:
I am submitting
- [x] a new package
- [ ] an update for a package

This is a typst transcription of my university's standard PhD cover. The files provided by the institution are currently either a .docx or a .tex file. A typst template fills the gap between these two formats. Only the cover is standardized; the formatting of the manuscript is left to the PhD candidate. This template is intended for generating a personalized PhD cover as a single-page PDF file.

- [ ] have chosen a license and added a `LICENSE` file or linked one in my `README.md`


mo-nonym:
Hello, and thanks for your comment !

For now the package is indeed not official:
+ The template provided by my institution are extremely buggy. I only heard of people ending up imitating the cover of an existing thesis as close as possible than actually making use of the provided official template files. Just to say that the word 'official' does not have much meaning here. It's more about a general appearance (;
+ As I know my institution, there's no way that the package woud become official without them seeing it in action. Therefore, my plan was to do my best and publish it (as unofficial, as you suggested) and then show them the benefits in action. No idea who that would be, but what would then be the process to mark it eventually as official ?

For the name, I was doubting. The whole university uses the same template replacing only the faculty and funding logo. The acronym of the university is ULB. Then, the format of the thesis is free, the template only applies to the cover of a submitted thesis. So I came up with the name `ulb-phd-cover`. In the end, I would propose `ulb-phd-cover-unofficial`.

Maybe there's something I don't understand about the naming conventions of typst packages. You want a proper noun ? In that case I can come up with some things losely connected to the university's motto (_Scientia vincere tenebras_) like the word _lumen_ or _spectra_… And I put the context and info of the package in the README.

elegaanz:
Hello and thank you for all these information. `ulb-phd-cover-unofficial` would meet our guidelines, but still feels a bit descriptive in my opinion. I would encourage you to incorporate another noun to the mix ("lumen" or "spectra" both sound good), it would give your package name a bit more personality. All information about what the package is should not be present in its name, putting the additional context in the README and/or the `description` field is actually a good practice.

mo-nonym:
I changed the name simply to `lumen` and moved all other info (+unofficial) to the description and README.

elegaanz:
Looks good, thank you!

PR #2374, opened 2025-05-13
shane-hhu-thesis:0.4.0

shaneworld:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

- Fix: no line break problem after numbering.
- Update outdated content in README and improve local deployment steps.
- Add reference format.
- Provides a more flexible label-content box, supporting adaptive length and manual line wrapping.


elegaanz:
Thank you.

PR #2385, opened 2025-05-15
cyberschool-errorteaplate:0.1.5

ErrorTeaPot:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I have added parameters in the logos grid to center them vertically


elegaanz:
Thanks :)

PR #2389, opened 2025-05-16
tiptoe:0.3.1

Mc-Zen:
I am submitting
- [x] an update for a package

- Fixed tiptoe for right-to-left text flow. 
- Internally changed the implementation of `arc` to use `curve` instead of `path`. Consequently, the default of `arc.shorten` is now `100%` which looks better with the new implementation.


Mc-Zen:
Weird, I ran the package checker locally and there were no issues.

PR #2390, opened 2025-05-16
apa7-ish:0.2.1

mrwunderbar666:
I am submitting
- [ ] a new package
- [x] an update for a package

**Changelog:**

- improved and tweaked spacing of headings, lists, and figure captions
- make figure/table captions in italics so that the design is closer to APA7 
- added more content to example (e.g., one figure)


elegaanz:
Thank you!

PR #2400, opened 2025-05-17
shane-hhu-thesis:0.5.0

shaneworld:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed reference format issues and mathematical formula display issues.


elegaanz:
Thank you.

PR #2403, opened 2025-05-18
grape-suite:2.1.0

piepert:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- add more templates (subtypes)
- add more customization options
- some code restructuring


piepert:
Clarified the versioning principles in Readme.md, therefore discarding this in favor of a version 3.0.0.

PR #2410, opened 2025-05-18
algorithmic:1.0.0

quachpas:
Hello! 

@lf- has transferred maintenance of the repository to typst-community, and I am taking over as the new maintainer, cf. https://github.com/orgs/typst-community/discussions/14. 

This is a v1.0.0 release after a few features were added. There wasn't much to add from what was already existing.

All images included in the patch are used in the README.md. If there is one thing I am unsure about is how the table will look like in the universe website. It renders OK on GitHub.


quachpas:
I think the [validate_github_handle](https://github.com/typst/packages/blob/a7df17259bb8e884ecdaba6035588628c97cac9e/bundler/src/author.rs#L42) is in the wrong here? I mean, https://github.com/lf- clearly ends with an hyphen!

elegaanz:
Given that this check is still valid almost all the time, you could use the full URL to the GitHub profile between the chevrons. It should prevent an issue from being raised, while keeping the link the same in the end. Would that work for you?

PR #2424, opened 2025-05-20
peace-of-posters:0.5.6

jonaspleyer:
I am submitting
- [ ] a new package
- [x] an update for a package

This PR includes minor bugfixes for the `peace-of-posters` package and template.


packages/preview/peace-of-posters/0.5.6/templates/pop-uni-fr-poster/main.typ
```diff
@@ -0,0 +1,130 @@
+#import "@preview/peace-of-posters:0.5.5" as pop
```
elegaanz, line 1:
```suggestion
#import "@preview/peace-of-posters:0.5.6" as pop
```

packages/preview/peace-of-posters/0.5.6/templates/pop-uni-fr-poster/thumbnail.png
elegaanz, line 1: Unless there is a good reason for this file to be duplicated, only one of the two should be kept, especially given their size.

jonaspleyer: I had a bug-report in which the user claimed that the thumbnail image was missing when initializing the template.
https://github.com/jonaspleyer/peace-of-posters/issues/19
It seems that the typst packaging system does not copy this thumbnail image when it is also used within the template itself. For this reason, my workaround is to keep another copy of the same image. If you know of another way to circumvent this, I am very open as I am not keen on keeping two identical copies.

packages/preview/peace-of-posters/0.5.6/templates/pop-uni-fr-poster/main.typ
jonaspleyer: I have corrected this in my updated PR. Thanks for the heads up! :pray:

packages/preview/peace-of-posters/0.5.6/templates/pop-uni-fr-poster/thumbnail.png
elegaanz: Ah, sorry, I didn't notice that you were using the thumbnail in the template itself, this is a bit uncommon. No problem then.

elegaanz: It would still be nice if you could reduce the size of the thumbnail a little using some compression tool.

And for info, the reason the bug happened is that thumbnail files are automatically excluded from the package archive even if they are not in the `exclude` array of the manifest.

jonaspleyer: > It would still be nice if you could reduce the size of the thumbnail a little using some compression tool.
> 
> And for info, the reason the bug happened is that thumbnail files are automatically excluded from the package archive even if they are not in the `exclude` array of the manifest.

Yes that's what I thought. I was about to write a bug-report for this in this repo but then thought that this was probably done purposefully. Do you think that it would be a good idea to introduce a flag which includes the thumbnail explicitly?

elegaanz:
Thank you.

jonaspleyer:
> Thank you.

Thank you and the team behind typst! :pray:

PR #2446, opened 2025-05-26
oxifmt:0.3.0

PgBiel:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added thousands separators with `fmt-thousands-separator: ","` (for example) and support for numeric formatting of fixed-point `decimal`. Fixed some type checking code incompatible with Typst 0.14.0 (edge case and no existing packages appear to trigger it from a quick search, so not too critical).

**Breaking change:** Format specifiers can no longer start with `fmt-`, as that prefix is reserved for oxifmt's own options.

Also updated license to MIT OR Apache-2.0 (previously MIT only).

### DIFF
--- a/packages/preview/oxifmt/0.3.0/LICENSE
+++ b/packages/preview/oxifmt/0.3.0/LICENSE
@@ -1,0 +1,1 @@
+MIT OR Apache-2.0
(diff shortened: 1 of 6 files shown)


elegaanz:
Thank you! :)

PR #2448, opened 2025-05-27
jsonpath:0.1.0

qjebbs:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: json_path extracts values from dictionary or array using a JSONPath expression as per [RFC 9535](https://www.rfc-editor.org/rfc/rfc9535.html), except the filter syntax is different.

Extract all titles of books from the store.

```typst
#{
  let obj = json("rfc9535-1-5.json")
  let result = json_path(obj, "$.store.book.*.title")
}
```

Filter elements based on a condition using a custom filter function.

```typst
#{
  let obj = (o: (1, 2, 3, 5, (u: 6)))
  // equivalent to "$.o[?@<3, ?@<3]" which is not supported.
  let result = json_path(
    obj,
    "$.o[?,?]", // or "$.o[?0,?0]",
    it => type(it) in (int, float) and it < 3,
  )
}
```

### DIFF
--- a/packages/preview/jsonpath/0.1.0/LICENSE
+++ b/packages/preview/jsonpath/0.1.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Jebbs
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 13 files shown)


elegaanz:
That's a good reason, but I think it should be on VS Code to know that having a dash in an identifier is valid in Typst and select the whole name. I wonder if that's something the language server can control. Anyways, thank you for making the change.

qjebbs:
Thank you all for the wonderful work to Typst.

PR #2452, opened 2025-05-27
simple-tubs-letter:0.1.2

Cangarw:
I am submitting
- [ ] a new package
- [x] an update for a package

The user can now choose their font by defining it in the metadata.typ. This is necessary for Windows user, as the standard font of this package is missing for a standard installation.


elegaanz:
Thank you for this update. :)

For info, it is possible to give multiple fonts to `text.font` (if the first one is not found, it will try the next in the array, and so on), maybe you could provide a default that works on Linux and fallback to a similar looking Windows font? But letting the user customize their font is a nice feature to have as well, I'm just suggesting that in case you want to provide better defaults that would work for everyone without having to edit `metadata.typ`.

Also, before merging, could you please update the import in `letter.typ` to 0.1.2?

Cangarw:
Thank you very much for this suggestion! I will fix it this week and also include the correct version in the letter.typ!

Cangarw:
I fixed the old version number in the `letter.typ` and added an array of fonts in the template. Should be good to go now. 

Thanks again for the suggestion!

elegaanz:
Thanks!

PR #2462, opened 2025-05-29
silky-report-insa:0.5.1

SkytAsul:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
- Added `defense-date` parameter to the `insa-pfe` template
- "Thanks" and "Autorisation de diffusion du rapport" pages are no longer outlined
- Figures with the "table" kind now automatically has their captions at the top


elegaanz:
Thank you.

PR #2470, opened 2025-05-29
rowmantic:0.1.0

bluss:
I am submitting
- [x] a new package
- [ ] an update for a package

The idea is a row-oriented way to input tables, with just a little less syntactical overhead than the usual table function in Typst.

For me personally, this is the way I want to input tables for typst, it's easier to input and more convenient while editing. Depending on type of table, of course, it doesn't fit every use case.

The readme (included, but also readable at https://github.com/typst-community/rowmantic ) gives a good example of how it works. (~~I think the way I've linked the image paths in the readme would work, I just looked at what package zero does, but I'm not sure how it will work on typst universe~~  This is now addressed by including the svgs in the package, but they are excluded from compiler download. After talking with Mc-Zen. There are multiple options, but went with this one.).


bluss:
Ok, hold on. I can reduce the size of the svgs to 50% with `scour` (down total 128 kb). Maybe that's going to be acceptable?

bluss:
I'm sorry for the many updates. I guess it comes with making an initial release, and that I don't consider it being released until it's on typst universe.

The reason for the update was that it now has a change so that the SVGs are 50% the size of what they were before, the size of the svg files is now around 128 kb total (4 files for the readme).

Other changes were also included compared with the initial posting. If you need me to squash the changes, I can do so but I assumed you squash merge all changes anyway.

- Function doc in README
- supports table.header, table.footer with rows now
- SVGs are reduced in size by 50% by optimizing

elegaanz:
No worries, the commits will indeed be squashed before the merge. And thank you for optimizing the images!

bluss:
Thanks! Nice to see this "launch" :slightly_smiling_face:

PR #2474, opened 2025-05-30
vibrant-color:0.2.1

SHAfoin:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed some bugs people reported and updated the quality of the assets.


elegaanz:
Thank you!

PR #2481, opened 2025-06-01
Delete packages/preview/modern-shu-thesis directory

XY-cpp:
It is moved to https://typst.app/universe/package/shuosc-shu-bachelor-thesis/


elegaanz:
It is not possible to delete a package once it was published, sorry. Some older projects may still depend on it. What you can do if you want to make it clearer that this package has been renamed, is to publish a new version that just changes the README to add a paragraph about the name change.

PR #2483, opened 2025-06-02
structogrammer:0.1.2

genericusername3:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This package generates Nassi-Shneiderman diagrams (or structograms) sometimes required by German schools. It had some issues with align when used in figures and tried to place grid cells atop one another in a rare function. Both should now be fixed.


elegaanz:
Thank you!

PR #2491, opened 2025-06-03
impressive-impression:0.1.0

JeppeKlitgaard:
I am submitting
- [x] a new package
- [ ] an update for a package

I have made what I reckon is a rather snazzy CV template and figured I would share it with the rest of the Typst Universe.

It features support for theming alongside 1 column or a split layout along with support for multiple pages.

Contrary to some other CV templates, it is made in such a way that the main layout blocks of the document can be any arbitrary `content`, which allows users to leverage the flexibility and power of Typst. To aide in setting up these blocks, a few helper functions and elements are given in the associated package.

Template asset directory does include some flags, which are all MIT licensed.


elegaanz:
Thank you.

PR #2494, opened 2025-06-03
rowmantic:0.1.1

bluss:
I am submitting
- [ ] a new package
- [X] an update for a package

Description: The idea is a row-oriented way to input tables, with just a little less syntactical overhead than the usual table function in Typst. 

Changed in v0.1.1:

- `table` parameter for `rowtable`, to configure which table function to use;
  this allows using rowtable with other table wrappers like `zero.ztable`.
- `table.header` and `table.footer` can take a single `expandcell` as their
  content.

Also new keyword `rowtable` (for the most used function function), minor edits to docs.


elegaanz:
Thanks!

PR #2497, opened 2025-06-04
oxifmt:1.0.0

PgBiel:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- **Breaking changes:** Replacement names can no longer contain braces for consistency with Rust (`{a {{ b}` and `{a }} b}` are now errors); all private symbols are now hidden (only the `strfmt` function can be imported).
- Added variably-sized thousands for insertion of separators through `fmt-thousands-count: (group 1 size, group 2 size, ..., remaining group sizes)`, e.g. `strfmt("1000000", fmt-thousands-count: (3, 2), fmt-thousands-count: ",")` outputs `10,00,000`
- Fixed bugs with fixed-point decimal formatting and fixed inconsistencies compared to Rust
  - A relevant one is that center-aligned padding can now not be perfectly centered if an odd number of pad fill characters is needed, in order to ensure exactly the requested width is what is returned at the end.

### DIFF
--- a/packages/preview/oxifmt/1.0.0/LICENSE
+++ b/packages/preview/oxifmt/1.0.0/LICENSE
@@ -1,0 +1,1 @@
+MIT OR Apache-2.0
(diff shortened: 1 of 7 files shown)


elegaanz:
Thanks!

PR #2512, opened 2025-06-06
tfguf:0.0.3

pammacdotnet:
I am submitting
- [ ] a new package
- [x] an update for a package

Added "dedication" section, styled URLs and SVG logo. 


elegaanz:
Thanks.

PR #2516, opened 2025-06-07
nutshell:0.1.0

baptiste:
I am submitting
- [x] a new package
- [ ] an update for a package

Template for 2-page resume with entries specified as definition lists, formatted as a table.

I picked example SVG icons from open source https://thenounproject.com and kept the attribution as required, so I think this is OK.


baptiste:
The warnings are about `LaTeX` and `TeX` macros seemingly being in CamelCase, but I would argue that's preferable here, since it maps to the standard way of writing these. I'm not using them in the dummy examples, but in my own actual CV it's handy to have them available.

elegaanz:
Yeah, it makes sense to keep them as is. Thank you for making this template, it looks nice :)

PR #2521, opened 2025-06-09
clean-dhbw:0.3.1

roland-KA:
I am submitting
- [ ] a new package
- [x] an update for a package

- fix to issue [clean-dhbw-typst-template#11](https://github.com/roland-KA/clean-dhbw-typst-template/issues/11) (pass signature as image)


elegaanz:
Thank you.

PR #2523, opened 2025-06-09
cetz:0.4.0

johannes-wolf:
I am submitting
- [ ] a new package
- [x] an update for a package

- Fixed a bug where passing a length as radius to a polygon resulted in an error
- Added `ctx.shared-state`, a context field to share canvas-global state
- **BREAKING** Removed the `flex` attribute for marks (default to `true`)
- Changed the internal path model to support compound paths (see new `compound-path` function)
- Braces are now drawn tapered by default
- Brace styling changed, see the documention of `decorations.brace`
- **BREAKING** Removed support for using ratios as canvas lengths. To get back the removed behavior, wrap your canvas in a `layout(ly => canvas(length: ly.width * <ratio>, ...))`
- Added support for specifying a baseline offset via `canvas(baseline: <coordinate>, ...)`. The canvas switches to an inline `box` as the top-level element if set
- Added support for passing a stroke style to the canvas top-level element, see `canvas(stroke: <stroke>, ...)`
- Added support for user-defined coordinate systems. See `register-coordinate-resolver(<callback>)`
- The default style value for `mark.transform-shape` changed to `false`


elegaanz:
Thank you.

PR #2524, opened 2025-06-09
rowmantic:0.2.0

bluss:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: A convenient, row-oriented way to input tables.

- `rowtable` now splits equations as rows, using the default separator `&` (the alignment marker) as cell separator. New examples added for this.
- `rowtable` parameter `separator-eq` to control the separator used for equations
- `rowtable` now supports table cells with `rowspan`
- `rowtable` parameter `column-width` allows configuring column widths without restricting column count


elegaanz:
Thank you!

bluss:
Thank you for the constant work on the packages :slightly_smiling_face:

PR #2535, opened 2025-06-12
marginalia:0.2.0

nleanba:
I am submitting
- [ ] a new package
- [x] an update for a package

## What's Changed
* https://github.com/nleanba/typst-marginalia/pull/9
    - notes (and notefigures) take a `block-style` parameter, which may be a function now. If it is a function, it receives the side ("left" or "right") of the note.
    - notes, notefigures, and wideblocks no longer take a `reverse` (or `double`) parameter, but a unified `side` parameter that can be `auto` = `"outer"`, `"inner"`, `"left"`, `"right"`, or `"both"` (the last one is wideblock-only).

        → use `side: "inner"` instead of `reverse: true`
        → use `side: "both"` instead of `double: true` with wideblocks

* https://github.com/nleanba/typst-marginalia/pull/10
    - New setup:
        ```typ
        #show: marginalia.setup.with(
          // inner: ( far: 5mm, width: 15mm, sep: 5mm ),
          // outer: ( far: 5mm, width: 15mm, sep: 5mm ),
          // top: 2.5cm,
          // bottom: 2.5cm,
          // book: false,
          // clearance: 12pt,
        )
        ```
    - Options `numbering` and `flush-numbering` have been moved to the `note()` and `notefigure()` functions.

**Full Changelog**: https://github.com/nleanba/typst-marginalia/compare/v0.1.4...v0.1.5


elegaanz:
Thanks!

PR #2539, opened 2025-06-12
rose-pine:0.2.1

oplik0:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: the update replaces the deprecated `path` usage with `curve`, resolving a deprecation warning


packages/preview/rose-pine/0.2.1/example.typ
elegaanz, line 1: To make this files (and the rendered PDF) more useful I would link to them in the README, as it is they are hard to discover, nobody browses this repository by hand I think. Alternatively, can you delete them both in this copy of the project? You can of course keep them in your own repository, people sometimes explore files here.

oplik0: Sorry for taking long, was a bit busy and it slipped my mind :)

Removed the files from here - there are screenshots in the README and I feel it's fairly standard to just go to the repo if you want to see *how* something in a package works - and for updates it'll be easier to remember to remove the additional files (since there already are files removed, e.g. GH Actions workflows) than to modify the README beyond bumping the version

elegaanz:
Thank you!

PR #2544, opened 2025-06-15
statementsp:0.1.1

Riley719:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I changed easy bugs and update this package. In the former version, this package missingly give you an error when you properly set the box label and the number key. I fix this bug. And I fixed README.md because there were wrong English. 


packages/preview/statementsp/0.1.1/typst.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "statementsp"
+version = "0.1.1"
+entrypoint = "statementsp.typ"
+exclude = ["/images/*","/examples/*"]
```
saecki: just changing this to `["images", "examples"]` should work

packages/preview/statementsp/0.1.1/examples/example.pdf
saecki, line 1: Could you delete this file? You seem to have added this example, but don't link to it anywhere in your readme. This will be very hard to discover for users and can also be reproduced by compiling the typst file.

packages/preview/statementsp/0.1.1/typst.toml
```diff
@@ -0,0 +1,9 @@
...
+authors = ["Riley"]
+license = "MIT"
+repository = "https://github.com/Riley719/typst-statementsp"
+description = "Happy statement box and its cross referrence system."
```
saecki:
```suggestion
description = "Happy statement box and its cross reference system."
```

saecki (commented):
Thanks for this update! I've found a small typo in the description and a file that probably isn't needed in this repository. There are also some warnings from the `typst-package-check` that should be resolved.

Riley719:
Thank you for your review! I fixed typo and deleted examples. Could you check it?

PR #2547, opened 2025-06-15
matador:0.1.0

AcqRel:
I am submitting
- [x] a new package
- [ ] an update for a package

This is a parser for .mat files, the file format matlab uses to save variables. The intention is to make it easier to for example plot results from matlab using the various plotting libraries available in typst instead of having to either mess with exporting figures from matlab and the pain of sizing them correctly or figuring out how to save data in different formats from matlab.

No code from matlab was used. The entire parsing code is written by me (and is available in the linked repository). Most of the .mat format is supported.


elegaanz:
Thank you!

PR #2548, opened 2025-06-15
griddle:0.1.0

micheledusi:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: _Griddle_ is a tool for creating and displaying crossword-style puzzles in Typst. The name comes from "grid" and "riddle". It reads an input file in the YAML format and produces a concise data structure with all the information for the crossword puzzle; then the user can print the definitions and the schema according to their preferences.


preview/griddle/0.1.0/src/loader.typ
```diff
@@ -0,0 +1,178 @@
...
+	return (
+		rows: rows_num, 
+		cols: cols_num, 
+		grid: grid,
+	)
+}
+
+// Loads a crossword from a YAML file and returns the grid structure.
+#let load-crossword(file, solved: false) = {
+	let data = yaml(file)
```
elegaanz: This call will be resolved relatively to this package, not to the user's project, so it's likely that the file will not be found. You should take the parsed data as input directly instead (`load-crossword(yaml("data.yml"))`), which also has the benefits of supporting other file formats or regular Typst variables.

micheledusi: Thank you so much for the fast feedback! I edited the code as requested.

elegaanz:
Thanks!

PR #2551, opened 2025-06-16
rubber-article:0.4.2

npikall:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- added a requested `abstract` function
- added a requested option to give the document some columns
- added a function, which can balance columns
- changed the default behavior of writing metadata to the pdf via the `maketitle` function 


elegaanz:
Thank you!

PR #2556, opened 2025-06-17
examination-paper:0.1.0

pdcxs:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a template to create an examination paper. It can auto-shuffle questions and choices according to a different seed. It supports choice questions, fill questions, true-false questions, and normal questions.


packages/preview/examination-paper/0.1.0/README.md
elegaanz, line 1: Can you write at least the heading and first paragraph in English, please? It is the language used by the global Typst community, and giving everyone a chance to understand what this package is about and whether it would fit their need is better. Having part of the README in other languages is fine.

Also, I think it should be a bit more complete, even if I don't understand what's written it feels very superficial to me.

packages/preview/examination-paper/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "examination-paper"
```
elegaanz, line 2: I think the name is a bit too descriptive given what your template aims at. Can you please change it to be a bit more creative and unique?

PR #2558, opened 2025-06-17
quill:0.7.1

Mc-Zen:
I am submitting
- [x] an update for a package

## Changelog

- Added the parameter `wires` to `quantum-circuit` that allows defining the number of (qu)bits explicitly. This parameter also accepts an array of wire counts, e.g., `wires: (1,) * qubits + (2,) * clbits` which is useful to avoid having to many `setwire` commands. 
- Added a `pass-through` parameter to `mqgate` that enables wires to pass through a multi-qubit gate instead of attaching themselves as in- and outputs. 
- Added the parameters `wire-count` and `stroke` to permute, allowing classical or even mixed permutation gates. 
- Added more flexibility to meters by allowing the content to be specified. 
- Fixed circuits with right-to-left text direction

### DIFF
--- a/packages/preview/quill/0.7.1/LICENSE
+++ b/packages/preview/quill/0.7.1/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023-2025 Mc-Zen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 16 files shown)


saecki:
Thanks for the update!

PR #2563, opened 2025-06-19
flyingcircus:3.3.0

Tetragramm:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed some errors that crept in with compiler changes and updated to the new Cetz.  Also a template flag to facilitate a common use-case for the template.


elegaanz:
Thank you.

PR #2567, opened 2025-06-19
elembic:1.1.0

PgBiel:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Pushed a critical fix for show rules on `e.selector`.
In addition, **outer labels on elements are now _optionally_ supported**, e.g. `#elem(...) <label-here>` instead of `#elem(..., label: <label-here>)`, by setting `labelable: true` on the element's configuration.

There are a few downsides to it, which is why the feature is disabled by default (set to `labelable: auto`):

- Show rules on `<label-here>` no longer provide the element's final fields;
- The element now requires `#show: e.prepare()`;
- The element cannot be inline anymore.

This is useful in some cases, such as for theorem / environment packages.


saecki:
Thanks, very cool!

PR #2569, opened 2025-06-20
sdu-touying-simpl:1.0.0

Dregen-Yor:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: A big update for a better templete.


packages/preview/sdu-touying-simpl/1.0.0/fonts/LXGWWenKai-Regular.ttf
saecki, line 1: Typst currently doesn't allow shipping fonts with packages. The rationale can be found here: https://github.com/typst/packages/blob/main/docs/resources.md#fonts-are-not-supported-in-packages

packages/preview/sdu-touying-simpl/1.0.0/fonts/SourceHanSerifCN-Regular.otf
saecki, line 1: Same here

packages/preview/sdu-touying-simpl/1.0.0/example/main.typ
```diff
@@ -0,0 +1,830 @@
+#import "@preview/sdu-touying-simpl:1.0.0" :*
+// #import "../lib.typ" :*
+// https://typst.app/universe/package/timeliney
+#import "@preview/timeliney:0.2.1"
```
saecki: This package seems to generate a warning because the outdated version of `cetz` uses the deprecated `path` function that will be removed in a future release. I've opened an issue on the repo: https://github.com/pta2002/typst-timeliney/issues/19

packages/preview/sdu-touying-simpl/1.0.0/example/main.typ
```diff
@@ -0,0 +1,830 @@
...
+
+#let TeX = {
+  set text(font: "New Computer Modern", weight: "regular")
+  box(width: 1.7em, {
+    [T]
+    place(top, dx: 0.56em, dy: 0.22em)[E]
+    place(top, dx: 1.1em)[X]
+  })
+}
+#import "@preview/octique:0.1.0": *
```
saecki: This package uses the deprecated `image.decode` function which will be removed in a future release, and break this package. There is an [open PR](https://github.com/0x6b/typst-octique/pull/2) on the repo, but it doesn't seem to be actively maintained.
Since it is quite small, you could maybe vendor a fork of the `octique` package in this repository until either the PR is merged, or someone else picks it up.

saecki (commented):
Thanks for this update! I've left a few comments about the included fonts and there are some outdated packages. Only two of them seem to be causing any warnings by the typst compiler, but it can't hurt to update the other dependencies too, if you have time.

packages/preview/sdu-touying-simpl/1.0.0/example/main.typ
Dregen-Yor: This package only used in example file, I have removed this part.

saecki: Could you just remove this section instead of commenting it out? This code will be copied directly to the user and will break with the next release, unless the package is updated. Once a new version of `timeliney` is released you can update the package and add back the section.

saecki:
Thanks!

PR #2582, opened 2025-06-24
elsevier:0.1.0

maucejo:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: The `elsevier` package is designed to closely resemble the `LaTeX` class used by Elsevier for article formatting, which is not publicly available. It is not intended for submission, but rather to help authors prepare articles that resemble the final published version. It mainly serves as a proof of concept, demonstrating that Typst is a viable option for academic writing and scientific publishing.


packages/preview/elsevier/0.1.0/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "elsevier"
```
saecki: We try to avoid too canonical package names, to avoid an unfair advantage when searching for packages. More on that [here](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).
To still make sure this package is discoverable you can use the `keywords` entry.

packages/preview/elsevier/0.1.0/docs/manual.pdf
saecki, line 1: In the readme you link to the `manual.pdf` inside your own repository. Could you either link to this file or remove it?

saecki (commented):
Thanks for this package! I've left a few comments regarding the name and the included `manual.pdf`.
The package check also shows an issue with the imported `els-globals.typ` file which doesn't seem to be present in this PR.

packages/preview/elsevier/0.1.0/docs/manual.pdf
maucejo: Sure I will remove `manual.pdf` from the PR and keep the link.

packages/preview/elsevier/0.1.0/typst.toml
maucejo: I will complete the toml to include `keywords`. I understand your concern about the name. To be honest, I don't have a better idea. I can suggest :
- `elsevier-journals`
- `elsevier-replica`
- `unofficial-elsevier`
- `elsevier-mockup`

Let me know what name seems preferable.

saecki: For the first one, I think there would need to be another qualifier. But any of the other ones is fine.

maucejo: I will `elsevire-replica` !

maucejo:
I close this PR because I changed the name of the repo. I resubmit the package. Thank you for your comments

PR #2604, opened 2025-06-26
grotesk-cv:1.0.5

AsiSkarp:
I am submitting
- [ ] a new package
- [x] an update for a package

- Improved metadata in compiled PDF
- Phone number entry is now optional
- Support for rectangle profile images

### DIFF
--- a/packages/preview/grotesk-cv/1.0.5/src/template/content/experience.typ
+++ b/packages/preview/grotesk-cv/1.0.5/src/template/content/experience.typ
@@ -1,6 +1,6 @@
 #let meta = toml("../info.toml")
 
-#import "@preview/grotesk-cv:1.0.4": experience-entry
+#import "@preview/grotesk-cv:1.0.5": experience-entry
 #import meta.import.fontawesome: *
 
 #let icon = meta.section.icon.experience
--- a/packages/preview/grotesk-cv/1.0.5/src/template/content/skills.typ
+++ b/packages/preview/grotesk-cv/1.0.5/src/template/content/skills.typ
@@ -1,6 +1,6 @@
 #let meta = toml("../info.toml")
 
-#import "@preview/grotesk-cv:1.0.4": skill-entry
+#import "@preview/grotesk-cv:1.0.5": skill-entry
 #import meta.import.fontawesome: *
 
 #let icon = meta.section.icon.skills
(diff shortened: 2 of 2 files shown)


packages/preview/grotesk-cv/1.0.5/docs/manual.pdf
saecki, line 1: This file will be very hard to discover unless someone is browsing this repository.
Could you either remove it, or make it easier to discover by linking to it in the readme?

packages/preview/grotesk-cv/1.0.5/README.md
```diff
@@ -0,0 +1,93 @@
+<h1 align="center"> grotesk-cv </h1>
+<div align="center">Version 1.0.4</div>
```
saecki: This version is out of date

saecki (commented):
Thanks for the update! I've left a few comments about the included `manual.pdf` file and there are some places which still the previous version of this package.

AsiSkarp:
I have updated the PR with fixes. The bot now complains about older versions - although the new version does not exist in the main project yet.

packages/preview/grotesk-cv/1.0.5/docs/manual.pdf
AsiSkarp: This has been removed

saecki:
Yes, but the template within this package should be using this version of the package. Once this PR is merged it will be available.

AsiSkarp:
@saecki I notice that you have put the `waiting-on-author` tag back on, but I have fixed the issues. Am I missing something?

packages/preview/grotesk-cv/1.0.5/src/template/content/other_experience.typ
```diff
@@ -0,0 +1,35 @@
+#let meta = toml("../info.toml")
+
+#import "@preview/grotesk-cv:1.0.4": experience-entry
```
saecki: Here

packages/preview/grotesk-cv/1.0.5/src/template/content/skills.typ
```diff
@@ -0,0 +1,157 @@
+#let meta = toml("../info.toml")
+
+#import "@preview/grotesk-cv:1.0.4": skill-entry
```
saecki: Here

packages/preview/grotesk-cv/1.0.5/src/template/content/experience.typ
```diff
@@ -0,0 +1,68 @@
+#let meta = toml("../info.toml")
+
+#import "@preview/grotesk-cv:1.0.4": experience-entry
```
saecki: Here

saecki (commented):
Yeah, the template should be using the same version of the package. The package checks shows warning for where it is still using the older `1.0.4` version.

packages/preview/grotesk-cv/1.0.5/src/template/content/other_experience.typ
AsiSkarp: Thank you. For the record, in a previous version I had introduced an entry in the toml file that I could use as a reference to the current version. That reference was then imported into all the relevant files, allowing me to update everything at once. I was told that that was not good practice, but I can't se the argument for removing that. Is explicit version importing a guideline for the project?

saecki: So if you're asking if you can import the package by using a relative part, that will just not work, since all template files will be copied when starting from a template and the `src/lib.typ` file won't be available.
Regarding the imports inside the `info.toml` file, I think you mean this comment: https://github.com/typst/packages/pull/2312#discussion_r2072990781?
My stance would be the same.

AsiSkarp: Ok. Thank you for clearing that up

saecki:
Just a heads up in case you forgot, there are still two places in the template which are using the `1.0.4` version.

AsiSkarp:
I have updated the last two file imports and the update should be ready now. I am unsure why the Linter step is now failing.

saecki:
Thanks! We're reworking the CI pipeline and the linting step is currently broken. I've run the check locally and everything passes.

PR #2584, opened 2025-06-24
latex-lookalike:0.1.0

mewmew:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Explain what the package does and why it's useful.

For users coming from LaTeX, there is an unmistakable look and feel of the documents that one has come to love over the years.

Re-creating this look and feel in Typst seems like a natural step, as it helps ease adaptation for both document producers and consumers.

The idea behind the `latex-lookalike` package is to provide Typst styles for different components and make them feel just like the output of LaTeX.

The styling of different components are kept separate, so users are free to pick and choose which styles to use (see the [Styles](https://github.com/mewmew/latex-lookalike?tab=readme-ov-file#styles) section of the README).

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name that isn't the most obvious or canonical name for what the package does
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] `exclude`d PDFs or README images, if any, but not the LICENSE

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #2605, opened 2025-06-26
latex-lookalike: fix import path of example to use @preview namespace

mewmew:
I am submitting
- [ ] a new package
- [x] an update for a package

This is just a small fix to use the `@preview` namespace instead of the `@local` namespace for the `latex-lookalike` example document.

Ref: https://github.com/typst/packages/pull/2584


saecki:
Good catch! Even though this is just a one line change, we try to avoid changing existing package files. Could you open a PR to release version `0.1.1` instead?

mewmew:
> Good catch! Even though this is just a one line change, we try to avoid changing existing package files. Could you open a PR to release version `0.1.1` instead?

Sure!

mewmew:
This PR is superseded by #2608.

PR #2608, opened 2025-06-27
latex-lookalike:0.1.1

mewmew:
I am submitting
- [ ] a new package
- [x] an update for a package

This is just a small fix to use the `@preview` namespace instead of the `@local` namespace for the latex-lookalike example document.

This PR supersedes #2605.


saecki:
Thanks!

PR #2610, opened 2025-06-28
lacy-ubc-math-project:0.2.0

lace-wing:
I am submitting
- [ ] a new package
- [x] an update for a package

## Update Note
This update greatly enhances the package's capability to allow
- dynamic question-solution tree;
- custom name display format;
- fine control over styles of each component;
- automatic question point assignment and solution targeting.

It also reliefs the package from a number of magical fields and string parsing.

Unfortunately, those changes are largely breaking.


packages/preview/lacy-ubc-math-project/0.2.0/README.md
```diff
@@ -0,0 +1,24 @@
+# Lacy-UBC Math Project Template
+This is a Typst template for UBC Math 100/101 group projects. It aims to reproduce most of the existing LaTex template while providing a more user-friendly interface for students who are not familiar with LaTex.
+
+For more information, please refer to the [documentation](/manual.pdf).
```
saecki: The relative link will also work when the file is not placed at the root of the repository, e.g. here in the package repository.
```suggestion
For more information, please refer to the [documentation](manual.pdf).
```

saecki (commented):
Thanks for the update! Just one little nitpick about the manual link.

Don't worry about the package check warnings, they are false positives.

PR #2612, opened 2025-06-29
wmglyph:0.1.0

kusaanko:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This plugin converts wmf(Windows Metafile) files to SVG.


packages/preview/wmglyph/0.1.0/lib.typ
```diff
@@ -0,0 +1,11 @@
+#let plugin = plugin("wmglyph.wasm")
+
+#let wmf-image(path, ..args) = {
+  let content = plugin.read_wmf(read(path, encoding: none))
```
elegaanz: I'm not sure this call to `read` (and the one in the other function below) will work as expected, as the paths will be resolved [relatively to the current file](https://github.com/typst/packages/blob/main/docs/resources.md#paths-are-package-scoped). You should take `bytes` directly as input, and let the user make the `read` call, even if this is a bit less convenient.

kusaanko: Thank you for your review. Fixed.

kusaanko:
I will cancel this PR because of inadequate implementation of wasm backend.

PR #2614, opened 2025-06-29
zero:0.4.0

Mc-Zen:
I am submitting
- [x] an update for a package

With this release, we introduce functions for formatting [units and quantities](https://github.com/Mc-Zen/zero?tab=readme-ov-file#units-and-quantities). Moreover, we add a new notion for setting up table alignment for seamless interoperability with other table Typst packages by making use of show rules. 

```typ
#figure(
  {
    show table: zero.format-table(none, auto, auto)
    table(..)
  },
  caption: []
)
```

## Changelog

- Adds the `zi` module for unit and quantity formatting. 
- Adds new way of applying table alignment via show-rules for seamless interoperability with other table packages. 
- Adds option to configure the group threshold individually for the integer and fractional part. 
- Fixes numbers in RTL direction context. 
- Fixes `figure.kind` detection of`ztable`. 
- Fixes direct usage of `nonum` in `ztable`. 
- Fixes uncertainties in combination with a `fixed` exponent.

PR #2615, opened 2025-06-29
pillar:0.3.3

Mc-Zen:
I am submitting
- [ ] a new package
- [x] an update for a package

This updates zero to 0.4.0 in the pillar packages. 

I will mark this PR ready for review when zero:0.4.0 is merged, see #2614.

### DIFF
--- a/packages/preview/pillar/0.3.3/LICENSE
+++ b/packages/preview/pillar/0.3.3/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024-2025 Mc-Zen
+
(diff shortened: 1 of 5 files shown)


saecki:
Thanks for this package! Seems like the same issue as with the `zero` package. In dark mode I only get the `alt` description because the images are missing.

Mc-Zen:
Why does it keep happening, I haven't even changed the links 😅

Mc-Zen:
Thanks for keeping such a sharp eye out. Next time I will test dark mode in a different browser :D

saecki:
Yeah, definitely seems weird. But now it works!

PR #2620, opened 2025-06-30
modern-ecnu-thesis:0.3.0

jtchen2k:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

In this update, we now support bilingual figure captions. We also fixed several typesetting details to ensure the template meets the official requirements.


packages/preview/modern-ecnu-thesis/0.3.0/0.3.0
elegaanz, line 1: Can you please delete this symlink, it creates a loop.

packages/preview/modern-ecnu-thesis/0.3.0/template/thesis.typ
```diff
@@ -0,0 +1,364 @@
+#import "@preview/modern-ecnu-thesis:0.3.0": documentclass, indent, no-indent, word-count-cjk, total-words
+#import "../utils/bilingual-figure.typ": *
```
elegaanz: If you want to use these functions in the template, you should re-export them in `lib.typ` and import them with the rest of the package above.

elegaanz:
Thank you.

PR #2622, opened 2025-06-30
muw-community-templates:0.1.0

felixbd:
I am submitting
- [x] a new package
- [ ] an update for a package

fix: corrected some spelling mistakes in my package

### DIFF
--- a/packages/preview/muw-community-templates/0.1.0/src/template/main.typ
+++ b/packages/preview/muw-community-templates/0.1.0/src/template/main.typ
@@ -2,6 +2,7 @@
 #import muw-presentation: *
 
 #set text(lang: "de")
+#polylux.enable-handout-mode(false)
 
 #let muw-logo-white(..args) = muw-box(fill: gray, figure(box([Hello], ..args)))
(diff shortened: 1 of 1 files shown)


felixbd:
just to clarify: this is just a small fix to the existing template (spelling mistakes fixed, information for use in the template added). The bot has unfortunately added the `new` label

saecki:
We are currently in the process of completely phasing out edits to already published packages, to ensure reproducibility. Could you instead create a new version of this package which includes these fixes?

felixbd:
@saecki  Since I have not changed anything in the template source (the behavior will not change), I think this is not necessary, especially since the template is only a day old. If you insist on a new version, I could update to `0.1.1`, but I think this would only clutter the repository.

saecki:
I know this seems a little silly, and in this case I could see an argument for just merging the changes. But we're trying to guarantee reproducibility, and finding a balance on a case to case basis is a lot more difficult to get right, than just having a clear rule that prohibits any changes. Additionally it's easy to miss something, that does in fact change behavior of a template or package.

felixbd:
yes ok, I can understand that

PR #2623, opened 2025-06-30
tabbyterms:0.1.0

bluss:
I am submitting
- [x] a new package
- [ ] an update for a package

This package allows laying out the term list as a regular table. By default, the terms form the first column and the descriptions are in the second column, but it can also be transposed. As an extension, additional columns can be added.

Aligning terms this way is a common modification of term list. It is also somewhat interesting as a different way to input certain tables (but it is in no way a replacement for the table function.)

The name is a combination of the `terms` element and :cat: (tabby cat) 

The README.md is generated from README.typ using `typlite` (a part of tinymist). It doesn't look great as plaintext, but it works as github markdown.


saecki:
Thanks for this package!

PR #2628, opened 2025-07-01
Link to the docs in the PR template

saecki:
This links to the relevant parts of the docs inside the PR template, so people don't have to search for (or guess) what we're referring to in the bullet points.


elegaanz:
Good idea, thank you

PR #2629, opened 2025-07-01
Add recommendation for linking files from external repositories

saecki:
Recommend linking to a specific tag or revision so the linked resources don't get out of sync with the package readme.


elegaanz:
Nice, thanks.

PR #2640, opened 2025-07-02
wenyuan-campaign:0.1.2

yanwenywan:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Adds fix by @dexterdy which means the outline respects your set page columns (outline used to always default to 2 column)


packages/preview/wenyuan-campaign/0.1.2/docs.pdf
saecki, line 1: This file will be very hard to discover, unless someone is browsing this repository directly. Could you either link to it in the readme directly instead of the one in your repository:
```md
[See the guide for more info](docs.pdf)
```
Or remove this file?

packages/preview/wenyuan-campaign/0.1.2/sample.pdf
saecki, line 1: Same here

saecki (commented):
Thanks for the update! I've left a couple of comments about the included `docs.pdf` and `sample.pdf` files.

packages/preview/wenyuan-campaign/0.1.2/docs.pdf
yanwenywan: I was under the impression that for documents we should use absolute links to our own repository rather than relative links inside this one, at least that was how it was before.

I'd actually prefer your suggestion, if the link would still work on the typst package overview page?

saecki: Linking to your repository is also ok, if so it would be even better to link to a specific tag or revision of the repository instead of just the main branch.
Linking to a relative path does also work on typst universe. So it's up to you :)
We just try to avoid having PDFs that aren't actively used in the documentation, because they contribute significantly to the size of this repository.

yanwenywan: Good to know, thanks! I'll link both these documents directly with the relative path then

saecki:
Thanks!

PR #2646, opened 2025-07-03
nerd-icons:0.1.0

TuTiuTe:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package is similar to [fontawesome](https://typst.app/universe/package/fontawesome), but works with the nerd fonts, which encapsulate a bigger set of icons.
When making a CV, I needed some niche programming language icons, thus I made this library.
Although with 10 000+ icons it doesn't only bring niche icons, some that can useful in a lot of contexts


saecki:
Thanks for the package! The package-check shows warnings because the names of the symbols aren't snake-case. I can see an argument for why the names should stay exactly the same, but making them all snake case could be a little more idiomatic. I'm fine with either way, just checking in with you :)

TuTiuTe:
Yup, saw all those warnings too
I'd prefer keeping them this way so the user can copy and paste the class name from the [cheat sheet](https://www.nerdfonts.com/cheat-sheet), without having to modify them

PR #2647, opened 2025-07-03
diagraph:0.3.5

Robotechnic:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added edges personalization in the `adjacency` function
- Updated graphviz version to 13.1.0

### DIFF
--- a/packages/preview/diagraph/0.3.5/src/html_processing.typ
+++ b/packages/preview/diagraph/0.3.5/src/html_processing.typ
@@ -1,0 +1,68 @@
+#let plugin = plugin("../graphviz_interface/diagraph.wasm")
+
+
+#let lower-string(text) = {
+	for c in text {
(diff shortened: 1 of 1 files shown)


packages/preview/diagraph/0.3.5/src/internals.typ
```diff
@@ -0,0 +1,627 @@
+#import "../graphviz_interface/protocol.typ": *
+#import "html_processing.typ": html-to-content
```
saecki, line 2: Seems like this file is missing

saecki (commented):
Thanks for this upate! There is an error reported by the package check.

packages/preview/diagraph/0.3.5/src/internals.typ
Robotechnic: I forgot to add it to my makefile. It's fixed

PR #2653, opened 2025-07-04
minimal-thesis-luebeck:0.5.0

fhemstra:
I am submitting
- [ ] a new package
- [x] an update for a package

This update is about some styling: Headings, figure numbers, headers, fonts.


saecki:
Thanks!

PR #2656, opened 2025-07-04
nerd-icons:0.2.0

TuTiuTe:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
This update adds a QoL feature: Prefixes are no longer required. Now `#nf-icon("dog")` will get expanded to `#nf-icon("nf-*-dog")` where the * is a valid prefix until it finds a working icon
Your favorite prefix can be set with `#set-favorite-nf-prefix` so it resolves to that one in case of duplicates  
New `#nf-icon-string`function which returns the utf-16 character without formatting  
Fixed some typos


elegaanz:
Thank you.

PR #2659, opened 2025-07-05
ezexam:0.1.1

gbchu:
I am submitting

- [x] an update for a package

Fix the issue where the width of an image cannot be retrieved when the 'choices' option is set to an image with a percentage-based width.


saecki:
Thanks!

PR #2662, opened 2025-07-07
elembic:1.1.1

PgBiel:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

Bug fixes:
- Synthesized fields are no longer removed when using `cond-set`, instead being computed again (GH#60)
- Required fields are no longer foldable to avoid bugs in some edge cases (GH#61)
   - This might be revised in the future for custom types.
- Fix union error message generation panic when using types with `check`, such as `literal(...)` or `dict(T)` (GH#63)


elegaanz:
Thank you!

PR #2680, opened 2025-07-09
algorithmic:1.0.1

quachpas:


packages/preview/algorithmic/1.0.1/typst.toml
```diff
@@ -0,0 +1,31 @@
...
+description = "Algorithm pseudocode typesetting for Typst, inspired by algorithmicx in LaTeX"
+repository = "https://github.com/typst-community/typst-algorithmic"
+keywords = ["algorithm", "pseudocode", "procedure"]
+categories = ["components", "visualization"]
+disciplines = ["computer-science", "mathematics"]
+compiler = "0.13.0"
+exclude = [
+  "docs/*.typ",
+  "docs/*.pdf",
+  "tests/**/out",
```
saecki, line 16: I think the whole `tests` directory could be excluded here, since the images are only used in the readme right?

quachpas: Hi, thx
This is deliberate, since the refs are used in the README. Without, it wouldn't render properly on universe

saecki: The `exclude` key only defines what is downloaded by the user. Everything that is in this repository will be available on typst-universe. See here: https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

quachpas: Ah! I can re-release a patch version then, since I'd prefer not touch the existing release.
I use the exclude key in my packaging script to make the package tarball.
It doesn't make sense to me not to include the README + dependencies in the package?

In the tips document there seems to be a typo as it says to include "the first two groups", then exclude the second group 😂

saecki: > Ah! I can re-release a patch version then, since I'd prefer not touch the existing release.
> I use the exclude key in my packaging script to make the package tarball.

You don't have to change the release tar ball in your own repository. You can just edit this PR to exclude those files. This won't break the usage of the package in any way I could think of, since paths are package/project scoped and this package doesn't access the files in the `tests` directory.

>  It doesn't make sense to me not to include the README + dependencies in the package?

READMEs are typically still included in the shipped packages:
> and the README (which is generally a lightweight file, and can provide minimal documentation in case the user is offline or can't access anything else than their local package cache for some other reason).

> In the tips document there seems to be a typo as it says to include "the first two groups", then exclude the second group 😂

I've just checked and the wording seems correct. I read it like this: the first and the second group of files should be committed to this repository. And the second group should be added to the `exclude` list. This will enable typst universe to display all relevant documentation linked in the readme, but will avoid shipping any resources not used by the package.

quachpas: Thanks

> This will enable typst universe to display all relevant documentation linked in the readme

In which case, did I fail to mention that the tests references are used in the README? Would that be fine with you?

saecki: This is to be expected, the included readme serves as a best effort fallback. It's a rare case that someone will manually browse the local package cache, in search for documentation. Have a look at [`cetz`](https://github.com/typst/packages/tree/main/packages/preview/cetz/0.3.4) for example. It's pretty much the same case, they have a `gallery` directory with a bunch of images used in the readme, which is `exclude`d in their `typst.toml`.

quachpas: > You don't have to change the release tar ball in your own repository.

My personal preference is not to diverge from the main project release. The package that I submit here should be identical to the one in the main project.

In any case, I'll send a new PR. Thanks for he help!

PR #2699, opened 2025-07-10
fibber:0.1.0

gauthamrmn:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Fibber makes it easy to create microfabrication process diagrams that are not only easy to write the first time but also easy to make edits to. This would be useful for people who need to work closely with and communicate with others about the microfabrication process, such as microelectronics researchers.


packages/preview/fibber/0.1.0/README.md
```diff
@@ -0,0 +1,185 @@
...
+- [Formatting Outputs](#formatting-outputs)
+- [Formatting Quickstart](#formatting-quickstart)
+
+## Usage
+
+Import like so
+```
+#import "@preview/fibber:0.1.0"
+```
+See examples in the examples directory.
```
elegaanz: I would link to this directory here (or even better, individual examples, with a link for the source and the produced PDF), otherwise it's hard to discover where that directory is when viewing the README from Typst Universe.

gauthamrmn: I linked to the github repo examples directory directly (so when you click on it you see source files and corresponding pdfs on github) so that I can push some more examples without pushing a new version of the package.

elegaanz: Good idea. Then I think it would make sense to delete the examples/ directory in this copy of the package entirely, since it is excluded anyways and can't really be accessed unless you browse this repository by hand (which nobody ever does, I think), meaning it only adds its weight to an already too large repository.

gauthamrmn: That's true. I've removed it now.

elegaanz:
Thank you :)

PR #2704, opened 2025-07-12
touying-simpl-hkustgz:0.1.2

exAClior:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update to support `touying:0.6.*`


saecki:
Thanks!

PR #2712, opened 2025-07-13
athena-tu-darmstadt-exercise:0.2.0

FussballAndy:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Made the lower box in the title more customisable, added a gray info box and allowed student numbers to be added to authors.


FussballAndy:
Hi, one small question regarding the licensing: in the normal use the user is should download a licensed file (the logo of the university) and put that into the logos directory. Thus technically the template directory is not free to distribute in the normal use - unless the user explicitly ignores the file. Similarly the fonts could also be a problem although those can easily be installed elsewhere. So is this licensing of a "normal use" file a problem or does it only apply to files directly shipped with the template?

elegaanz:
It is the user's responsibility, not yours, to respect the license of the file they download themselves. You should only take care of it for files you directly ship in your package/template.

elegaanz:
Thank you.

PR #2721, opened 2025-07-15
elspub:0.2.0

maucejo:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: `elspub` is an update of the template formerly named `elsevier-replica` to avoid conflicts with Elsevier. In addition to the name, the template provides the following improvements:
- Add a new journal dictionary for Elsevier journals (Journal of Fluids and Structures)
- Add field `font` in journal dictionaries to specify the font used in the journal. This field a dictionary with the following keys:
  - `text`: Font used for the text
  - `math`: Font used for the math
  - `homepage`: Font used for the homepage

### DIFF
--- a/packages/preview/elspub/0.2.0/src/els-journal.typ
+++ b/packages/preview/elspub/0.2.0/src/els-journal.typ
@@ -8,7 +8,7 @@
   font: (
     text: "New Computer Modern",
     math: "New Computer Modern Math",
-    homepage: "Roboto",
+    homepage: ("Roboto", "New Computer Modern"),
   ),
   numcol: 1,
   foot-info: [Published by Elsevier Inc.]


elegaanz:
Thank you for this update. Regarding the errors reported by CI, it would be nice to fix the equations labeling to prevent compilation errors. For the math font error, it would be nice to add a math font that ships with Typst as a fallback (I think `New Computer Modern Math` is always available), so that the template still compiles even if the font you are supposed to used is not available.

maucejo:
> Thank you for this update. Regarding the errors reported by CI, it would be nice to fix the equations labeling to prevent compilation errors. For the math font error, it would be nice to add a math font that ships with Typst as a fallback (I think `New Computer Modern Math` is always available), so that the template still compiles even if the font you are supposed to used is not available.

Thank you for the report. I have modify the font specs and test my template locally using `typst-package-check`. I think that the problem reported with equate is related to the font. I hope that it will solve the compilation errors.

elegaanz:
Great, thank you!

PR #2727, opened 2025-07-16
hamnosys-includer:1.0.0

TestTimothy:
I am submitting
- [ ] a new package
- [x] an update for a package

Add a new `#ham` variable which is easier to work with. Update documentation.


packages/preview/hamnosys-includer/1.0.0/README.md
```diff
@@ -0,0 +1,51 @@
...
+HamNoSys is a system for the phonetic transcription of signed languages. The HamNoSysUnicode font uses the Unicode Private Use Area for the HamNoSys symbols.
+
+The font (and the symbol names) are taken from the TeX package [_hamnosys_](https://ctan.org/pkg/hamnosys?lang=en), maintained by Thomas Hanke & Marc Schulder. You will first need to download the TeX package and find the HamNoSysUnicode font.
+
+If you're working locally, install the font (or pass its location to Typst with the `--font-path` argument). If you're using the [Typst web service](https://typst.app), you just need to include the font among your uploaded files. Without the font, this package is useless.
+
+Second, include the line
+
+```Typst
+#import "hamnosys.typ": ham, hamnosys, hamnosys-text
```
elegaanz: This is not the recommended way to install packages, it is better to suggest to use Typst Universe.

```suggestion
#import "@preview/hamnosys-includer:1.0.0": ham, hamnosys, hamnosys-text
```

packages/preview/hamnosys-includer/1.0.0/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+in your document.
+
+## Direct Entry
+
+You can enter the symbols directly into your Typst document. There's no way to actually type most of them, as they're mostly in the Unicode Private Use Area, but this is handy if you want to copy and paste from elsewhere. If you do this, you should wrap them in the `#hamnosys-text` function.
+
+(I cannot demonstrate this here, as there's no way to force rendering with a specific font here. View `docs/README.pdf` in the GitHub project for a clear example.)
```
elegaanz: I would recommend linking to the PDF file here, otherwise it can be a bit hard to find.

packages/preview/hamnosys-includer/1.0.0/README.md
```diff
@@ -0,0 +1,51 @@
...
+#import "@preview/hamnosys-includer:1.0.0": ham, hamnosys, hamnosys-text
+```
+
+in your document.
+
+## Direct Entry
+
+You can enter the symbols directly into your Typst document. There's no way to actually type most of them, as they're mostly in the Unicode Private Use Area, but this is handy if you want to copy and paste from elsewhere. If you do this, you should wrap them in the `#hamnosys-text` function.
+
+This is demonstrated in [`README.pdf`](https://github.com/TestTimothy/Typst-HamNoSys/blob/main/docs/README.pdf).
```
elegaanz, line 21: Sorry, I'm a bit nitpicking here, but if you link to your own repository, can you:

1. link to a specific commit or tag
2. remove the `docs/` folder from this copy of the project, to not bloat the already too large package repository.

Thank you.

elegaanz:
Hey, I just wanted to check with you if you will have time to fix the last issue in the README so that we can merge this PR? No worries if you can't for now, I'll just close the PR then, but you'll be able to open a new one once you're ready.

PR #2730, opened 2025-07-16
pc-letter:0.2.0

thatfloflo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

**Added:**
- Author's details (`author` argument of `pc-letter.init`) now also accepts a `web` argument where an author can include their homepage address.

**Changed:**
- The default `medium` (`style` argument of `pc-letter.init`) is now `"print"` (previouslt `"digital"`), because a letter set for printing will work digitally without problems, but one set for digital-only use (with background fill and such) may well produce problems when printing.
- Author's address (`author.address` argument of `pc-letter.init`) is now optional.
- Author's contact details (phone, email, web) (`author.phone`, `author.email`, `author.web` arguments of `pc-letter.init`) are now all optional and break across two lines iff more than two of them are specified.
- Headings (used for subject lines and as section separators) now get 1.25em spacing above and below (previously 1em), making them subtly more prominent as separators in the text body.
- Internal calculations for spacing of different fields is now relative to page dimensions and margins, which will help with implementing different paper sizes (e.g. `us-letter`) in future versions of `pc-letter`.
- The package now has full [`tidy`](https://typst.app/universe/package/tidy/) 0.4.0-compatible doc comments. While `tinymist` still only parses the pre-0.4.0 style comments, it is expected that this will also enhance hinting for users once `tinymist` implements the up-to-date doc comment style. The dictionary-passing method of initialisation may still pose a problem for this, but the change should generally aid in future-proofing the package.

**Fixed:**
- Missing translation of pagination for non-Austrian German fixed.


saecki:
Thanks!

PR #2737, opened 2025-07-17
typsium:0.3.0

tryptophawa:
I am submitting
- [x] an update for a package

Description: ~Explain what the package does and why it's useful.~

We have successfully added support for content in `#ce[]` blocks. Also, more symbols and oxidation state can be written now.


tryptophawa:
👍🏼 First time seeing all green check. 🐱

elegaanz:
Sorry, it was just that the CI was under maintenance. I restarted it and it reports a few warnings about elembic being updated (but because the new version is only a patch release I think it is fine to ignore it).

PR #2752, opened 2025-07-19
algorithmic:1.0.3

quachpas:
.

### DIFF
--- a/packages/preview/algorithmic/1.0.3/README.md
+++ b/packages/preview/algorithmic/1.0.3/README.md
@@ -1,0 +1,438 @@
+<!--
+SPDX-FileCopyrightText: 2023 Jade Lovelace
+SPDX-FileCopyrightText: 2025 Pascal Quach
+SPDX-FileCopyrightText: 2025 Typst Community
+SPDX-FileCopyrightText: 2025 Contributors to the typst-algorithmic project
+SPDX-License-Identifier: MIT
+-->
+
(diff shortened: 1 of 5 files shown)


packages/preview/algorithmic/1.0.3/README.md
```diff
@@ -0,0 +1,438 @@
...
+    algorithm(inset: inset, ..bits),
+  )
+}
+```
+
+In order to use the `algorithm-figure` function, you need to style the figure
+with the `style-algorithm` show rule.
+
+```typst
+#import algorithmic: algorithm-figure, style-algorithm
```
saecki, line 119:
```suggestion
#import "@preview/algorithmic:1.0.3": algorithm-figure, style-algorithm
```

saecki (commented):
Thanks for this update!
We've reworked the CI and there are still some minor issues, everything should work now :)

I've left one minor suggestion about a readme example.

packages/preview/algorithmic/1.0.3/README.md
quachpas: Hello @saecki, thanks for the review! For that one, it's more or less the same as the first example, so the package import is not repeated, and we assume the namespace already exists.
~~I did notice I was not updating the version number in the README though, so I will do that from now on.~~ I was looking at something else, sorry!

quachpas:
Thank you @saecki !

PR #2758, opened 2025-07-19
kthesis:0.1.2

RafDevX:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Add support for theses with no subtitle, and update package dependency versions.


elegaanz:
Thank you!

PR #2765, opened 2025-07-21
wut-thesis:0.1.1

fuine:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I've updated the template to basically force the user to install an appropriate font, which is used for styling the faculty header in the title page. This means that the template does not compile by default on the web app - user is prompted to install the font via assertion error. In the previous version it only generated warnings about unknown font and fell back on some other font, which resulted in a wrong, unacceptable header.  Template can still be compiled without the font if the user sets `titlepage-info: none`, which is explained in the README. 

One way to make the template compile fine without the need for the font would be to create `.svg` headers for all faculties and thesis types, but this would result in 48 `.svg` files ((20 faculties + 3 thesis types + university name) * two language versions), totaling ~380KB. I don't know which approach is better from the Typst's team point of view, I would appreciate your input on the matter.


elegaanz:
I think a more appropriate way would be to detect whether the font is available or not (judging by the metrics of the some rendered text) and display a very prominent warning in the document itself if not. I know some other packages use this approach, and they have the benefit of providing the user with a compiling document, while making sure they can't submit it without fixing the issue.

fuine:
Ok, thanks, I assume then that you'd prefer a template without so many svg files? I will render the warnings in the titlepage and get back to you

fuine:
I've changed the assertions into rendered warning blocks and tweaked the header rendering, so I think it should be ready for merge now

elegaanz:
Thank you!

PR #2771, opened 2025-07-21
rubber-article:0.5.0

npikall:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- new option to set default font
- new option to set spacings around the title-block
- new option to set paragraph spacing
- new option to set the paper size
- new function to have a common vertical spacer between paragraphs
- patch of justify option
- refactor of the argument names
- better documentation


saecki:
Thanks!

PR #2777, opened 2025-07-22
clean-hwr:0.1.1

Testspieler09:
I am submitting
- [ ] a new package
- [x] an update for a package

# Update Fixes / Features
- automatic date format when selecting german (feat)
- logo not displaying because of relative path (fix)


elegaanz:
Thank you.

PR #2790, opened 2025-07-25
clean-math-paper:0.2.3

JoshuaLampert:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This update fixes a bug, where the color of the authors with ORCID was hardcoded to be black also if other text colors were used.


elegaanz:
Thank you.

PR #2799, opened 2025-07-27
tablem:0.3.0

OrangeX4:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- **(Breaking Change)** Removed support for using empty spaces (` `) as cell merging markers.
- **(Breaking Change)** Removed `ignore-second-row` and `use-table-header` parameters.
- Table header detection and column alignment are now automatic:  
  - If the second row matches `---`, `:---`, or `:---:`, the first row is treated as a header row.
  - Alignment is determined by the syntax:
    - `| --- |` for **auto** (default)
    - `| :--- |` for **left-aligned**
    - `| :---: |` for **center-aligned**
    - `| ---: |` for **right-aligned**


elegaanz:
Thank you.

PR #2803, opened 2025-07-28
fontawesome:0.6.0

duskmoon314:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

Font Awesome released v7 on 2025-07-22. We now support both versions 7 and 6 in v0.6.0.

- The icon map is separated into three: common, v6, and v7.
  - Icons with the identical Unicode or that only appear once (e.g., v7 only) belong to the common
  - Icons with different Unicode codes are separated into a version-specific map
- Generate helper functions with a version suffix to distinguish conflicting Unicode
  - e.g., `fa-user-large-6` and `fa-user-large-7` (with `fa-user-large` as the latter's alias)
- Add state `_fa_version` to help switching between v7 and v6.
  - `fa-version("6")` helps set the default font list to v6 and use common + v6 as the default icon map
- Format typst codes
- Refactor Python helper script


saecki:
Thanks for the update!

PR #2806, opened 2025-07-28
olaii-upn-qr:0.0.1

Klemen1337:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package helps rendering UPN QR (Univerzalni plačilni nalog QR) payment slip in Slovenia.


packages/preview/upn-qr/0.0.1/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "upn-qr"
```
saecki: This name seems very canonical. We try to avoid too generic names, so multiple packages with for the same purpose can co-exist without one having an unfair advantage. See our [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

Klemen1337: Would the rename to 'olaii-upn-qr' or 'upn-qr-olaii' be ok?

The debug fields are a little off but when printed they fit better. I think its some sort of printer scaling or rasterization issue. I tested it on "HP LaserJet 400 M401dn" laser printer with scaling on 100%.

saecki: If I understand correctly Olaii is the name of your company right? That would require us to verify this is official. To do that the companies founder should send an email to hello@typst.app from a company email approving this package.
Otherwise you could just pick another name that isn't associated with any entity.

Klemen1337: Alright, I've sent the email :)

saecki: Thanks, we've received the email. I would suggest using `olaii-upn-qr` because that's more in line with the naming scheme of university packages.

Klemen1337:
❤️

PR #2815, opened 2025-07-30
bookletic:0.3.2

harrellbm:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Maintenance release. Tested to be compatible with Typst 0.13.1 and minor documentation fixes.

- [ ] added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys
- [ ] added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package
- [ ] have chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) and added a `LICENSE` file or linked one in my `README.md`
- [ ] tested my package locally on my system and it worked

### DIFF
--- a/packages/preview/bookletic/0.3.2/typst.toml
+++ b/packages/preview/bookletic/0.3.2/typst.toml
@@ -1,7 +1,7 @@
 [package]
 name = "bookletic"
 version = "0.3.2"
-compiler = "0.14.0"
+compiler = "0.11.0"
 repository = "https://github.com/harrellbm/Bookletic.git"
 entrypoint = "src/lib.typ"
 authors = [


packages/preview/bookletic/0.3.1/example/example.pdf
saecki, line 1: This file will be very hard to discover, unless someone is browsing this repository directly. Could you either remove it or link to it in the readme?

packages/preview/bookletic/0.3.1/example/example.typ
```diff
@@ -0,0 +1,244 @@
+#import "@preview/bookletic:0.3.0" // To use through preview
```
saecki: This should use the newest version.

saecki (commented):
Thanks for the update! I've left a few comments regarding the examples.

harrellbm:
Let me know if there are any other outstanding issues that need fixed before merging!

saecki:
Thanks, this fixes most things, and the failing package check is a false positive. But the `example.pdf` file is still not linked in the readme. Could you just remove it, since it can easily be reproduced by compiling the example?

harrellbm:
@saecki So sorry for the delay. Finally had some time and removed the example files. I also tested 0.14.0 and things worked as expected.

saecki:
No problem, the `compiler` field is a minimum required version to used this package though. Setting it to `0.14` means that it won't compile using `0.13`. If that was your intention that's fine, otherwise it might be better to leave it at `0.13` while there is only a release candidate for `0.14`

harrellbm:
@saecki  That is my bad! It compiles just fine all the way down to 0.11 so I will change the toml reflect that.

PR #2818, opened 2025-07-31
isc-hei-bthesis:0.6.0, isc-hei-exec-summary:0.6.0 and isc-hei-report:0.6.0

pmudry:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

- Added template version in document template metadata 
- Panic message if no URL for repos
- Panic message if no document ID
- Typos fixed
- Header modified for right and left pages 
- Header modifier for executive summary, only space for 2 lines. 
- Subtitles for long titles
- README.md corrected with correct refs, image of templates corrected as well for Typst universe


elegaanz:
I think it was already the case in previous versions, but I'm a bit confused by your READMEs, which as far as I can tell are the same for all three packages. I feel like they should be more specific to a single package (while they can mention the others and explain how they relate, of course), especially for the "Project initialization and compilation" section. What do you think?

pmudry:
You are right. I refactored the READMEs in 60767c0 so that each template is clearer. Better like this!

elegaanz:
Great, thank you!

PR #2821, opened 2025-07-31
clean-hwr:0.1.2

Testspieler09:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- Added a copyright notice for the used university logo
- Added the `paper_type` field to the metadata field to enable the user to specify if they are writing a bachelorthesis or something else


elegaanz:
Thanks!

PR #2829, opened 2025-08-01
clean-hwr:0.1.3

Testspieler09:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- fixed the line spacing of the default `acrostiche` configuration


elegaanz:
Thanks.

PR #2831, opened 2025-08-01
retrofit:0.1.1

xkevio:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog

- fix: use relative page counter for collecting pages
- fix: deduplicate pages before passing them into format function
- fix: use `bib.style` instead of `style` when reading in CSL files


elegaanz:
Thank you!

PR #2840, opened 2025-08-04
asciim:0.1.0

waterlens:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: displaying ASCII tables with a 16x16 grid and overlays on elements.

An example of the generated diagram


packages/preview/ascii-matrix/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "ascii-matrix"
```
saecki: The name seems a little generic. We try to avoid too canonical names, so packages serving the same purpose can coexist without one having  an unfair advantage. See our [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

packages/preview/ascii-matrix/0.1.0/example.typ
saecki, line 1: Currently the included example files will be very hard to discover, unless someone is browsing this repository directly. Could you either remove them or link them in the readme?

packages/preview/ascii-matrix/0.1.0/typst.toml
waterlens: Thanks for the suggestion. How about "asciim"? By the way, I noticed there's also a package called "ascii-ipa." Is it allowed since it was submitted before the naming rules took effect?

packages/preview/ascii-matrix/0.1.0/example.typ
waterlens: Sure

packages/preview/ascii-matrix/0.1.0/typst.toml
saecki: Yeah, some packages have names that probably wouldn't pass review today. And sometimes package names just slip past review. Nonetheless we try to enforce the naming rules as best we can.

PR #2856, opened 2025-08-06
rowmantic:0.4.0

bluss:
I am submitting
- [ ] a new package
- [x] an update for a package

Define tables row by row with inline separators between cells.

- New `row` function, which allows transforming row elements
- One minor bugfix for treating `[ ]` as a rows
- new package category visualization


saecki:
Thanks for the update!

PR #2858, opened 2025-08-07
cyberschool-errorteaplate:0.1.10

ErrorTeaPot:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I have patched an unwanted behavior pushed in the latest update that were hiding references to other things than headings. I have also added a variable in the header to turn on or off the outline.


ErrorTeaPot:
I have removed all ref changes because it was breaking stuff. It is basically a  revert to version 0.1.8.

saecki:
Thanks for the update!

PR #2866, opened 2025-08-09
hsrm-internship-report:0.0.1

leonardhwg:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A template for internship reports that highlights who wrote what. Wrap text with #author("name")[…] to color‑code passages per author, with an optional legend mapping colors to authors on outline page. 


packages/preview/hsrm-internship-report/0.0.1/author-marker.typ
```diff
@@ -0,0 +1,13 @@
+#import "template/data.typ": authors
```
elegaanz, line 1: This won't work: files in `template/` will get copied when a user creates a new project from your template, but this import will refer to the original version, so even if the user modifies their copy, the changes won't be reflected in the original file.

elegaanz: What you could do instead is take the `authors` array as a parameter of the `author` function. If you want to keep the short syntax you currently have, you can encourage people to partially call the function using `.with`:

```typst
#let RZ = author.with(authors, "RZ")

#RZ[R. Z. wrote this.]
```

You could also return an array of author functions from the `author` function:

```typst
#let (RZ, SG, DU) = author(authors)

#RZ[
  = Introduction
  #lorem(100)
]

#DU[
  = State of the art
  #lorem(140)
]
```

packages/preview/hsrm-internship-report/0.0.1/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "hsrm-internship-report"
```
elegaanz, line 2: Because this is not an official template, another name would be more appropriate. [We have specific naming guidelines](https://github.com/typst/packages/blob/0488b38bb7fdfbce81981803ebbcf5612f8919f8/docs/manifest.md#naming-rules).

packages/preview/hsrm-internship-report/0.0.1/author-marker.typ
leonardhwg: Alright, thank you for your feedback! I'll close the PR for now and open it again as soon as I fixed it.

PR #2870, opened 2025-08-10
ugent-style:0.1.0

th1j5:
I am submitting
- [x] a new package

Description: An unofficial template to bring the UGent style to Typst.
For students and staff alike.
"Write documents, presentations and dissertations in the same, uniform UGent style. Unofficial template."
The UGent is Ghent University in Belgium.

  - => the guidelines say to use `<something>-ugent`, I use `ugent-<something>`, which I hope is ok.


packages/preview/ugent-style/0.1.0/typst.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "ugent-style"
```
saecki: As you mentioned the [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) suggest using a non-descriptive prefix instead of a suffix, which is preferred. 
`style` is a little too descriptive to function as that non-descriptive part, so this name would need another qualifier to make it less canonical. You could also keep the `style` suffix (e.g. `organized-ugent-style`)

th1j5: We're brainstorming here :)
Would `SOS-ugent` be allowed? or is there a preference for lowercase `sos-ugent`?

saecki: Yeah that would be fine, but package names are required to be lower kebab case, so it would have to be: `sos-ugent`.

th1j5:
ok, struggled a bit with git sparse-checkout, but the package is renamed

saecki:
Thanks!

PR #2873, opened 2025-08-12
hallon:0.1.0

mewmew:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: The hallon package provides a simple way to add subfigures and named references.

The handling of subfigures was inspired by the example of @RaulDurand in https://github.com/typst/typst/issues/246#issuecomment-2953421350 and the https://typst.app/universe/package/headcount package by @jbirnick..

I have read and followed the submission guidelines and, in particular, I
- [x] selected [a name](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) that isn't the most obvious or canonical name for what the package does
- [x] added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys
- [x] added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package
- [x] have chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] [`exclude`d](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude) PDFs or README images, if any, but not the LICENSE

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #2878, opened 2025-08-12
smartaref:0.1.0

mewmew:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Provides a "smart" way to handle and display consecutive references (i.e. `cleveref` in Typst).


mewmew:
**Edit:** this PR depends on https://github.com/typst/packages/pull/2873 since `smartaref` uses the `@preview/hallon:0.1.0` package.

YDX-2147483647:
Hi! You mentioned the remote context problem in https://github.com/typst/typst/pull/6757#issuecomment-3192020572.
The following forum post provides two solutions:
- Replace any `counter("…")` with `counter(figure.where(kind: "…"))`, eliminating the need of `show ref`. (This may not be applicable here.)
- Store it in `metadata`. https://github.com/jbirnick/typst-great-theorems/issues/4#issuecomment-3082272389

[How to `counter.display-at(location)` or get remote context? - Questions - Typst Forum](https://forum.typst.app/t/how-to-counter-display-at-location-or-get-remote-context/5096/5)

mewmew:
> **Edit:** this PR depends on #2873 since `smartaref` uses the `@preview/hallon:0.1.0` package.

I removed the dependency on `@preview/hallon` from the `smartaref` package, as it was only using a single function `title-case` (which is now included in `smartaref`).

There are examples for different packages using subfigures (both `subpar` and `hallon`).

Since the library has no external dependencies and the example files are only there to showcase different use cases, this PR don't depend on #2873 anymore.

Also, I've updated the `cref` function to support compacting of consecutive references!

E.g. "figures 1, 2, 3 and 4" becomes "figures 1 to 4".

See https://github.com/typst/packages/pull/2878/files#diff-cd459423910fa5d46b5a7b2e0257609878894facd374edc69224660f8eb7d275 for an example.

compacting references

mewmew:
Note, the warning reported by the linter seems to be incorrectly reported.

From https://github.com/typst/packages/pull/2878/checks?check_run_id=48515314919
> This value seems to be public. It is recommended to use kebab-case names.

https://github.com/typst/packages/pull/2878/files#annotation_38040071633
```typst
	supplement: supplement-func.with(capital: true),
```

As far as I can tell, it is using `kebab-case` (parhaps the `.with(...)` is what confuses the linter?).

elegaanz:
The warning applies to the line above, it is about the `Cref` function. But I guess it makes sense to have named this way, so it's fine to ignore the warning here.

mewmew:
> The warning applies to the line above, it is about the `Cref` function. But I guess it makes sense to have named this way, so it's fine to ignore the warning here.

It was named as such to be analogous to the `cleveref` package of LaTeX, where `cref` gives a lower-case reference supplement, and `Cref` gives an upper-case reference supplement.

I wanted to use the same defaults for Typst, for transfer learning.

PR #2881, opened 2025-08-12
clean-hwr:0.1.4

Testspieler09:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- Addressed that glossary entries were not working in headings.
- further glossary options
- Custom signature labels
- confidentiality notice field

Just out of curiosity, is there a better way of dealing with an empty glossary given by the user, when using the `glossarium` package?
```typst
register-glossary(if glossary.entries != () {glossary.entries} else {((key:"KPI", short: "KPI", long: "Key Performance Indicator"),)})
```

### DIFF
--- a/packages/preview/clean-hwr/0.1.4/README.md
+++ b/packages/preview/clean-hwr/0.1.4/README.md
@@ -40,8 +40,18 @@ The `hwr(...)` function is the main entry point to configure and render the **PT
     (key: "GitHub", value: "yourhandle", index: 0),
   ),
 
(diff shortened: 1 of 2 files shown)


Testspieler09:
I will add the confidentiality notice (Sperrvermerk) as an optional field to this version.

elegaanz:
I don't know enough about the glossarium API to answer your question, but if the way you did it works it doesn't seem too bad to me. I'll merge it as it is for now, you can update the call to glossarium in a future release if you find a nicer way to do it.

PR #2884, opened 2025-08-13
fix package issue link and description

SillyFreak:
When going to https://github.com/typst/packages/issues/new/choose, the "Package-specific issue" link target does not exist and I assume the description also refers to the UI on the old site. This PR updates them.

I chose the search page as the link target, but I think the Universe home page would work too.


elegaanz:
Thanks!

PR #2889, opened 2025-08-15
umbra:0.1.1

davystrong:
I am submitting
- [ ] a new package
- [X] an update for a package

Replaced `path` with the new `curve` function. This removes deprecation warnings but shouldn't change anything else.


davystrong:
If anyone can give me any guidance, do I need to resolve that failed test? It's for a package that isn't on the newest version (specifically, the random number generator for one of the examples has updated from 0.3.0 to 0.4.0). However, updating to the newest version results in a different output, so I'd rather not.

elegaanz:
Then you can safely ignore the warning. And thank you for this update.

PR #2891, opened 2025-08-15
Presentate:0.2.0

pacaunt:
I am submitting
- [ ] a new package
- [x] an update for a package

The package has change internal system of animation states for adding relative index specification. 
Added `render` function along with `animation` module for providing functions that are compatible with other packages.


elegaanz:
Thanks!

PR #2894, opened 2025-08-15
payqr-swiss:0.3.0

philippdrebes:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Changed default behavior from forcing page breaks to floating mode, allowing Swiss QR bills to be embedded within custom invoice layouts, instead of always being printed on a new page. Added `standalone: true` parameter to restore the previous behavior when needed.


elegaanz:
Can you please delete the PDF files? They can easily be reproduced by compiling the Typst file and only add their weight to this already too large repository.

philippdrebes:
@elegaanz thanks for the heads up. The pdf files are now removed.

elegaanz:
Thank you.

PR #2904, opened 2025-08-17
add json-curriculum-vitae@0.1.0

Its-Just-Nans:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package can create a CV from a JSON


packages/preview/json-curriculum-vitae/0.1.0/README.md
```diff
@@ -0,0 +1,22 @@
...
+## Compilation
+
+```sh
+typst compile cv.typ --input CV_LANG=en CV_en.pdf
+typst compile cv.typ --input CV_LANG=fr CV_fr.pdf
+```
+
+## Lisence
+
+- [MIT](./LICENSE)
```
saecki: The license file is missing

packages/preview/json-curriculum-vitae/0.1.0/typst.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "json-curriculum-vitae"
+version = "0.1.0"
+compiler = "0.13.0"
+entrypoint = "json-curriculum-vitae.typ"
+authors = ["n4n5 <its.just.n4n5@gmail.com>"]
+license = "MIT"
+description = "Create a CV from a JSON file"
+keywords = ["CV", "Curriculum Vitae", "json"]
+categories = ["cv"]
+exclude = ["*.jpg*", "*.json", ""]
```
saecki: The empty string in the `exclude` should be removed

packages/preview/json-curriculum-vitae/0.1.0/typst.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "json-curriculum-vitae"
+version = "0.1.0"
+compiler = "0.13.0"
+entrypoint = "json-curriculum-vitae.typ"
```
saecki: This name seems pretty generic, could you change it something else? We try to avoid too canonical names, so that multiple packages serving the same purpose can coexist. See our [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

packages/preview/json-curriculum-vitae/0.1.0/json-curriculum-vitae.typ
```diff
@@ -0,0 +1,390 @@
...
+      v(0.8em),
+      translate(data.subtitle),
+    ),
+  )
+}
+
+
+#let show_cv(json_path: "cv_data.json", lang: "en") = {
+  let path_json = sys.inputs.at("CV_JSON", default: json_path)
+  let data = json(path_json)
```
saecki: This won't work, because paths are package scoped. The path passed to this function will be searched inside of this package directory, not directory in which the user is compiling their CV. Instead you can accept the json data directly and instruct the user to use `json("path.json")` when calling this function. This will also have the advantage of supporting other data formats.

packages/preview/json-curriculum-vitae/0.1.0/README.md
Its-Just-Nans: fixed

packages/preview/json-curriculum-vitae/0.1.0/json-curriculum-vitae.typ
Its-Just-Nans: fixed

packages/preview/json-curriculum-vitae/0.1.0/typst.toml
Its-Just-Nans: Not a big fan of these naming rules but anyway I fixed that

packages/preview/porygon/0.1.0/README.md
saecki, line 1: Unless I'm missing something, I don't see any instructions on what kind of JSON input is expected by this package. If there is some sort of standard format that I'm not aware of, could you please link that? Otherwise there should be at least one example of a json data file.

Its-Just-Nans: The json can be build from the typst errors

saecki: I'm sorry, but I don't see how that's supposed to work. While typst universe is not exactly a curated list of packages, we try to make sure all packages are at least reasonably usable by other people, and that includes [writing documentation](https://github.com/typst/packages/blob/main/docs/documentation.md#writing-package-documentation). I assume you have a local example of a json file that you used for testing. Just replacing that with some mock data and including it in the readme would be enough. Alternatively you could also document the schema of the json data.

Its-Just-Nans: OK
I have added a template

saecki: Thanks!

PR #2917, opened 2025-08-20
tlacuache-thesis-fc-unam:0.1.2

davidalencia:
I am submitting
- [ ] a new package
- [x] an update for a package

This is a thesis template for the Science Faculty at Universidad Nacional Autónoma de México (UNAM) based on my thesis.


saecki:
Thanks for this update!

PR #2925, opened 2025-08-20
zero:0.5.0

Mc-Zen:
I am submitting
- [x] an update for a package

## Changelog
- Adds a new unit construction method that allows for more complex units involving math, symbols or basically anything. 
- ⚠️ Breaking Change: The rounding setup has been streamlined. The `mode` can no longer be `none`, instead it defaults to `"places"` while the default precision is `none` meaning that no rounding is applied. This is more convenient when rounding single numbers (e.g., `num(round: (precision: 2))[9.80665]`) because the mode does not have to be set repeatedly.


elegaanz:
Thank you!

PR #2953, opened 2025-08-26
retrofit:0.1.2

xkevio:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog

- fix: don't error when an unknown key is encountered
- fix: allow explicit usage of `#cite()`


elegaanz:
Thank you.

PR #2958, opened 2025-08-27
herodot:0.2.0

gnist-dev:
I am submitting
- [ ] a new package
- [ X ] an update for a package

Description: Two new fields have been added allowing for more manual customization of the look of the timeline (if the user wants to), and making it possible to also use the package not just for horizontal but also vertical timelines.


gnist-dev:
Accidentally used the wrong branch, closed it as a result (sorry).

PR #2959, opened 2025-08-27
herodot:0.2.0

gnist-dev:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added two new fields to allow for manual setting of event and eventspans's text title rotation, to allow for making vertical timelines and more manually set styling.


elegaanz:
Thank you.

PR #2966, opened 2025-08-28
pergamon:0.1.0

alexanderkoller:
I am submitting
- [x] a new package
- [ ] an update for a package

Pergamon is a package for typesetting bibliographies in Typst. It is inspired by BibLaTeX, in that the way in which it typesets bibliographies can be easily customized through Typst code. Like Typst's regular bibliography management model, Pergamon can be configured to use different styles for typesetting references and citations; unlike it, these styles are all defined through Typst code, rather than CSL.

Pergamon has a number of advantages over the builtin Typst bibliographies:

- Pergamon styles are simply pieces of Typst code and can be easily configured or modified.
- The document can be easily split into different refsections, each of which can have its own bibliography (similar to Alexandria).
- Paper titles can be automatically made into hyperlinks - as in [blinky](https://typst.app/universe/package/blinky/), but much more flexibly and correctly.
- Bibliographies can be filtered, and bibliography entries programmatically highlighted, which is useful e.g. for CVs.
- References retain nonstandard Bibtex fields (unlike in Hayagriva), making it e.g. possible to split bibliographies based on keywords.


alexanderkoller:
Dear maintainers, regarding the failing checks:

- Two of them are caused by me importing oxifmt:0.2.1, which was current when I started implementing the package a few weeks ago. I have now updated the dependency to oxifmt:1.0.0 in my codebase, so this problem would resolve itself when I submit the next version of Pergamon.
- The third one concerns a positional function parameter that is not in kebab case. I have fixed this too in my codebase, but users of Pergamon will never notice the difference (because it's a positional parameter).

It would be nice if you could approve the pull request anyway. Thanks!

alexanderkoller:
I suppose I could just submit another pull request with these changes made. Let me do that.

PR #2974, opened 2025-08-29
Use correct units for information about file and image

Andrew15-5:
DPI is for printers, not digital images. MB is multiple of 10, which code uses
multiple of 2, which is MiB. There must be a single whitespace between quantity
number and unit of measurement.

After RTFM, I refreshed my memory and PPI in the CLI and this docs is
incorrect. PPI refers to physical size, not digital. As such, if you keep the
image size (in pixels), PPI will not change its appearance when viewed
digitally. So providing `--ppi` is confusing and imprecise. Moreover, the
default is 144 PPI, which will not translate `set page` pt to px like I would
assume. For that, you would have to use `--ppi 72`. Compiler instead must just
set some EXIF image tag that describes the set PPI. As for image size, perhaps
a `--scale` ratio or `--width` px flag should be added, to easily get the right
image size depending on how you configure your `set page` or your requirements.

### DIFF
--- a/bundler/src/main.rs
+++ b/bundler/src/main.rs
@@ -404,7 +404,7 @@ fn process_thumbnail(
     // Get file size.
     let file_size = fs::metadata(&original_path)?.len() as usize;
     if file_size > 3 * 1024 * 1024 {
-        bail!("thumbnail must be smaller than 3MB");
+        bail!("thumbnail must be smaller than 3 MiB");
     }
 
     let mut image = image::open(&original_path)?;
@@ -412,7 +412,7 @@ fn process_thumbnail(
     // Ensure that the image has at least a certain minimum size.
     let longest_edge = image.width().max(image.height());
     if longest_edge < 1080 {
-        bail!("each thumbnail's longest edge must be at least 1080px long");
+        bail!("each thumbnail's longest edge must be at least 1080 px long");
     }
 
     // We produce a full-size and a miniature thumbnail.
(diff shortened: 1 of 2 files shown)


elegaanz:
Regarding the `--ppi` flag, my understanding is that it does change the image size (in pixels): Typst measures the page dimension in "real world" units (i.e inches), `px` is not a valid unit in Typst. So asking for a higher PPI will render the page with more pixels.

Andrew15-5:
> Regarding the `--ppi` flag, my understanding is that it does change the image size (in pixels)

Yes, and it's not correct. See [this](https://www.reddit.com/r/photoshop/comments/17dz73h/why_reducing_ppi_doesnt_reduce_file_size/) and [this](https://asktimgrey.com/2024/01/02/file-size-and-ppi-resolution/) as quick Internet search results. And also importantly, it's confusing and unpredictable (unless some formula can be used that you have to always keep in mind): if I need specific image size, I can't directly use `--ppi` or `page.width`, the workaround is `--ppi 72` and use `page.width` as pixels, if I need specific PPI and image size, Typst won't allow me to set them separately to given requirements.

Regardless, it doesn't block the PR changes. Further changes and discussions can take place separately.

elegaanz:
Yes, I will let you report that somewhere else (even if I think the current behavior is correct, but maybe I'm mistaken and it's worth discussing it). Thank you for making these fixes to the docs. :)

Andrew15-5:
Here you go: https://github.com/typst/typst/issues/6846.

PR #2975, opened 2025-08-29
polytonoi:0.2.0

dei-layborer:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update to provide two fixes: (1) ~ accent not combining properly, and (2) letter eta not being treated as a vowel.

### DIFF
--- a/packages/preview/polytonoi/0.2.0/polytonoi.typ
+++ b/packages/preview/polytonoi/0.2.0/polytonoi.typ
@@ -1,6 +1,6 @@
 // assigns a Roman character to a code point representing its Greek equivalent
 // anything not in this list (and not handled elsewhere) will be rendered literally
-#let letterDictionary = (
+#let letter-dictionary = (
   // lower-case letters
   a: 0x03B1,
   b: 0x03B2,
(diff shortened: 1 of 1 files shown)


packages/preview/polytonoi/0.2.0/typst.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "polytonoi"
+version = "0.2.0"
+entrypoint = "polytonoi.typ"
+authors = ["Dei Layborer"]
+license = "GPL v. 3.0"
```
elegaanz: The correct syntax is:

```suggestion
license = "GPL-3.0"
```

packages/preview/polytonoi/0.2.0/README.md
```diff
@@ -0,0 +1,88 @@
+# Polytonoi | Πολλοί Τόνοι
+A typst package for rendering text into polytonic Greek using a hopefully-intuitive transliteration scheme.
+
+## Usage
+
+First, be sure you include the package at the top of your typst file:
+
+```typ
+@import "preview/polytonoi@0.1.0: *
```
elegaanz:
```suggestion
@import "preview/polytonoi@0.2.0: *
```

elegaanz (commented):
Two small things to fix. It would also be nice if you could rename your public variables to kebab-case as suggested by CI, and/or mark them as private by prefixing them with a `_` if they are not supposed to be used by the user (note that adding this underscore is only a convention, it doesn't prevent anyone from actually accessing them).

elegaanz:
Thank you!

PR #3007, opened 2025-09-04
tgm-hit-thesis:0.4.0

SillyFreak:
I am submitting
- [ ] a new package
- [x] an update for a package

This is an official thesis template for the informatics department of the TGM technical high school in Vienna. This update fixes some lingering warnings when using Typst 0.13 (and requires 0.13 as the minimum version). It also removes the suggested outline, as students tend to take it as required, and improves the internal package organization.


elegaanz:
Thank you.

PR #3009, opened 2025-09-04
dashy-todo:0.1.1

Otto-AA:
I am submitting
- [ ] a new package
- [x] an update for a package

Fix bug that outline only included inline todos (worked before 0.1.0, and now it's fixed again).


elegaanz:
Thank you!

PR #3011, opened 2025-09-05
transl:0.1.1

mayconfmelo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- Fixed: `#transl(mode: str)` used to return `content` because of state update
- Updated: `#transl(data)` don't update database when `#transl(mode: str)` anymore
- Updated: Fluent arguments `#transl(args: (foo: bar))` &rarr; `#transl(foo: bar)`


saecki:
Thanks for the update!

PR #3014, opened 2025-09-05
nikonova:0.1.1

atraphaxia:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Made nikonova-box() breakable, thus fixing issues with long outlines and problems. Added title and level parameters to nikonova-problem(), allowing you to use them in more or less nested headings. Removed the license header from the template. Set the author and font fields in template. Made the default values of nikonova() more generic. 


saecki:
Thanks for the update!

PR #3016, opened 2025-09-05
meander:0.2.0

Vanille-N:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes from 0.1.0:

Syntax
- changed to use code instead of content

Obstacle control
- retiling functions added to control the precise boundaries of obstacles
- margin around obstacles is now parameterizable
- phantom objects that do not count as obstacles

Typesetting
- contextually correct justified text
- experimental support for hyphenation

### DIFF
--- a/packages/preview/meander/0.2.0/README.md
+++ b/packages/preview/meander/0.2.0/README.md
@@ -3,7 +3,8 @@ _Text threading and image wrap-around for Typst._
 
 `meander` provides a core function `reflow` to segment a page and wrap content around images.
 
-See the [documentation](https://github.com/Vanille-N/meander.typ/tree/v0.2.0/docs/main.pdf).
+<!-- @helper(if publish; grep https) -->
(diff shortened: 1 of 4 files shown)


packages/preview/meander/0.2.0/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Quick start
+
+The function `meander.reflow` splits the input sequence into
+- obstacles: all objects created via the function `placed` (which is like `place`),
+- containers: produced by `container`, optionally specifying an alignment and dimensions,
+- flowing content: produced by `content`.
+
+```typ
+#import "@preview/meander:0.2.0"
```
saecki, line 18: The first example looks like this on my machine, which seems a little unfortunate:
image

Vanille-N: Suboptimal indeed. Thanks for catching that, the behavior itself is a known bug but the fact in occurs here was a typo.

PR #3025, opened 2025-09-07
penpo:0.1.0

Vanille-N:
I am submitting
- [X] a new package
- [ ] an update for a package

Description: penpo is a spellchecker and transliteration library for toki pona (a conlang). toki pona having many writing systems, a dedicated library to manage them is necessary in order to render documents in multiple different alphabets.


Vanille-N:
Oops, gotta fix a link in the readme.

packages/preview/penpo/0.1.0/O-LUKIN.md
```diff
@@ -0,0 +1,167 @@
...
+  ilo Insanjuwisi /sp/
+  (toki Inli: "#penpo.esc[Ingenuity]") li ken tawa lon kon. /sp/
+  ilo ante li tawa sike e mun Masi li pali e sitelen pi selo ona. /sp/
+
+  mun lili tu li sike e mun Masi. /sp/
+  nimi ona li mun Popo li mun Temo. /sp/
+  ona li kiwen li sike lukin ala. /sp/
+][
+  #figure(
+    image("assets/mun-Masi.jpg"),
```
saecki: The example doesn't compile because these images are missing. Consider removing the figures, so this example can be directly copy pasted.

packages/preview/penpo/0.1.0/src/nimisin.typ
```diff
@@ -0,0 +1,128 @@
...
+    }
+  } else {
+    initials.update(seen => {
+      seen.insert(key, (nimi, spelling))
+      seen
+    })
+  }
+}
+
+#let nimisin(word, letters, _lili: auto) = context {
```
saecki: Is there any reason this argument is prefixed by an underscore? Otherwise could you fix the warnings shown by the package check?

packages/preview/penpo/0.1.0/O-LUKIN.md
Vanille-N: Thanks for the suggestion. I sketched a placeholder instead.

saecki:
Sorry, I pressed `start review` instead of `add comment` yesterday, so the second comment didn't show up.

packages/preview/penpo/0.1.0/src/nimisin.typ
Vanille-N: Right. The original reasoning was that this function should have the same parameters as `nimisin-mute`, and since `nimisin-mute` can in principle accept an arbitrary word, I wanted a parameter name that cannot be a word. I'll think of a fix to remove the warning and keep a consistent interface.

saecki:
Thanks!

PR #3029, opened 2025-09-07
rowmantic:0.5.0

bluss:
I am submitting
- [ ] a new package
- [x] an update for a package

Rowmantic introduces the `rowtable` function as a row-oriented "frontend" to the usual `table` function in typst.

- **Change**: `separator: " "` (a space) now splits on spaces and whitespace in markup
- **Add**: `rowgrid` function which forwards to `rowtable` but with other defaults and creates a grid; rowtable now supports both grids and tables when the appropriate options are passed.
- **Change**: The representation of `expandcell()` changed (but the representation is internal anyway)

In addition, the .svg files for the readme have been updated slightly, so they have changed, even if visually they are almost exactly the same as before.


packages/preview/rowmantic/0.5.0/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+<!--
+### Installation
+
+TBD
+-->
+
+## Function Documentation and Manual
+
+Please <a href="docs/rowmantic-manual.pdf?raw=1">see the manual</a> for more explanations, examples and function documentation.
```
bluss, line 91: Oh I was going to ask - this link change is new - `?raw=1` was added to this link. This makes the link work best on github. But Typst universes' readme is the highest priority - so does this work well for the published package? Ideally it resolves the same way it already did, with no change.

If you think it causes a problem, let's fix that before this upload.

saecki: I found [this](https://typst.app/universe/package/isc-hei-report/0.5.3/) package, where it doesn't work. But just using a normal markdown link would work, and will be converted to a raw url on typst universe.

EDIT: never mind the files are just missing on the example

I *think* this should work, since just appending `?raw=1` to a `raw.githubusercontent` url also works.

saecki:
Thanks for the update!

packages/preview/rowmantic/0.5.0/README.md
saecki: It works :+1:

bluss:
Thanks for the review :slightly_smiling_face:

packages/preview/rowmantic/0.5.0/README.md
bluss: thanks for checking this before/after publish!

PR #3032, opened 2025-09-07
datify-core:1.0.0

Jeomhps:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: 

Hi, I'm Jeomhps, the maintainer of Datify. Someone once pointed out that Datify was combining two major aspects: a localization API and user-friendly formatting functions. When I discovered the CLDR project—an extensive source of localization data—I realized it was the perfect backend for this kind of project. While Datify had grown thanks to contributions from the Typst community, CLDR offers much broader coverage, so I decided to build datify-core as a dedicated API library for localization, powered by CLDR.

Datify-core is open to further contributions from the Typst community, and I'd love to see it continue growing together—your help has always been invaluable. In the near future, Datify itself will be updated to use datify-core as its backend, which may introduce some breaking changes (I’ll be sure to notify users in advance). With this new structure, both projects will be more maintainable, scalable, and useful for anyone needing localized date patterns in Typst!


Jeomhps:
Hmm, I realize the file is huge—about 106,000 lines! I wasn't quite sure of the best approach, so I started by putting all the data into one big file. If you prefer, I can split it into smaller files, one per locale. For now, I haven’t implemented user-selectable locale imports because I’m not sure if it’s even possible in Typst, but it’s something I’d like to keep in mind for the future.

If I do split the data into several files, each getter would need to load the relevant TOML file as a dictionary at runtime. My concern is that if you call the getter multiple times, it might re-import the file every time, which could impact performance. As an insight, with the current setup (one big file and about 85k tests), the getter takes around 1.5 seconds to compile, which seems reasonable—but I’m not sure how splitting would affect that. I’m open to suggestions or advice on the best path forward!

Jeomhps:
Today, I did some tests and may have found a possible approach to allow users to import only a selected number of locales.

For example, if I define a locale (say, `fr`) as a dictionary in `main` like this:
```
#let fr = (a: 1, b: 2)
```
Then, in another file, I can do:
```
#import "main.typ": fr
```
This could enable users to choose which locale they want to import. The imported locale would provide a dictionary containing all the data for that locale, which would then be passed to the getter, and the getter could just iterate over the dictionary.

This approach looks practical, but I’m questioning how it would work with `datify`, since importing `datify` already imports everything from `datify-core`. Would users still be able to cherry-pick what they want by specifying the import, considering that imports from `datify-core` are merged into `datify`?

At this point, I think there are still too many open questions to transition to multi-file TOML for locales. If the team agrees, I’d prefer to keep using the big file for now and maybe revisit this dynamic import method for a v2 in the future.

Jeomhps:
I fixed a few things, especially around data access and the API contract:

- The user should not be able to import or access the raw CLDR data directly. As this is meant to be an API, exposing the underlying data defeats the purpose and breaks encapsulation.
- I changed get-date-pattern to only accept API pattern keys ("full", "long", "medium", "short", etc.). Previously, I confused datify and datify-core—custom pattern strings should be handled by datify, not the core API.
- This reinforces my view that granular importing of locale data isn’t practical: if users could import, say, a "fr" dictionary, they’d have the raw data and could bypass the API—which isn’t the intended usage.
- For maintainability, I still plan to split the TOML into several files in the future, so contributors can add or improve locales more easily—especially for under-covered languages.
- As an aside, I considered loading the TOML data inside each getter for "just-in-time" access, which could be more performant for a few calls, but would be much slower for scenarios with massive repeated calls (like my 86k-call test). I’ll benchmark with tytanic before making any changes in the future —for now, the current approach stays.

And of course, I re-runed my tt test on the files to make sure everything is properly working and it passes so it should be ready to go to me, if you guys accept the big toml.

saecki:
Thanks for this package! I think the large file is fine, especially since it is generated from the CLDR data anyway right? I don't really see a point in splitting up the file to make it maintainable, since I think changes should be directly contributed to the CLDR project.

Jeomhps:
Hey, thanks for merging the package! You're right—since the data comes from CLDR, it's not really a problem. I just thought some people might want to contribute translations, like they did before with datify, so I wanted to keep that option open.

I also considered splitting things up because I might want to support multiple data providers in the future. For example, I could use CLDR JSON now, but maybe later add another provider with data that isn’t yet marked as "contributed" or "approved." But for now, this is just an idea. Typst doesn’t let you mutate variables inside functions, so caching isn’t possible at the moment. Also, having multiple providers could make the package too large, so I’m still figuring out what’s best.

Anyway, thanks again for merging!

PR #3020, opened 2025-09-06
definitely-not-isec-slides:1.0.0

ecomaikgolf:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Touying slide theme for the ISEC TUGraz Institute. Followup package after https://typst.app/universe/package/definitely-not-isec-thesis

I have read and followed the submission guidelines and, in particular, I
- [x] selected [a name](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) that isn't the most obvious or canonical name for what the package does
- [x] added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys
- [x] added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package
- [x] have chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] [`exclude`d](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude) PDFs or README images, if any, but not the LICENSE

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #3036, opened 2025-09-08
definitely-not-isec-slides:1.0.1

ecomaikgolf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixes https://github.com/typst/packages/pull/3020#issuecomment-3265810045. The images on Typst Universe are stretched, so I removed the specific width and height so the renderer can infer them. Asked if it could be amended but I also made a small compatible change to the default `template/slides.typ` showing off more default options, so I guess 1.0.1 makes sense.

See https://typst.app/universe/package/definitely-not-isec-slides

For the rest, the only change in the template code is adding the default parameters explicitly in the template so users can quickly tweak them.

```diff
[ecomaikgolf@laptop ../preview/definitely-not-isec-slides/]$ diff 1.0.0/template/slides.typ 1.0.1/template/slides.typ 
1c1
< #import "@preview/definitely-not-isec-slides:1.0.0": *
---
> #import "@preview/definitely-not-isec-slides:1.0.1": *
5a6,8
>   progress-bar: true,
>   institute: [isec.tugraz.at],
>   logo: [#tugraz-logo],
17a21
>       primary: rgb("e4154b"),
```


ecomaikgolf:
And updated `cetz` from 0.4.1 to 0.4.2

saecki:
Thanks for the update!

PR #2859, opened 2025-08-07
Updated `elegant-polimi-thesis` to version 0.1.1

VictuarVi:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:  Corrected the transparency for the sunburst and updated the documentation.

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #3037, opened 2025-09-08
elegant-polimi-thesis:0.1.1

VictuarVi:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:  Corrected the transparency for the sunburst, corrected localization and frontispiece attributes, updated documentation.

### DIFF
--- a/packages/preview/elegant-polimi-thesis/0.1.1/README.md
+++ b/packages/preview/elegant-polimi-thesis/0.1.1/README.md
@@ -8,12 +8,12 @@ Elegant and simple thesis template for [Typst](https://typst.app/), a modern typ
 They are rather similar. The main difference is the PoliBlu colour spread across the headings. I recommend to check them out.
 
 > [!NOTE]
-> See the [manual](https://github.com/VictuarVi/PoliMi-PhD-Thesis/blob/main/docs/manual.pdf) and [documentation](https://github.com/VictuarVi/PoliMi-PhD-Thesis/blob/main/docs/docs.pdf) for more information.
+> See the [manual](docs/manual.pdf) and [documentation](docs/docs.pdf) for more information.
 
 ## Preview ✨
 
 <p align="center">
-  <img alt="Frontspiece" src="thumbnail.png" width="45%">
+  <img alt="Frontispiece" src="thumbnail.png" width="45%">
 </p>
 
 ## Usage 🖋
@@ -29,7 +29,7 @@ A very simple document:
 ```typ
 #import "@preview/elegant-polimi-thesis:0.1.1": *
 
-#show: polimi_thesis.with(
+#show: polimi-thesis.with(
   title: "Thesis Title",
   author: "Vittorio Robecchi",
   advisor: "Prof. Donatella Sciuto",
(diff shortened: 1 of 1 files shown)


packages/preview/elegant-polimi-thesis/0.1.1/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+Elegant and simple thesis template for [Typst](https://typst.app/), a modern typesetting program alternative to LaTeX. I based the design of the document on the following two templates:
+
+- [PhD Thesis Template](https://www.overleaf.com/latex/templates/phd-thesis-template/nwjkggvhrzmz)
+- [DEIB PhD Thesis Template](https://www.overleaf.com/latex/templates/politecnico-di-milano-deib-phd-thesis-template/ydsvtyzwxfdk)
+
+They are rather similar. The main difference is the PoliBlu colour spread across the headings. I recommend to check them out.
+
+> [!NOTE]
+> See the [manual](https://github.com/VictuarVi/PoliMi-PhD-Thesis/blob/main/docs/manual.pdf) and [documentation](https://github.com/VictuarVi/PoliMi-PhD-Thesis/blob/main/docs/docs.pdf) for more information.
```
saecki: Continuing https://github.com/typst/packages/pull/2859#discussion_r2270481187
Yes the current link works. We just try to avoid adding PDF files to this repository when they are not needed, since they tend to take up quite a lot of disk space. If you link a `manual.pdf` and `docs.pdf` that is in another repository, there is no need to include the files here as well. Especially since they can be easily reproduced by compiling the typst files. So you can either leave the README as is, and remove both PDF files, or edit the readme to link to the local files like this: `[manual](docs/manual.pdf)`

packages/preview/elegant-polimi-thesis/0.1.1/README.md
```diff
@@ -0,0 +1,101 @@
...
+```shell
+typst c main.typ --pdf-standard a-3b
+```
+
+A very simple document:
+
+```typ
+#import "@preview/elegant-polimi-thesis:0.1.1": *
+
+#show: polimi_thesis.with(
```
saecki: This example doesn't compile, `polimi_thesis` isn't found, it should use kebab-case

packages/preview/elegant-polimi-thesis/0.1.1/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+// sommario in Italian
+
+#show: acknowledgements
+
+// acknowledgements
+
+#toc
+#list_of_figures
+#list_of_tables
```
saecki: Same here

packages/preview/elegant-polimi-thesis/0.1.1/README.md
VictuarVi: Oh I didn't understand exactly what you meant – I'll fix it by linking the local file.

VictuarVi:
All the problems related to the README should be fixed. Sorry for the trouble.

saecki:
Thanks! No worries :)

PR #3038, opened 2025-09-09
stellar-iac:0.5.0

shunichironomura:
I am submitting
- [ ] a new package
- [x] an update for a package

The template has been updated so that it can be used with the new Typst compiler (v0.13.1) without errors. Namely, it now wrap the `counter(page)` in a context expression.

It also supports figures spanning multiple columns.


saecki:
Thanks for the update!

PR #3042, opened 2025-09-09
modern-iu-thesis:0.1.2

bojohnson5:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I've updated the template to include another section for showing code listings in the front matter.


packages/preview/modern-iu-thesis/0.1.1/lib.pdf
saecki, line 1: Could you restore this file, we don't allow any edits to already published packages, to ensure reproducibility.

bojohnson5: Should be there now! Thanks!

saecki: Unfortunately the edit still shows up. Could you try to remove the last comit, amend [this](https://github.com/typst/packages/pull/3042/commits/9cab4297f9effee3a933af4ca6128e9c3a662e73) commit to only remove the `main.pdf` files, and then force push instead?

bojohnson5: Hi let me know if that worked. I'm not the most adept at git.

saecki: Oh the problem was that the `lib.pdf` was already changed in the first commit. I've restored it now.

PR #3059, opened 2025-09-11
basic-resume:0.2.9

stuxf:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I added an optional parameter that allows people to change the language of thier resume


packages/preview/basic-resume/0.2.9/docs/manual.pdf
saecki, line 1: This file will be pretty hard to discover on typst universe. Could you either link it from the readme or remove it?
See: https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

saecki:
Sorry for the slow repsonse

PR #3062, opened 2025-09-12
min-book:1.2.1

mayconfmelo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This patch removes the red background in `src/assets/frame.svg` — it was a silly development hack that I forgot to remove, — as this will probably become a bug in future Typst versions (typst/typst#6892)


saecki:
Thanks for the update!

PR #3063, opened 2025-09-12
hallon:0.1.1

mewmew:
I am submitting
- [ ] a new package
- [x] an update for a package

Updating hallon to version 0.1.1 which uses the active heading numbering for multi-level numbering of figures and subfigures (see https://github.com/mewmew/hallon-typ/issues/1).


saecki:
Thanks for the update!

PR #3071, opened 2025-09-14
modern-nenu-thesis:0.1.0

virgiling:
I am submitting
- [x] a new package
- [ ] an update for a package


packages/preview/nenu-thesis/0.1.0/template/thesis.typ
```diff
@@ -0,0 +1,374 @@
+#import "@preview/nenu-thesis:0.1.0": thesis
+#import "@preview/kouhu:0.1.0": kouhu
+#import "@preview/codly:1.1.1": codly, codly-init, no-codly
```
saecki: This version of `codly` uses a deprecated feature. Could you update it to silence the warning?
```
warning: checking whether a type is contained in a string is deprecated
   ┌─ @preview/codly:1.1.1/args.typ:90:24
   │
90 │     if arg.function and function in arg.ty {
   │                         ^^^^^^^^^^^^^^^^^^
   │
   = hint: this compatibility behavior only exists because `type` used to return a string
```

packages/preview/nenu-thesis/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "nenu-thesis"
```
saecki: If this isn't an official package, you'll need to add a non-descriptive prefix so multiple packages for the same university can coexist. See: https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules

saecki (commented):
Thanks for this package! I've some comments regarding the name and an outdated version of `codly`. If you don't strongly prefer snake case, could you also fix the snake-case warnings from the package check?

virgiling:
Thanks for reviewing! I will update and modify this package as soon as possible.

saecki:
Thanks!

PR #3072, opened 2025-09-14
latex-lookalike:0.1.4

mewmew:
I am submitting
- [ ] a new package
- [x] an update for a package

Adding `show: latex-lookalike.style` show rule to simplify use of package. Also handles abstract and `make-title` command from LaTeX.


saecki:
Thanks for the update!

PR #3077, opened 2025-09-15
diagraph:0.3.6

Robotechnic:
I am submitting
- [ ] a new package
- [x] an update for a package

- Fixed Robotechnic/diagraph#56 caused by uninitialized html struct member of cluster labels
- Fixed Robotechnic/diagraph#36 caused by an invalid free on error
- Fixed an invalid memory access caused by the math mode detection
- Fixed the margin of subgraphs clusters being set to 0


elegaanz:
Thank you.

PR #3078, opened 2025-09-15
May:0.1.0

Carlos-Mero:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I have refactored the original code in may template, and added a theme for touying slides, along with some updates in README and more showcases. I believe slides created with this theme will look great.


elegaanz:
Thank you.

PR #3085, opened 2025-09-15
efter-plugget:0.1.0

mewmew:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: a template for University lab reports, essays and exams.

Just replace the logo with the one of your University to use the template.


mewmew:
**Note:** this PR depends on `latex-lookalike:0.1.4`, which is submitted in #3072.

elegaanz:
Can you please change the name to something less generic? Also, you should add at least one category in the manifest, [as reported here](https://github.com/typst/packages/actions/runs/17747437066/job/50435573749?pr=3085#step:7:18).

mewmew:
> Can you please change the name to something less generic? Also, you should add at least one category in the manifest, [as reported here](https://github.com/typst/packages/actions/runs/17747437066/job/50435573749?pr=3085#step:7:18).

Hi Ana!

Thanks for the review.

I've now renamed the package to "efter plugget" which is a tribute to a Swedish song about finishing Uni.

I've also added the "report" category to the Typst manifest, as this package is intended for lab reports, Uni assignments, etc.

Cheers,
/robin

mewmew:
Converted to draft for now, as I need to update some show rules for figures before publishing the package.

mewmew:
> Converted to draft for now, as I need to update some show rules for figures before publishing the package.

Now the package is ready for review again : )

It depends on the latest version of the `hallon` package (see #3103).

packages/preview/efter-plugget/0.1.0/typst.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "efter-plugget"
+version = "0.1.0"
+entrypoint = "src/lib.typ"
+authors = ["Robin Eklind"]
+license = "0BSD"
+description = "Lab report, essay and exam template for Typst."
+repository = "https://github.com/mewmew/efter-plugget"
+categories = ["report"]
+exclude = ["inc"]
```
elegaanz: This also matches `template/inc`, can you please replace it with `/inc` to only match the top-level directory?

mewmew: > This also matches `template/inc`, can you please replace it with `/inc` to only match the top-level directory?

Ah, wow. Good catch! Didn't know that's how it worked. I'll send an update.

**Edit:** Done in https://github.com/typst/packages/commit/014b0826e52f97b5294fb5c9397a962eb6e870f5

elegaanz:
Thank you.

mewmew:
@elegaanz, thank you for the review :)

I tried to use `efter-plugget:0.1.0` from `Typst.app`.

However, it seems like the webapp is using an old/wrong version of the package?

See the [following exmple](https://typst.app/project/rTZiVnEK3xVCsHftDPfhN4) which was generated from the `efter-plugget:0.1.0` template (no changes made). It fails to compile, and gives the following error:

"File not found (searched at inc/default_logo.png)."

efter_plugget_file_not_found

When I try to do the same locally it works without error.
```bash
typst init @preview/efter-plugget:0.1.0
cd efter-plugget
# Compiles successfully.
typst compile report.typ
```

So it seems like what was deployed for the `Typst.app` is not the same as what is stored in the packages repository? It seems like it uses an old version.

Note, the image imported is `/inc/default_logo.png` in the `0.1.0` version in the `typst/packages` repository. https://github.com/typst/packages/blob/38d2fc2566e7b46a811dfd8dddc0859f9a8476cd/packages/preview/efter-plugget/0.1.0/src/lib.typ#L8

My guess is that what was deployed to `Typst.app` uses a slightly older version which imports `inc/default_logo.png`? Or that the exclude rule didn't take effect (https://github.com/typst/packages/pull/3085#discussion_r2367081478)?

elegaanz:
I think paths are normalized to be relative to the project root before being displayed, at least in the web app, so the missing leading slash is not relevant here. The web app should use the exact same archive as the CLI, I wonder if maybe there was some kind of race condition that made the editor load before the image file was actually created, did you try to close and re-open the project, to force it to reload all the files?

mewmew:
> I think paths are normalized to be relative to the project root before being displayed, at least in the web app, so the missing leading slash is not relevant here. The web app should use the exact same archive as the CLI, I wonder if maybe there was some kind of race condition that made the editor load before the image file was actually created, did you try to close and re-open the project, to force it to reload all the files?

I tried again and get the same results. [Typst.app example project](https://typst.app/project/r2wY5ms7OwyvWjbvUIDO4z)

screenshot_2025-09-23_15:37:56

It should be easy to reproduce this:

1. Go to https://typst.app/universe/package/efter-plugget and click "Create project in app".
2. Notice error "File not found (searched at inc/default_logo.png).

Do the same but from the command line.

```bash
typst init @preview/efter-plugget:0.1.0
cd efter-plugget
# Successful compile.
typst compile report.typ
```

Notice that in the web app we get a "file not found" error, and from the command line it works. But using the same version of efter-plugget. At least as far as I can tell.

Please try to see if you can reproduce. And then we know if it is only on my machine.

Cheers,
Robin

elegaanz:
I can reproduce the issue, both in the web app and using the CLI. The issue is that the `/inc` directory should not have been excluded at all in the first place, since you need the `default_logo.png` file for your package to compile. The thumbnail is automatically excluded anyways. Sorry I didn't catch that during review.

Can you please publish a new version, without the `exclude` key in the manifest, to fix the issue?

mewmew:
> I can reproduce the issue, both in the web app and using the CLI. The issue is that the `/inc` directory should not have been excluded at all in the first place, since you need the `default_logo.png` file for your package to compile. The thumbnail is automatically excluded anyways. Sorry I didn't catch that during review.
> 
> Can you please publish a new version, without the `exclude` key in the manifest, to fix the issue?

No worries! And thanks for reproducing it.

I updated the `exclude` key to only exclude the thumbnail:

https://github.com/typst/packages/pull/3130/files#diff-d4b1b588e17cbb4799c081fa32a8d7fdea43e025ab5b9e876b84b91dc7f2986eR10
```toml
exclude = ["/inc/thumbnail.png"]
```

Hope this should work :)

PR #3098, opened 2025-09-17
may:0.1.1

Carlos-Mero:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: I have updated the dependencies and some styles in this template, so that it would look better and is easier to use.


saecki:
Thanks!

PR #3099, opened 2025-09-17
ctyp:0.2.0

HPDell:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Enable setting numbering formats of headings via argument `heading-numbering`. Allow a scalar configuration applied to all headings, or a list for level-by-level settings.


saecki:
Thanks for this update!

PR #3113, opened 2025-09-20
meander:0.2.2

Vanille-N:
I am submitting
- [ ] a new package
- [X] an update for a package

This update focuses on styling options and support for more convenient multi-container pages, with also behind-the-scenes work on making the internal logic reentering to support more complex layouts in the future.

### Bisection
- now capable of splitting links and aligned content
- hyphenation now auto-detects the language
- parameters can be passed to `content` to update the style locally

### Segmentation
- hack to solve the issue where a single `container` on the page will overflow vertically
- `colbreak` jumps to the next container
- full containers now count as obstacles to other containers
- containers have a configurable margin

### Threading
- box spacing now respects `text.size` and `par.leading`.
- styling options can be passed to `container` to apply alignment and color post-hoc
- placement options can control the relationship to other content on the page

### Internal
- tiling algorithm is now interactive, which will enable more complex layouts in the future

### Bugfix
- "unreachable" path was hit

### DIFF
--- a/packages/preview/meander/0.2.2/LICENSE
+++ b/packages/preview/meander/0.2.2/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Neven Villani
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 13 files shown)


elegaanz:
Thanks!

PR #3114, opened 2025-09-20
matofletcher:0.1.1

Andrew15-5:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: [This](https://codeberg.org/Andrew15-5/matofletcher) is a fixed fork of unmaintained https://github.com/3akev/autofletcher. Until upstream is maintained again, this should be a drop-in replacement.


packages/preview/matofletcher/0.1.1/README.md
```diff
@@ -0,0 +1,38 @@
...
+## Credits
+
+- [fletcher]
+- [autofletcher]
+
+[Matofletcher]: https://codeberg.org/Andrew15-5/matofletcher
+[fletcher]: https://github.com/Jollywatt/typst-fletcher
+[autofletcher]: https://github.com/3akev/autofletcher
+[an open and untouched PR]: https://github.com/3akev/autofletcher/pull/1
+[manual]: https://codeberg.org/Andrew15-5/matofletcher/raw/branch/main/manual.pdf
```
elegaanz: Can you use a link with an explicit tag or commit hash, please? This way, even if you publish updates to your repository in the future, the manual link will still be in sync with what is on Universe. And also, can you delete the copy of the `manual.pdf` here, if you don't link to it at all? Alternatively, you could keep it and link to it directly here, which would also make sure it is never out of sync.

Andrew15-5: Right. Is this fixed now?

P.S. I do link to `manual.pdf`, just not relative to the repository it lives in. I also wanted to use the raw link as in original repo, but I think it's not too bad to open the human-readable version while the readme source is auto-locked via relative path (I also can't provide a link to the current commit, only to previous ones).

elegaanz: Yes, it looks good now, thanks.

Andrew15-5:
https://github.com/typst/packages/actions/runs/17913864855/job/50932732364

Is this a network hiccup on the action runner that caused the deploy to fail? I would've thought that this CI thing is more robust. But I guess it doesn't matter where it's run, the universal issues apply everywhere, right?

elegaanz:
Yeah, looks so. I've seen a similar errors in a few other GitHub Action jobs recently, maybe they changed something in their network configuration that made it less reliable, or they are performing some maintenance :woman_shrugging:

PR #3115, opened 2025-09-21
dmi-basilea-thesis:0.1.2

Nifalu:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
This update introduces a single, unified heading function, eliminating redundant code and allowing for arbitrary nesting of subsections. It also refactors section definitions into an array, where each element corresponds to a specific subsection level and defines its styling, thereby removing the need for individual subsection variables.


packages/preview/dmi-basilea-thesis/0.1.2/docs/manual.pdf
elegaanz, line 1: To make the manual easier to discover, can you link to it in your README, please?

elegaanz:
Hey, just a quick reminder that this PR is still open. Could you please take some time to apply my above suggestion (or just tell me if you don't think it is relevant and don't want to), so that we can get this version published?

elegaanz:
Hello, I'll close this PR for now to keep things organized, but please open a new one once you have fixed my previous comment.

PR #3119, opened 2025-09-22
herodot:0.3.0

gnist-dev:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update to provide the ability for the user to set how the date information next to an event is to be displayed as one of either day/month/none or day-month/day-month-year rather than the previous default of always having the year on display (and it being the only option). See gnist-dev/herodot#2 for more.


elegaanz:
Thank you.

PR #3121, opened 2025-09-22
arkheion:0.1.1

mgoulao:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

- Docs: Add API section with parameters for arkheion.with

- Feature: Allow abstract to be none

- Feature: Add option to override authors 


packages/preview/arkheion/0.1.1/template.typ
```diff
@@ -0,0 +1,146 @@
+#let arkheion(
+  title: "",
+  abstract: none,
+  keywords: (),
+  authors: (),
+  custom_authors: none,
```
elegaanz: Can you rename this parameter to `custom-authors`, please? This would be more in line with the rest of the package, and the broader Typst ecosystem.

mgoulao: Yes

elegaanz:
Thanks!

PR #3123, opened 2025-09-23
ezexam:0.1.9

gbchu:
I am submitting
- [x] an update for a package

### DIFF
--- a/packages/preview/ezexam/0.1.9/ezexam.typ
+++ b/packages/preview/ezexam/0.1.9/ezexam.typ
@@ -151,7 +151,7 @@
       #if chapter-location.contains(current) {
         place(
           dx: -_width - 1em,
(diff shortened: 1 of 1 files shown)


packages/preview/ezexam/0.1.9/README.md
```diff
@@ -0,0 +1,66 @@
+# `ezexam`
+## 简介
```
elegaanz: Can you restore the English description, please? It is fine to keep the chinese version, but having a small sentence explaining what the package is about in English at the beginning of the README allows for non-chinese speaking users to understand why they are not the target audience for this package.

gbchu: ok  done

elegaanz:
Thank you.

PR #3141, opened 2025-09-26
catppuccin:1.0.1

TimeTravelPenguin:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Minor deps updates & corrected minimal typst version to an older version.

If this should instead be `v1.1.0`, let me know. The only other changes are to used deps, so I didn't think it warranted a minor version bump. The compiler version changed from `0.12.0` to `0.11.0`. It was incorrectly set to match the usage in some CI workflows.


packages/preview/catppuccin/1.0.1/src/lib.pdf
elegaanz, line 1: Can you delete this file?

elegaanz (commented):
The version number seems correct too me, we are not enforcing semantic versioning very strictly anyways. Just compilation artifact to remove and the PR should be good to be merged. Thanks for the update.

packages/preview/catppuccin/1.0.1/src/lib.pdf
TimeTravelPenguin: Ah, thanks. It snuck in there! I will remove it now :)

elegaanz:
I'm not sure what happened with the git history, but commits from `main` have appear in the history of this branch. I think you need to rebase your original commit on top of `main` to fix that.

TimeTravelPenguin:
> I'm not sure what happened with the git history, but commits from `main` have appear in the history of this branch. I think you need to rebase your original commit on top of `main` to fix that.

Yeah, sorry about that. I don't know what happened, but it looks like I may have pulled something by mistake. I have been having some issues with my git client lately :/

I think things are now as they should be?

elegaanz:
Yes, looks good now, thank you.

PR #3143, opened 2025-09-26
cob-unofficial:0.1.1

mewmew:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: add abstract function and fix build of template.

(The logo used by the template was excluded in the manifest, causing the build of the template to fail.)


elegaanz:
Thank you.

PR #3144, opened 2025-09-27
Template Version 1.0.0

tibirius24:
I am submitting
- [ ] a new package
- [x] an update for a package

What changed:
- affiliation as parameter was deleted
- slide main content block height could now customized by user to prevent pagebreaks from footnotes
- renaming a function for more intuitive use


elegaanz:
Can you please rebase your branch with the latest main so that the diff only includes the version you added? Otherwise, everything looks good to me.

tibirius24:
I think that i removed all commits that are not belongig to my template.

Is it fine now?

elegaanz:
Yes, thank you.

PR #3149, opened 2025-09-28
hy-dro-gen:0.1.2

Vanille-N:
I am submitting
- [ ] a new package
- [x] an update for a package

Changelog: `hy-dro-gen` is no longer limited to languages supported natively by Typst, because it can
- compile hyphenation tries on the fly,
- dynamically load precompiled tries.

Pending some more testing, this could be a temporary workaround to typst/typst#5223


elegaanz:
Nice, thank you.

Issue #2671, opened 2025-07-08
New policy regarding package edits


elegaanz:
Starting today, minor package edits without a version bump will no longer be accepted. This change should not affect you if you don't maintain any Typst package or don't plan to.

So far, it was possible to make PR to fix small mistakes in your READMEs, manuals, or package manifests directly in a previously published version. This was convenient to quickly fix a small typo or outdated version number that wasn't caught during review.

However, this meant that packages were not fully reproducible, [as discussed here](<https://github.com/typst/packages/issues/2043>). To make it easier to ship packages with external package managers — such as Nix — and more generally to be able to check that a package archive contains the expected files using a checksum, we decided to remove this exception. Any change to a package will now have to be part of a new release.

drupol:
Excellent decision, thanks !

PR #3157, opened 2025-09-29
volt-internship-ensea:0.2.0

Dawod-G:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: v0.2.0
- Add the function `named-equation()`  
- Add a list of equations and the variable `enable-list-equations`  
- Add equation numbering based on chapter (e.g., 1.1, 1.2, 2.1)  
- Add bold to equation numbering and figure prefix, and also to figure prefix in the list of tables and list of figures  
- Move the abstract just after the acknowledgements  
- Fix the link in the outline that isn't working  
- Fix appendices in outline for headings with level > 1  
- Fix groups in glossary  
- Fix heading spacing


Dawod-G:
I am submitting an update to the template.

Furthermore, I would like to know if it is possible to delete version 0.1.0 and version 0.1.1 of this template. Indeed, those were the first versions I created, and they contain import issues: the imported files such as "abstract.typ" and "acknowledgements.typ" are incorrectly referenced in "template-internship.typ", etc..
As a result, modifying the content of these files unfortunately has no effect.

https://typst.app/universe/package/volt-internship-ensea/0.1.0/
https://typst.app/universe/package/volt-internship-ensea/0.1.1/

I would not want users to waste their time trying to edit the text in vain.

saecki:
Hey, thanks for this update! Unfortunately [our policy](https://github.com/typst/packages/issues/2671) doesn't allow any package edits.

PR #3158, opened 2025-09-29
volt-lab-ensea:0.1.1

Dawod-G:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: v0.1.1
- Disable hyphenation and paragraph justification in outline  
- Add bold to equation numbering and figure prefix  
- Fix heading spacing


saecki:
Thanks for the update!

PR #2589, opened 2025-06-25
staves:0.1.0

mdavis-xyz:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This Typst package is used to draw musical scales. It draws 5 lines, clefs, key signatures, and then a series of notes. The series can be predefined (e.g. B flat major), or a custom scale.

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name that isn't the most obvious or canonical name for what the package does
- [x] added a `typst.toml` file with all required keys
- [ ] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] tested my package locally on my system and it worked
- [x] `exclude`d PDFs or README images, if any, but not the LICENSE

My repo is: https://github.com/mdavis-xyz/staves-typst

# Packaging

I am confident in the functionality of the package itself. However I have really struggled to understand how to package it to submit to this repo. I found the instructions to be spread out over too many different places, lacking crucial details. I read all the instructions I could find multiple times, but I'm still confused. So consider this a draft, and please tell me what to do.

How do I configure the imports? `/src/core.typ` or `./src/core.typ` or `@preview/staves:0.1.0/core` or what? I have no idea. I checked [documentation.md](https://github.com/typst/packages/blob/main/docs/documentation.md),  [resources.md](https://github.com/typst/packages/blob/main/docs/resources.md) and [tips.md](https://github.com/typst/packages/blob/main/docs/tips.md). None of those mention how to do imports. One mentions that in the README you should import from `@preview` in code blocks shown to the user.

[...]

PR #3160, opened 2025-09-30
staves:0.1.0

mdavis-xyz:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: 
This package is used to draw musical scales. The use case I have in mind is writing your own musical scale books.
There are many types of scales built in (harmonic minor, natural minor etc), and it's flexible enough for users to add their own scales (e.g. broken chords). 
This package can be used to write arbitrary notes, but is not intended to be used for entire songs.

This is the same package as #2589, but with some issues fixed.


packages/preview/staves/0.1.0/docs/docs.pdf
saecki, line 1: This file will be very hard to find on Typst Universe. Could you either remove this file, or link to it from the top-level readme?

packages/preview/staves/0.1.0/examples/random-scales.pdf
saecki, line 1: Same as for `docs.pdf`

packages/preview/staves/0.1.0/examples/scale-book.pdf
saecki, line 1: Same as for `docs.pdf`

saecki (commented):
Thanks for this package! I've left a couple of comments about the included PDF files.

packages/preview/staves/0.1.0/docs/docs.pdf
mdavis-xyz: Ok, is this better? Is that the right relative link?

saecki:
Thanks!

PR #3189, opened 2025-10-03
meander:0.2.3

Vanille-N:
I am submitting
- [ ] a new package
- [X] an update for a package

# Changelog

### Segmentation
- container margin can be customized on all 4 sides
- `anchor` option on `placed` allows controlling where the alignment occurs

### Queries
- new `query` module allows using properties of previous elements during the layout
- obstacles and containers can have labels
- containers have an `invisible` option to ignore some obstacles
- implemented `query.position`
- implemented `query.width`
- implemented `query.height`

### Threading
- added `colfill` as an alternative to `colbreak`

### Convenience
- `reflow` and `regions` no longer crash if passed
  an empty sequence.
- overflow can be saved to a global state
- some error messages improved

### Performance
- better choice of horizontal marks by eliminating obvious failures
- refactored shared code between layout functions

### Bugfix
- overflow preserves styling options


elegaanz:
Thank you.

PR #3195, opened 2025-10-05
muw-community-templates:0.1.2

felixbd:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: New version with some bug fixes and slightly cleaner code ^^


elegaanz:
Thank you!

PR #3201, opened 2025-10-06
scribbling-hm:0.1.1

fine-seat:
I am submitting
- [ ] a new package
- [x] an update for a package

- added course of study variable
- titlepage adjustments
- use formatted date to match with the German language


elegaanz:
Thanks!

PR #3214, opened 2025-10-08
modern-iu-thesis:0.1.3

bojohnson5:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update equation numbering to properly reference equations in other sections.


elegaanz:
Thanks!

PR #3220, opened 2025-10-09
pubmatter:0.2.2

rowanc1:
I am submitting
- [ ] a new package
- [x] an update for a package

We have updated some of the logic around equal contributors and corresponding authors to be more flexible.

image


elegaanz:
Thank you.

PR #3240, opened 2025-10-14
modernpro-cv:1.2.1

jxpeng98:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

Allow the user to control the page margins for better customisation and avoid a cut edge caused by the printer settings.


saecki:
Thanks for the update!

PR #3244, opened 2025-10-15
unscale:0.0.1

davystrong:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package allows you to undo scaling for specific elements, such as text elements in a diagram. This is particularly useful in cases in which you want to scale a diagram based on a `layout` function, while keeping textual labels inside it constant.

If this is too simple for an entire package, feel free to reject it.

### DIFF
--- a/packages/preview/unscale/0.0.1/README.md
+++ b/packages/preview/unscale/0.0.1/README.md
@@ -22,7 +22,7 @@ Sometimes (particularly with diagrams) it is useful to be able to scale the enti
 
 The following code sets up your document to use this package:
 ```typst
-#import "../src/lib.typ": unscale, rescale, scale-rule
+#import "@preview/unscale:0.0.1": unscale, rescale, scale-rule
 
 #show scale: scale-rule
 ```


packages/preview/unscale/0.0.1/gallery/figure.typ
saecki, line 1: This example seems a little off when I compile it:
image

packages/preview/unscale/0.0.1/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "unscale"
```
saecki, line 2: This name seems a little too descriptive. We try to avoid too canonical names, so multiple packages serving the same purpose can coexist without one having an unfair advantage. See our [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

packages/preview/unscale/0.0.1/gallery/figure.typ
davystrong: That's weird... Tested again here and the CLI compiler (macos, version 0.13.0) produces the expected output for both PNG and PDF and Tinymist produces the expected output in the preview. Your emoji looks slightly different, so maybe the source is offset slightly? Is that possible?

packages/preview/unscale/0.0.1/typst.toml
davystrong: Thought this sort of fitted the "sliding" example from naming, no? Close, but not the most generic name. Haven't got any good ideas, but could change it to "delime" if you think...

packages/preview/unscale/0.0.1/gallery/figure.typ
saecki: Oh yeah, never mind, the emoji isn't even the point here.

packages/preview/unscale/0.0.1/typst.toml
saecki: I see what you mean, but if not `unscale` what would be the most descriptive name?
`delime` works :+1:

davystrong:
Based on the conversation, I'll close this PR and open a new one with the correct name.

PR #3248, opened 2025-10-15
abbr:0.3.0

jhalmen:
I am submitting
- [ ] a new package
- [x] an update for a package

This version includes two big updates:
- alternate definitions!
- @reference-style abbreviations!


packages/preview/abbr/0.3.0/example_alt.csv
saecki, line 1: As the package check warning suggests this file should be added to the `exclude` list.

jhalmen: fixed

PR #3250, opened 2025-10-15
scribbling-hm:0.1.2

fine-seat:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
- minor layout changes
- catch empty variables
- add variable list to pre-define frequently-used values


saecki:
Thanks for the update!

PR #3256, opened 2025-10-16
typsy:0.2.0

patrick-kidger:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- Added default values to (tree-)counters, usable as e.g. `tree-counter(()=>{}, default: 3)`.
- Added `safe-state` to create anonymous `states` without needing a unique string.
- Added `Pattern` to pattern-match on pattern objects themselves, e.g. `matches(Pattern, Array(..Int))`.
- Added `pattern-repr` to pretty-print pattern objects, e.g. `pattern-repr(Array(..Int)) == "Array(..Int)"`
- Bugfix: `tree-counter.subcounter.update` no longer requires `context`.
- Breaking change: class fields are now validated via pattern-matching rather than simple `type(value) == expected_type` checks.

### DIFF
--- a/packages/preview/typsy/0.2.0/LICENSE
+++ b/packages/preview/typsy/0.2.0/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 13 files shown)


saecki:
Thanks for the update!

PR #3271, opened 2025-10-17
dashy-todo:0.1.3

Otto-AA:
I am submitting
- [ ] a new package
- [x] an update for a package

Now supports inside/outside margin and rtl text direction. Also does not influence line numbering anymore.


elegaanz:
Thank you!

PR #3272, opened 2025-10-17
ergo:0.2.0

EsotericSquishyy:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This update provides more intuitive syntax for setting global and local to environments. It also provides more colorschemes and styles for users to use upon release. Users now also have the ability to use entirely custom colorschemes and styles by defining them in their own filesystem.


elegaanz:
Thank you.

PR #3285, opened 2025-10-19
pannotyp:0.0.1

voroskoi:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: PannoTyp is a package for Hungarian language support in Typst.

### DIFF
--- a/packages/preview/pannotyp/0.0.1/LICENSE
+++ b/packages/preview/pannotyp/0.0.1/LICENSE
@@ -1,0 +1,121 @@
+Creative Commons Legal Code
+
+CC0 1.0 Universal
+
+    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
+    LEGAL SERVICES. DISTRIBUTION OF THIS DOCUMENT DOES NOT CREATE AN
+    ATTORNEY-CLIENT RELATIONSHIP. CREATIVE COMMONS PROVIDES THIS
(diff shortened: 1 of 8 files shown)


voroskoi:
The licensing documentation says: "Packages must be licensed under the terms of an [OSI-approved](https://opensource.org/licenses/) license or a version of CC-BY, CC-BY-SA, or CC0".

I would like to use CC0 if it is allowed. Thanks,

elegaanz:
For Typst packages, we recommend using CC licenses only if your work is derived from other work that is itself under a CC license or if the package content is not really code but mostly data or graphics. If you want to put your code in the public domain, I can recommend the [Unlicense](https://unlicense.org/), would that suit you?

voroskoi:
Hi Ana,

Thanks for Your reply! I do not want to use Unlicense, because it is just bad. See https://chrismorgan.info/blog/unlicense/

Please let me know if CC0 is allowed or not. I can live with 0BSD as well, but at this point I am very curious what is the reasoning behind accepting Unlicense, but rejecting CC0.

Thanks,

elegaanz:
CC licenses are generally considered to be unsuitable for code, that's why I suggested an alternative. But I didn't know about these issues with the Unlicense to be honest, thanks for the link. I also think that the limitations of CC licenses that generally make them poorly suited for code doesn't really apply to CC-0 since it is not really a license and just puts your work in the public domain. I think we can say CC-0 is fine here, despite the linter error.

PR #3289, opened 2025-10-20
gantty:0.5.0

John-Toohey:
I am submitting
- [ ] a new package
- [x] an update for a package

### Added

- Task dependencies under the `dependencies` field.
- Drawers API to allow complete customization of various parts of the Gantt
  chart. This decision was made because there are many different ways of drawing
  dependencies (and, in fairness, many other parts of the Gantt chart) and
  because it will integrate better with show/set rules when they are available
  for custom elements.
- The `drawers.default-drawer` dictionary containing the default drawer
- An `x` field to intervals, tasks, milestones, and dependencies: to facilitate
  adding custom descriptions for custom drawers.
- Expose the following functions under the `util` namespace: `EPSILON`, `rects-intersect`,
  `content-if-fits`, `task-anchor`, `task-anchor-line`, `task-anchor-sidebar`,
  `id-level`, `styles-for-level`, `styles-for-id`, `foreach-task`, `task-start`,
  `task-end`, `gantt-range`, `date-ratio`, `date-x-coord`, and `date-coord`.
- Ability to provide a function to change how the sidebar items are styled.
- Hide a milestone's date with `milestone.show-date`.

### Changed

- The `taskgroups` key is now `tasks`
- The `tasks` key is now `subtasks`. All tasks can have subtasks.
- Tasks are now drawn with rectangles, not lines.

### Fixed

- Poor normalization of the `today` milestone

### Removed

- The `style` key.
- The `viewport-snap` key. Specify a `start` and `end` date instead.
- Nameless tasks.
- The `show-today` and `today-localized` keys. These should be configured on the
  milestone drawer instead.


elegaanz:
Nice update, thank you!

PR #3305, opened 2025-10-24
itemize:0.2.0

tianyi-smile:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update `itemize` to version 0.2.0

- Customize labels and bodies for enums and lists by level and item
  - 🆕 Allow per-item configuration
  - 🆕 Enable loop usage of array element values: `LOOP`
  - 🆕 If a property can be set at both the level and item, it allows passing a function
  - 🆕 When using the `*-enum-list` method, you can separately configure enums and lists: `enum-config`, `list-config`
  - Label customization
    - 🆕 `label-align`
    - 🆕 `label-baseline`
    - 🆕 `label-format`
    - 🆕 `label-width`
  - 🆕 Label alignment between items at different levels: `auto-label-width`
  - 🆕 `body-format`:
      - `text-style`: `style`
      - Border settings: `outer`, `inner`, `whole`: (`stroke`, `radius`, `outset`, `fill`, `inset`,)
    - Experimental: `item-format`
- 🆕 Resume enum [`issue#1`](https://github.com/tianyi-smile/itemize/issues/1)
- List enhancements
  - 🆕 Terms-like functionality: `item`
  - 🆕 Checklist (from the `cheq` package, with minor enhancements and fixes for compatibility with `itemize`)
    - Configure using `config.checklist`
- Add manual
- Fixes
  - 🆕 Block-level elements displayed on the same line ([`issue#4`](https://github.com/tianyi-smile/itemize/issues/4))
    - Now better handles blank content in typst, ensuring correct alignment
    - Due to limitations in typst, content within `context` cannot be retrieved in pure typst, making it impossible to correctly process
    - Similarly, output content in `ref(...)` cannot be processed
    - For block-level elements, our handling is more refined compared to native behavior:
      - If a block-level element behaves like a paragraph (`par`, `pad`, `block`, `repeat`, `layout`), it is treated as paragraph behavior, ensuring alignment of labels and bodies within the same paragraph
      - If it is not a paragraph (`block-equation`, `block-raw` `rect`, `table`, `grid`, `stack`, `heading`, `figure`, etc.), it follows native behavior
  - Resolved some cases where "layout did not converge within 5 attempts" occurred
  - Other fixes:
    - Support for `enum.start`
    - ⚠️ *Breaking change*: For compatibility with the native `list.marker` behavior, unlike the methods provided by `itemize`, the level here starts from 0 (see [`issue#3`](https://github.com/tianyi-smile/itemize/issues/3)).
    - [X]  Fixed: Incorrect label display when mixing enums and lists with `*-enum` and `*-list` configurations
      - Now, we rewrite both behaviors of `enum` and `list`, but `*-enum`does not configure list formatting, and similarly for`*-list`
    - [X]  Compatibility with typst 0.14 behavior (`[#6609]` and `[#6242]`)

### DIFF
--- a/packages/preview/itemize/0.2.0/README.md
+++ b/packages/preview/itemize/0.2.0/README.md
@@ -602,8 +602,8 @@ The parameters for `default-enum-list` and `paragraph-enum-list` are similar.
   - `auto-base-level: bool = false`: To maintain compatibility with native behavior, the display of `numbering` and `marker` still uses absolute levels. This means even if you reconfigure `enum.numbering` and `list.marker` in sublists, the display of `numbering` and `marker` in sublists follows the absolute level rules. Default: `false`. 
     - If `auto-base-level` is set to `true`, then it treats the current level as 1.
     - ⚠️ *Breaking change*: When configuring enums and lists using `*-enum-list`, `*-enum`, or `*-list`, the current level is treated as 1.
-  - `checklist: array | bool = false`: Enables checklist. 
-  - `auto-resuming: auto | bool | none = none`: Relate to the feature `Resuming Enum`. Alternatively, You can also enable and configure checklist-related features using the method `config.checklist`.
+  - `checklist: array | bool = false`: Enables checklist. Alternatively, You can also enable and configure checklist-related features using the method `config.checklist`.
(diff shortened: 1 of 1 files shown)


saecki:
Thanks for this update!

PR #3122, opened 2025-09-23
physica:0.9.6

Leedehai:
I am submitting
- [ ] a new package
- [x] an update for a package

* Remove differential's smart spacing -- though this space is recommended by TeXBook, it creates a bug https://github.com/Leedehai/typst-physics/issues/63.
* *Adding rising and falling edge support (#62)
* Allow passing multiple conditions to Set by using the array syntax (#59)

PR #3312, opened 2025-10-25
physica:0.9.7

Leedehai:
I am submitting
- [ ] a new package
- [x] an update for a package

* **(breaking)** In derivative `dv` and partial derivative `pdv`, rename argument `s` to `style`. For the original argument value `s:\/`, write `style:"horizontal"` instead, just like what you would write in [`math.frac`](https://staging.typst.app/docs/reference/math/frac/) starting from Typst 0.14.0.
* **(breaking)** In differential `dd`, rename argument `p` to `prod`.
* Fix isotope so that it can also be rendered in text mode
* Make hbar based on `U+210e` directly instead of on `sym.planck`, because since Typst 0.14.0 `sym.planck` is the reduced Planck constant, taking the place of the deprecated `sym.planck.reduced`
* Replace `angle.r` and `angle.l` with `chevron.r` and `chevron.l`, as required by the deprecation notice of Typst 0.14.0.

### DIFF
--- a/packages/preview/physica/0.9.7/physica.typ
+++ b/packages/preview/physica/0.9.7/physica.typ
@@ -143,6 +143,8 @@
 
 // == Braces
 
+// Use a semicolon to delimit the expression and condition,
+// e.g. Set(a_n), Set(a_i; forall i), Set(vec(1,n); forall n, n|2)
 #let Set(..args) = {
   let expr = args.pos().at(0, default: none)
   let cond = args.pos().at(1, default: none)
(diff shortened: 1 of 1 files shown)


Leedehai:
Hi - the failed check are all warnings generated by https://github.com/typst/package-check, and I confirm they are actually intended (like [before](https://github.com/typst/packages/pull/3122#issuecomment-3326272915)). If I can, I wish to suppress those warnings with in-file annotations, like `@typst-package-check:disable-next-line:kebab-case`.

elegaanz:
Thanks for the reminder. Don't worry too much about it, ultimately packages are reviewed by humans and we can tell that kebab-case doesn't make sense in this context. The warnings reported by CI never prevent a package from being merged, they are here to bring our attention to some potential issues. I want to provide some way to turn them off, but I don't know if it should done with in-file annotations, I still need to think about it.

PR #3319, opened 2025-10-25
october:1.0.1

extua:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

Small changes to the design of the calendar to set margin and text size in main document; remove the border around month headings and align them to the left. Mostly inspired by @sermuns.


elegaanz:
Thank you.

PR #3320, opened 2025-10-25
ilm:1.4.2

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

### Changes
- Changed `author` option from string to content type.


elegaanz:
`document.author` can only be a string or an array of strings, passing content there would break the compilation. You should convert it back to a string.

talal:
My bad, thanks for the review.

I've reverted the changes in my repo and I'm closing this PR.

PR #3329, opened 2025-10-26
numty:0.1.0

PabloRuizCuevas:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added Matrix determinant for nxn matrix
- Added test and updated documentation


packages/preview/numty/0.1.0/README.md
```diff
@@ -0,0 +1,139 @@
+## Numty
+
+### Numeric Typst
+
+A library for performing mathematical operations on n-dimensional matrices, vectors/arrays, and numbers in Typst, with support for broadcasting and handling NaN values. Numty’s broadcasting rules and API are inspired by NumPy.
+
+```typ
+#import "numty.typ" as nt
```
elegaanz: It would be better to show how to import the package from Universe, with `"@preview/numty:0.1.0"`.

packages/preview/numty/0.1.0/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+Numty uses standard typst list as a base type, most 1d operations like dot are suported directly for them.
+
+For matrix specific operations we use 2d arrays / nested arrays, that are also the standard typst list, but nested like in: ((1,2), (1,1)). 
+
+For convenience you can create column or row vectors with the #nt.c and #nt.r functions.
+
+```typ
+#import "numty.typ" as nt
+#import "numty.typ": c, r
```
elegaanz: Same here.

packages/preview/numty/0.1.0/test.typ
elegaanz, line 1: I'm not sure it is relevant to copy this file here, it seems to be only useful for development purpose.

packages/preview/numty/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "numty"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Pablo Ruiz Cuevas"]
+license = "MIT"
+description = "Numeric Typst"
```
elegaanz: I think the description could be more complete. Rephrasing your README, it could say something along the lines of "Perform mathematical operations on n-dimensional matrices, vectors/arrays, and numbers".

elegaanz:
Hey, just a quick reminder that there are some small things to fix before this version can get published. Do you think you will have some time in the coming days to finalize that?

PabloRuizCuevas:
Yes, will fix them soon, didn’t see the post before, thanks for the reminder.On Mon, 1 Dec 2568 BE at 23:56 Ana Gelez ***@***.***> wrote:elegaanz left a comment (typst/packages#3329)
Hey, just a quick reminder that there are some small things to fix before this version can get published. Do you think you will have some time in the coming days to finalize that?

packages/preview/numty/0.1.0/test.typ
PabloRuizCuevas: I exclude it in the toml

packages/preview/numty/0.1.0/typst.toml
PabloRuizCuevas: done

packages/preview/numty/0.1.0/README.md
PabloRuizCuevas: ok

packages/preview/numty/0.1.0/README.md
PabloRuizCuevas: ok

PabloRuizCuevas:
Should be it i think test.typ is already excluded in the toml, so changing the path will break it.

PabloRuizCuevas:
Please don't merge yet, I need to fix something more

elegaanz:
No worries, take all the time you need.

> Should be it i think test.typ is already excluded in the toml, so changing the path will break it.

Excluding it will only remove it from the final archive that users download when importing the package, the file will still be committed to this repository. We try to not commit development files here, to keep the repository as small as possible. Even we can agree that your test file is not the largest, I would prefer to always apply this rule. If we start making exceptions, we would have to judge where the limit between "small enough" and "too large to be acceptable" is.

The `exclude` key in your manifest could be edited to not contain this path anymore, or kept as is: if an excluded file is not found, the bundler will proceed without errors., considering it is already absent.

PabloRuizCuevas:
Ok I deleted the test file and take out from the exclude , should be ready to merge

elegaanz:
Thanks!

PR #3333, opened 2025-10-27
tdtr:0.2.0

Vertsineu:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- add some pre-defined drawing functions, e.g. `circle-draw-node`
- support edge labels while drawing a tree, using enum list item as edge label
- update README for better readability


saecki:
Thanks for the update!

PR #3337, opened 2025-10-27
rexllent:0.4.0

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: `rexllent 0.4.0` replaces the slow `rquickjs` runtime with a pure Rust implementation for number formatting, which would result in a boost in performance!


saecki:
Thanks for this update!

PR #3351, opened 2025-10-28
formalettre:0.3.0

Brndan:
I am submitting
- [ ] a new package
- [x] an update for a package

What's new : 
- convert to kebab-case
- API improvements
- page numbering
- various bug fixes


saecki (commented):
Thanks for this update! This PR also includes the `0.2.0` version, could you rebase onto `main` so only the `0.3.0` version is included? You may need to [sync your fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork)

PR #3356, opened 2025-10-29
formalettre:0.3.0

Brndan:
I am submitting
- [ ] a new package
- [x] an update for a package

What's new :

- convert to kebab-case
- API improvements
- page numbering
- various bug fixes


elegaanz:
Thank you.

PR #3358, opened 2025-10-29
modern-wku-thesis:0.1.2

timmycheng:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:  Modify the format by the requirements of the professor, 1) Add pagebreak before every level 1 heading; 2) center-align level 1 heading.


elegaanz:
Thank you.

PR #3366, opened 2025-10-30
larrow:1.0.0

Mambouna:
I am submitting
- [x] a new package

Description: The package allows easily drawing and styling arrows between labels in the document. It's useful because there is currently no easy in-built way to do so. Base typst does not have arrows and cetz (which larrow uses) is non-trivial for putting drawings into the flow of content. The package resolves those.

### DIFF
--- a/packages/preview/larrow/1.0.0/typst.toml
+++ b/packages/preview/larrow/1.0.0/typst.toml
@@ -7,5 +7,5 @@ license = "MPL-2.0"
 description = "Draw and style arrows between labels."
 repository = "https://github.com/Mambouna/larrow"
 keywords = ["arrow", "arrows", "label", "labels", "connect", "connection"]
-categories = ["Components", "Visualization", "Utility"]
+categories = ["components", "visualization", "utility"]
 exclude = ["example"]


elegaanz:
Can you please move your `1.0.0` folder to the `packages/preview/larrow` directory, and revert the deletion of the license, README and .gitignore files in the root?

Mambouna:
> Can you please move your `1.0.0` folder to the `packages/preview/larrow` directory, and revert the deletion of the license, README and .gitignore files in the root?

Apologies, maybe my understanding of a sparse checkout is insufficient. From the documentation I thought the root of my git checkout is the root of the directory given in the sparse checkout command. Is that incorrect?

Just to be clear now. The files I deleted were from the top level of the entire git repo? And the root of my checkout is still the root of the entire repo, just without all the files to go along with it? So now I create the entire directory tree to my package as seen from the root of the repo and revert the deletion of the top level files. Then everything should be correct?

Thanks for your patience.

elegaanz:
Yes, you are correct. This is indeed not very clear in the docs and could be improved.

Mambouna:
> Yes, you are correct. This is indeed not very clear in the docs and could be improved.

Thank you for the advice, I've reverted the changes and fixed the toml after a failed test.

elegaanz:
Looks good, thank you for this package :)

PR #3370, opened 2025-10-30
pollux:0.1.0

taka255:
I am submitting
- [x] a new package
- [ ] an update for a package

A clean and minimal academic poster template inspired by gemini, a popular LaTeX poster design.


packages/preview/gemini-like-poster/0.1.0/typst.toml
```diff
@@ -0,0 +1,16 @@
...
+entrypoint = "package.typ"
+authors = ["Kaito Takanami"]
+license = "MIT"
+description = "A clean and minimal academic poster template inspired by Gemini."
+keywords = ["poster", "template", "academic"]
+categories = ["poster"]
+repository = "https://github.com/taka255/gemini-like-poster"
+
+[template]
+path = "."
```
elegaanz: Your template should go in its own subdirectory. Only files in this directory will be copied to the user's project when they start from your template.

packages/preview/gemini-like-poster/0.1.0/typst.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "gemini-like-poster"
```
elegaanz: I'm not entirely sure about that name. Technically I would say it fits our guidelines (even if a bit descriptive), but it is a bit weird to name it in reference to the equivalent LaTeX package. For example, `polylux` is not called `beamer-like`. If you have a better idea for a name, I think it would be nice to change it. You can still explain that it was inspired by the gemini LaTeX template in your README, and add it to the keywords for better SEO.

packages/preview/gemini-like-poster/0.1.0/typst.toml
taka255: Thanks! I’ve moved the template files into their own subdirectory and updated the path in typst.toml.

packages/preview/gemini-like-poster/0.1.0/typst.toml
taka255: Thanks for the suggestion! I’ve renamed the package to pollux and mentioned in the README that it was inspired by the Gemini LaTeX template, as you suggested.

packages/preview/pollux/0.1.0/package.pdf
elegaanz, line 1: I think this file was added by mistake, can you delete it please?

elegaanz (commented):
Looks good now, just one small thing and I will merge.

packages/preview/pollux/0.1.0/package.pdf
taka255: Thanks for pointing that out. I’ve deleted the file now.

elegaanz:
Thanks!

PR #3382, opened 2025-10-31
diatypst:0.8.0

skriptum:
I am submitting
- [ ] a new package
- [x] an update for a package

version 0.8.0 introduces a new counter, that displays the number of pages per section and not the whole document, improves upon visibility for long presentations


saecki:
Thanks!

PR #3383, opened 2025-10-31
pc-letter:0.3.0

thatfloflo:
I am submitting
- [ ] a new package
- [x] an update for a package

#### Added

- New options to configure the letterhead:
  - The `alignment.letterhead` style-option allows the letterhead to be aligned to the `left`, `center`, or `right`. The default remains `center`.
  - The `components.letterhead.ascent` style-option allows the spacing between the letterhead and the content below (typically the address field) to be overwritten, which might help with better accommodating slightly larger logos, if used.
  - `pc-letter.init` now accepts a `logo` argument which can be used to add a logo to the letterhead.
- New example documents:
  - examples/example-logo.typ shows how to use a logo in the letterhead.
  - examples/example-signature.typ shows how to use a image as a signature.

#### Changed

- The readme now contains a "How do I ...?" section with examples for things users might want to know how to do without reading all the configuration options. More items can be added to this in the future.

#### Fixed

- A typo in the French letter example in examples/example-fr.typ.


saecki:
Thanks for the update!

PR #3398, opened 2025-11-03
identified:0.1.0

Andrew15-5:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: [Identified](https://codeberg.org/Andrew15-5/identified) (IDEntiFied) is a package that allows creating IDEF (IDEF0) diagrams in seconds.


Andrew15-5:
The code blocks should have something like [`#[should_panic]`](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic), to remove the errors. Otherwise, you cannot sketch anything.

elegaanz:
Nice package, thank you.

> The code blocks should have something like [#[should_panic]](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic), to remove the errors. Otherwise, you cannot sketch anything.

Yeah, I agree these errors are a bit aggressive in practice, but I would like to find a way to ignore them when relevant without people having to modify their README just to get a green checkmark. The legit uses cases I've seen so far of invalid syntax are basically the ones you are using here: `let` without a body and `…` as placeholders, so maybe making an exception for these cases could work. In general, the CI linter is here to bring attention to some potential issues, and a human should always judge whether the warning is actually relevant given the context.

Andrew15-5:
Yeah, I also think that maybe editing readme for this is not convenient. I definitely don't want to change the info string, because that will break syntax highlighting. Using special comments also feels a bit janky. So if parser indeed can exclude special cases, then that would be great.

Especially for beginners, it'll probably be a bit intimidating or worrying to see errors in something that works and was tested. There is a feeling of it not being merged or something.

If the message had a clarification that those errors are not necessarily bad, this would help.

PR #3399, opened 2025-11-04
Deploy: don't purge the CDN endpoint

elegaanz:
It is not necessary anymore, as the archives are now fully immutable, and the correct cache headers are defined for each uploaded file.

It was causing the CI to fail these last days because Azure Front Door configuration changes are temporarily forbidden, following the recent outage.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -46,17 +46,3 @@ jobs:
           azcopy_v10 cp --recursive ./dist/* "https://$ACCOUNT_NAME.blob.core.windows.net/\$web" \
             --overwrite true --include-pattern '*.json' \
             --cache-control 'must-revalidate' --put-md5
-
-      - name: Log in to Azure
-        uses: azure/login@v2
-        with:
-          creds: ${{ secrets.AZURE_PACKAGE_CREDENTIALS }}
-
-      - name: Purge CDN endpoint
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -46,17 +46,3 @@ jobs:
...
-        uses: azure/login@v2
-        with:
-          creds: ${{ secrets.AZURE_PACKAGE_CREDENTIALS }}
-
-      - name: Purge CDN endpoint
-        uses: azure/CLI@v2
-        with:
-          inlineScript: az afd endpoint purge -g Landing-Page --profile-name LandingPage --content-paths '/*' --endpoint-name 'packages'
-
-      - name: logout
```
elegaanz, line 60: I'm not sure what was the reason for explicit log out, so I can't really tell if it is still necessary, but installing `az` just to be able to run `az logout` seemed a bit useless, so I removed this step too. As far as I can tell, `azcopy` gets its credentials from environment variables, and I don't think `az logout` clears that. I don't really think there is a security risk with having these variables set, as this job can only run on the main branch anyways.

reknih: `azcopy` gets its credentials on line 36 and stores them in the OS trust store if I understood this right so this is fine

reknih (approved):
LGTM

PR #3401, opened 2025-11-04
Improve sparse-checkout documentation

### LINKED ISSUE #2588: Update sparse checkout instructions to use a fork
The tips include [instructions on how to do a sparse checkout](https://github.com/typst/packages/blob/main/docs/tips.md#sparse-checkout-of-the-repository). However this is how to do a checkout of `typst/packages` as the `origin` remote. Unless I'm mistaken, that's actually not a useful thing to do. What you typically want to do as a contributor is _fork_ the repo, clone that sparsely (typically as `origin`), and also add `typst/packages` as another origin (typically `upstream`). If you do forget to set _both_ of those remotes as sparse checkouts, then when you push a commit, it will download everything. So the sparse thing won't work.

I propose changing the sparse checkout code snippet to:

```
git clone --depth 1 --no-checkout --filter="tree:0" git@github.com:{your-github-username}/packages
cd packages
git sparse-checkout init
git sparse-checkout set packages/preview/{your-package-name}
git remote add upstream git@github.com:typst/packages
git config remote.upstream.partialclonefilter tree:0
git sparse-checkout reapply
```

(I'm not sure if all those steps are necessary.)

elegaanz:
Fixes #2588

### DIFF
--- a/docs/tips.md
+++ b/docs/tips.md
@@ -8,15 +8,25 @@ only work in a few specific directories (the ones for your own packages). Git
 allows for "sparse checkouts", which reduce the time it takes to clone the
 repository and its size on your disk.
 
-Follow these steps to clone the repository:
+First, make sure you have forked the repository to your own GitHub account.
+Then, follow these steps to clone the repository:
 
 ```sh
-git clone --depth 1 --no-checkout --filter="tree:0" git@github.com:typst/packages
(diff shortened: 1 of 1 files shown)


Mambouna:
I think this makes it a lot clearer, yes!

PR #3405, opened 2025-11-05
lucide:0.1.0

Lieunoir:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Add lucide:0.1.0: a library for [Lucide](https://lucide.dev/) icons

### DIFF
--- a/packages/preview/lucide/0.1.0/lib.typ
+++ b/packages/preview/lucide/0.1.0/lib.typ
@@ -11,7 +11,6 @@
 /// Parameters:
 /// - `name`: Name of the icon
 /// - `..args`: Additional arguments passed to the `text` function
-/// - `fill`: Fill color, none by default. Officially not supported (see https://lucide.dev/guide/advanced/filled-icons)
 ///
 /// Returns: lucide icon as text element
 #let lucide-icon(name, ..args) = {


elegaanz:
I'm generally a bit worried about these kind of packages, embedding a whole icon set as SVG. We try to limit the size of packages, both to keep this repository small, and to provide users with archives that download in reasonable time. Here I think it should be fine, since there are not too many icons and their SVG definitions don't seem too large from what I've seen. But ideally, if there is a font version of the icon set, I think it would be preferable to ask users to install the TTF/OTF file and only provide convenient aliases to glyphs (this is what is usually done by other icon sets). If it is not possible, I think we can make an exception here, given the relatively small size of the icon set.

Lieunoir:
I've changed it so that it relies on the ttf file, which is indeed smaller (and also generally simplier). There is one downside to this approach though, the "stroke width" parameter in the original icons (see the slider here https://lucide.dev/icons/) doesn't seemed mapped to font weight. But that's probably an issue on the font size, and not very important

elegaanz:
Yeah, I think the issue should be fixed on Lucide's side, there is not much Typst can do about it I think. Maybe you could try to emulate it by setting the stroke width on the text element, but the result would probably not look that good. I'll merge this as is for now, and if they provide a way to configure the stroke width in the future, you can publish a new version then.

LukasKalbertodt:
I'd like to suggest switching back to the original implementation of embedding the SVG data directly into the package. It is much more convenient than also having to download a random font file from somewhere. Further, it is more reproducible, as the same package version will always result in the same output. Currently, it depends on the font file which, yes, should be checked into the project, but that's not a requirement.

Regarding the size: I don't think this is a problem at all. The `lib-map.typ` file from the initial commit is 344KB large, but inside `.tar.gz` that's only 83KB. That is way smaller than many other packages in this registry. It is also way smaller than the Lucide font file (TTF: 817KB, woff2: 267KB).

If the Typst team agrees, I'll gladly help update the package, if @Lieunoir is fine with that.

PR #3433, opened 2025-11-10
problemst: fix the path to thumbnail.png

ada4a:
Note: The path seems to be correct in the original repo


saecki:
We have a policy that doesn't allow package edits after they have been published. If you want to contribute to packages, I'd suggest opening a PR on the package repository, then a new version can be published.

This link also already seems to be corrected in `0.1.2`.

PR #3436, opened 2025-11-11
peano:0.2.1

vanleefxp:
I am submitting
- [ ] a new package
- [x] an update for a package

- `peano.number`
  - Support infinity and NaN values for `mp.integer`.
  - make zero value in `mp.integer` signed.
  - Support radix parsing for `mp.integer`.
  - Extended arithmetics in `mp.integer` to support infinity and NaN values. Now division by zero in `mp.integer` won't cause an error.
  - Fixed the bug that `mp.rational.pow` for negative zero and negative infinity returns the wrong sign.
  - Added `floor` and `ceil` for `mp.rational`
  - Fixed the problem that negative values in `mp.integer` gets multiple minus signs.
  - Renamed named argument `signed-infinity` to `signed-inf` for `to-str` in `rational`, `mp.integer`, `mp.rational`
  - Added `cmp` and `to-float` for `rational`.
  - Added named argument `display` for `rational.to-math` and `mp.rational.to-math`


saecki:
Thanks for the update!

PR #3450, opened 2025-11-12
free-cv:1.0.0

Freezy727:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: CV template useful for quickly drafting a CV.

### DIFF
--- a/packages/preview/free-cv/1.0.0/template/example.pdf
+++ b/packages/preview/free-cv/1.0.0/template/example.pdf
@@ -1,4911 +1,0 @@
-%PDF-1.7
-%€€€€
-
-1 0 obj
-<<
-  /Type /Pages
-  /Count 2
-  /Kids [340 0 R 342 0 R]
->>
-endobj
-
-2 0 obj
-<<
-  /Type /Outlines
-  /First 3 0 R
-  /Last 7 0 R
-  /Count 5
->>
-endobj
-
-3 0 obj
-<<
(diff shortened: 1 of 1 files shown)


Freezy727:
Hi, how do I correctly specify the exclude key in the toml file to ignore the pdf under the template directory? ``"template/example.pdf"`` didn't seem to work. (Also in the section on [what to exclude](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude), it says that any file in the template directory is required, so this could maybe be clarified to not always be the case... Unless I missed somewhere that I wasn't supposed to put the pdf under the template directory in the first place 😅)

Also not sure why I'm failing the test on the readme being valid UTF-8 so if someone could explain the issue I'll try to fix it!

elegaanz:
Thank you for building this template. Can you please fix the CI errors?

Freezy727:
@elegaanz Fixed the readme encoding issue, but can't find how to exclude the pdf from being included in the package (``"template/example.pdf"`` did not work as an exclude key in the toml). Can you point me to how to fix this, or is it necessary to outright delete the file from the repository?

elegaanz:
Yes, in this case you should just delete the file, otherwise it would get copied to user projects when using your template, which is not desirable, especially in the web app where no PDF is produced unless asked.

elegaanz:
Thanks.

PR #3452, opened 2025-11-13
curryst:0.6.0

pauladam94:
I am submitting
- [ ] a new package
- [x] an update for a package

## What have changes
- line width adapt now to text size
- premise and conclusion ordering is now more logical to what is display on the document. This is a breaking change.
- add the `rule-set` to layout multiple rules


MDLC01:
Yes this can be merged.

elegaanz:
Thanks.

PR #3462, opened 2025-11-15
cmarker:0.1.7

SabrinaJewson:
I am submitting
- [ ] a new package
- [x] an update for a package

Additions
- Add a `heading-labels` option to configure how heading labels are
  automatically generated. In particular, one can match Jupyter’s
  algorithm (instead of GitHub’s) for better interopability.
- Add `render-with-metadata` to allow reading YAML frontmatter.
- Support negative `h1-level` values to allow shifting headings
  backward.
- Convert level 0 headings into Typst titles using the `#title`
  function.
- Support inline SVG by default. This is converted into a Typst `image`
  call.

Changes
- Update to Typst 0.14.0
- Change the behaviour of raw text and escapable raw text HTML elements
  to attempt to consume as much text as possible before reaching the end
  tag, instead of eagerly auto-closing.

Fixes
- Change the footnote implementation to include them inline. This allows
  footnotes to appear on the correct page in some edge cases.
- Support the full range of labels allowed by Typst.
- Use the correct GitHub slugification algorithm.
- Add more type validation.

### DIFF
--- a/packages/preview/cmarker/0.1.7/README.md
+++ b/packages/preview/cmarker/0.1.7/README.md
@@ -100,7 +100,7 @@ The parameters are as follows:
 	For example, to render math equation as a Typst math block,
 	one can use:
 	```typst
-	#import "@preview/mitex:0.2.4": mitex
+	#import "@preview/mitex:0.2.6": mitex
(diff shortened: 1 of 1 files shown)


elegaanz:
Thanks!

PR #3463, opened 2025-11-15
decasify:0.11.0 and decasify:0.11.1

alerque:
I am submitting
- [x] an update for a package

Description: Upstream release completely refactors the core case conversion library interface. Instead of infallibly converting anything you throw at it and using default values for language and other settings if they could not be parsed properly, the conversion is now fallible and raises relevant errors if it can't parse the input.

This update changes the version following [ZeroVer](https://0ver.org/) bumps the 'minor' version segment of the SemVer tag since the API changed in a breaking way.

The Typst wrapper should function with no changes from users prospectives since the Typst code is also adjusted to properly consider the underlying changes to error handling, and any resulting errors should pass up the error as `panic!` with appropriate flattened text error messages that make it across the WASM boundary instead of getting lost in a trap like in previous releases.


elegaanz:
Looks good, thank you!

PR #3472, opened 2025-11-16
docs: Explain in greater depth what content packages should include

har7an:
with examples and cross-references to particular definitions. This should help to prevent misunderstandings with the connection between the `package.exclude` key in the manifest and this repository.

After publishing my last package I had to learn the hard way that I misunderstood parts of the "What to commit? What to exclude?" section. Now that I know what I did wrong, the documentation starts making sense. However, it wasn't clear to me before and I had read that section 3 or 4 times.

Here's a proposal for a more elaborate description of the connection between a packages original repository, the contents of a PR published here, and what the users finally downloads to their machine when compiling documents. I haven't yet embedded this into the existing documentation because I wanted to get feedback on whether you think this is helpful first.

Thank you in advance!

### DIFF
--- a/docs/tips.md
+++ b/docs/tips.md
@@ -39,53 +39,8 @@ Git will replace the copied files with a submodule.
 
 ## What to commit? What to exclude?
 
-In some case, simply copying and pasting the contents of the repository in which
-you developed your package to a new directory in this repository is enough to
-publish a package. However, this naive approach may result in unnecessary files
-being included, making the size of this repository and of the final archives
-larger than they need to be.
-
-There are two solutions to limit this problem: excluding files from the archive
-(using the `exclude` key in your [package manifest][manifest]), or simply not
-committing the files to this repository in the first place.
-
-To know which strategy to apply to each file, we can split them in three groups:
-
-__1. Required files__\
-Files that are necessary for the package to work. If any of these files are
-removed, the package would break for the end user. This includes the manifest
-file, main Typst file and its dependencies, and in case of a template package,
-any file in the template directory.
-
-__2. Documentation files__\
-Files that are necessary for the package to be displayed correctly on Typst
-Universe. This includes the README, and any files that are linked from there
-(manuals, examples, illustrations, etc.). These files can easily be accessed
-by opening the package README.
-
-__3. Other files__\
-This generally includes test files, build scripts, but also examples or manuals
-that are not linked in the README. These files would be almost impossible to
-access for the final user, unless they browse this GitHub repository or their
-local package cache.
-
-The first two groups (required and documentation files) should be committed to
-this repository. And files that are not strictly necessary for the package to
-work (documentation files) should be excluded in `typst.toml`. They will still
-be available on typst universe to link to from the README.\
-The third group should simply not be committed to this repository. If you think
-some of the remaining files are important, they probably belong to the second
-group and should be linked in the README, so that they are easily discoverable.
-A good example showing how to link examples and a manual is [CeTZ][cetz].
-
-The only exceptions to this rule are the LICENSE file (that should always be
-available along with the source code, so it should not be excluded), and the
-README (which is generally a lightweight file, and can provide minimal
-documentation in case the user is offline or can't access anything else than
-their local package cache for some other reason).
-
-Also note that there is no need to exclude template thumbnails: they are
-automatically left out of the archive.
+Please refer to the standalone documentation on the [Contents of a
+package][contents].
 
 ## Tools that can be useful
 
@@ -97,6 +52,7 @@ The community created some tools that can help when developing your package:
 - [showman], to help you document and publish your package.
 
 [cetz]: https://typst.app/universe/package/cetz/0.3.4
+[contents]: contents.md
 [typst-package-check]: https://github.com/typst/package-check
 [tytanic]: https://typst-community.github.io/tytanic/
 [typship]: https://github.com/sjfhsjfh/typship


docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+1. [Published package version][c]
+1. [Imported package version][d]
+
+[a]: #categories-of-files
+[files_required]: #required-files
+[files_docs]: #documentation-files
+[files_other]: #other-files
+[b]: #package-development-version-stage-1
+[c]: #published-package-version-stage-2
+[d]: #imported-package-version-stage-3
```
elegaanz: We don't really have ToC in other documentation files, and I don't think this file is long enough to deserve one. Can you remove it, please?

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Categories of files
+
+The files possibly contained in your package are broadly grouped into the
+following three categories.
+
+
+### Required files
+
+Files that are necessary for the package to work. If any of these files are
```
elegaanz: I would have written a full sentence here, something along the lines of "These files are necessary for the package to work".

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+- `LICENSE`
+- `typst.toml`
+- `src/lib.typ` (see `package.entrypoint` in `typst.toml`)
+- `my-plugin.wasm` (for WASM plugins)
+- Images/icons/other assets if your package reproduces these in documents
+
+
+### Documentation files
+
+Files that are necessary for the package to be displayed correctly on [Typst
```
elegaanz: Same here.

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+  anything else than their local package cache for some other reason
+
+*Examples*:
+
+- `README.md`
+- `LICENSE`
+- `typst.toml`
+- `src/lib.typ` (see `package.entrypoint` in `typst.toml`)
+- `my-plugin.wasm` (for WASM plugins)
+- Images/icons/other assets if your package reproduces these in documents
```
elegaanz: "images, icons and other assets…"

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+of a template package, any file in the template directory. On top of this, the
+following files are also included:
+
+- The LICENSE file (e.g. `LICENSE` or `COPYING`): This should always be
+  available along with the source code
+- The README file (e.g. `README.md`): This is generally a lightweight file, and
+  can provide minimal documentation in case the user is offline or can't access
+  anything else than their local package cache for some other reason
+
+*Examples*:
```
elegaanz: I would remove the emphasis here (and in other sections as well).

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+- `examples/try-this.typ` (if your README links to that)
+
+
+### Other files
+
+This generally includes test files, build scripts, but also examples or manuals
+that are not linked in the README.
+
+These files would be almost impossible to access for the final user, unless
+they browse [this GitHub repository][this_repo] or their local package cache.
```
elegaanz, line 72: This is not correct, since these files should not be included in the final archive and thus should never be stored in the cache. Can you remove the end of the sentence, starting at "or"?

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+
+*Examples*:
+
+- `.github`
+- `CHANGELOG.md`
+- `Makefile`
+- `tests/`
+
+
+## Package development version (Stage 1)
```
elegaanz: I think "Stage 1: Package Development" sounds better.

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+
+This is the state of the package that you work on. Your package is most likely
+published to some git forge (GitHub, GitLab, ...) but it doesn't have to be.
```
elegaanz: This is really nitpicking, but can you replace `...` with a proper ellipsis (…), please?

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+│   ├── a-subfile.typ
+│   └── lib.typ
+├── test
+│   ├── ...
+│   └── utils.typ
+└── typst.toml
+```
+
+
+## Published package version (Stage 2)
```
elegaanz: "Stage 2: Published Package"

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+├── LICENSE
+├── README.md
+├── src
+│   ├── a-subfile.typ
+│   └── lib.typ
+└── typst.toml
+```
+
+
+## Imported package version (Stage 3)
```
elegaanz: "Stage 3: Imported Archive" or "Downloaded Archive"

docs/contents.md
```diff
@@ -0,0 +1,208 @@
...
+
+This is the state of the package that users get when they `#import` your code.
```
elegaanz: Maybe you should be explicit here and say that these files will be downloaded and stored in the local package cache?

elegaanz (commented):
Thanks for writing this, I think it is an improvement to be more explicit. I left a few comments, mostly nitpicking, but I think it would improve what you wrote a bit and make it more consistent with other documentation files.

I think it would be good to link to this file in the relevant section(s) of the existing docs, so that it is easier to discover for people who may need this extra information.

docs/contents.md
har7an: I've rephrased the sentence, do you think it's better now?

docs/contents.md
har7an: Well, in a way I think it is, basically, correct. It's just that we don't want this to be the case, right?

If I were to copy my bare package contents (as taken from my Git forge of choice) into this repository and somehow it passed review, the files would in fact be here. And they would show up in the archive that users download (assuming i haven't configured `package.exclude`). It's just that due to how Typst Universe and the actual package ecosystem (using `#import`) interact with the code I upload here, those files will be invisible to most users. If you insist I can still remove it, but I think as it is (and I took that verbatim from the previous docs in `tips.md`) this helps users understand *why* we don't want those files to be here.

YDX-2147483647:
Perhaps the current version is so detailed that readers can easily get lost.
For example, there are long paragraphs between file trees, so it requires scrolling back and forth to see what files are dropped.
I agree that exploring the details is better than omitting them (especially nowadays when LLMs are widely available), but I suggest conveying a rough core conclusion first.
Maybe adding a diagram like this at the beginning would suffice.

Frame 1


_Disclaimer:_ The above image is only for demonstrating the idea. I'm not suggesting using black & white or specific fonts.

har7an:
> I agree that exploring the details is better than omitting them (especially nowadays when LLMs are widely available)

Rest assured, neither the existence nor the general availability of LLMs influences the way I write docs (so far at least).

>  Maybe adding a diagram like this at the beginning would suffice.

Good call, I'd actually like to do something like that. But for the start I'd like to stick with the textual description since that just expands on what was there before. I feel adding an image will open a lot of potential for discussion as you hinted, so I wanted to do that in a subsequent PR. :)

docs/contents.md
har7an: @elegaanz Do you have any objections or is the text good to go?

docs/contents.md
saecki: Maybe this could be clarified by starting the sentence with:
`Even if included, these files ...`

elegaanz:
Hello, sorry for the long delay without updates. After discussing these changes a bit with the team and proofreading them again, I feel like these instructions being in their own file and the length of it would make most people skip it. For docs to be useful, people have to read them and a very detailed document like this one may be discouraging for newcomers. For this reason, I prefer to close this PR.

If you have more specific edits that could be made to the current text that would have helped you understand, we would be happy to discuss that with you. For example, one potential change to the current docs that we discussed is to change the three file category names to:

```
__1. Required files (commit & include)__\
__2. Documentation files (commit & exclude)__\
__3. Other files (don't commit)__\
```

Which would already clarify what you are supposed to do with each of these files, without making this section any longer.

We still think that what you wrote is valuable and can be useful to others, so it could be nice to publish somewhere else, for example on a blog if you have one.

Sorry again to have you wait for so long, and for asking for changes in my previous review without first really pondering whether or not we would like to integrate the changes. I hope that you understand this decision, and as I said above, we can continue to discuss how we can improve this part of the docs if you want, but with more focused edits.

PR #3481, opened 2025-11-17
bamdone-rebuttal:0.1.2

avonmoll:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Previously, there was a lot of vertical space around `quote`s, so I reduced it to `1em` in the template. Also, this has a PR which fixed the handling of the paper size.


elegaanz:
Thank you.

PR #3482, opened 2025-11-17
abiding-ifacconf:0.2.1

avonmoll:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: allowed for listing multiple affiliations for authors


saecki:
Thanks for the update!

PR #3485, opened 2025-11-18
chordx:0.6.1

ljgago:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes in chordx 0.6.1:

- Measure the entire word instead of individual characters in single chord


saecki:
Thanks for the update!

PR #3496, opened 2025-11-20
Added oasis-align v0.3.1

jdpieck:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: fixed bug where content would be duplicated when using `oasis-algin-images()`


jdpieck:
The first three warnings are known, but I'm okay leaving them since they are useful on the GitHub. The last warning is a testing file, and does affect deployment. Perhaps the file should be excluded?

elegaanz:
Yes, any file that is only useful for development and that is not referenced in your README is basically impossible to discover from Universe (because we don't offer an easy way to browse all package files). It's best to just not copy it to this repository in that case.

jdpieck:
@elegaanz, I went ahead and removed all the demo/example files from `oasis-align`. Let me know if this causes any issues.

Thanks for the feedback!

elegaanz:
Can you revert the changes for previous versions of the package please? It is a bit unfortunate that these unnecessary files were published at the time, but now that the archives have been built we can't really modify them for reproducibility reasons. Demo files should only be removed for the new and future versions.

jdpieck:
@elegaanz apologies for the mishap. The commit has been reverted, and the package should be ready to update now. 

Thank you for the continued support!

elegaanz:
Thank you!

PR #3500, opened 2025-11-21
eggs:0.3.0

retroflexivity:
I am submitting
- [ ] a new package
- [x] an update for a package

- Support for separate numbering in footnotes (on by default)
- `ex-ref` now properly formats the ref instead of tinkering with regexes.


elegaanz:
Thank you.

PR #3520, opened 2025-11-24
aspirationally:0.1.1

breezykermo:
I am submitting
- [ ] a new package
- [x] an update for a package

Provides conditional setting via bullseye so that HTML export does not error

Currently aspirationally throws an error when you try to export to HTML. I intend to provide support for HTML styling eventually as the featuers in upstream Typst mature; but for now, we should just 'let the text through'.


elegaanz:
Can you update the two versions in imports, as reported in CI, please?

elegaanz:
Thank you :)

PR #3523, opened 2025-11-25
tiptoe:0.4.0

Mc-Zen:
I am submitting
- [x] an update for a package

- Added a `ring()` function for drawing ring arcs. 
- Fixed positioning of marks when `set rotate(reflow: true)` has been set. 
- The functions `line`, `curve`, `arc`, and `ring` now respond to set rules on `std.line` and `std.curve`. 

image


saecki:
Thanks for the update!

PR #3525, opened 2025-11-25
simple-ntnu-report:0.1.1

torsteinnh:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
This update to the template addresses several pain points identified over the last months, including:
- Disabling numbers on all headings for really short documents.
- Adding appendices after the sources.
- Improving formatting and listing of listings.
- Removing numbering form equations when desired.
- Aligning matrix and vector formatting with university norms.
- Adding list of useful packages that complement the template for most users.
- Updating third party packages that are used in the template.
- Translating a few more supplements.
- Including the unofficial ITK institute logo in the template.
The template should be fully backwards compatible.

### DIFF
--- a/packages/preview/simple-ntnu-report/0.1.1/README.md
+++ b/packages/preview/simple-ntnu-report/0.1.1/README.md
@@ -1,12 +1,14 @@
-# About
+# Unoficcial template for reports at NTNU
+
+## About
 This is a small template for writing reports tailored to the most common needs at NTNU in Norway, the goal is to cover everything from exercises to specialization projects.
 Note that many subjects have their own requirements for styling, this template is compatible with the requirements for TTK4101 in 2025 and 2026.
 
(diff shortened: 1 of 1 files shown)


packages/preview/simple-ntnu-report/0.1.1/README.md
```diff
@@ -0,0 +1,55 @@
+# About
```
saecki: Typst universe strips level one headings, since only one `<h1>` element is expected and that is considered the title of the page.
Your README should have one level one heading (the title), and all other ones should be level 2 headings. See: https://github.com/typst/packages/blob/main/docs/documentation.md#differences-from-standard-markdown

saecki (commented):
Thanks for the update! I've left a comment regarding the readme headings

PR #3528, opened 2025-11-25
tgm-hit-protocol:0.2.0

ensko:
I am submitting
- [ ] a new package
- [x] an update for a package

This is an official protocol template for the informatics department of the TGM technical high school in Vienna. This update makes the template compatible with Typst 0.13+ and requires 0.13 as the minimum version.


saecki:
Thanks for the update!

PR #3530, opened 2025-11-25
biz-report:0.2.0

garethwebber:
I am submitting
- [ ] a new package
- [x] an update for a package

Colour and Font are now stored in state when report created and passed to other components without the user needing to do it.
New dropcappara function.
Example uses IBM Sans rather than Arial as online tool doesn't have that font.

- [ X] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.


saecki:
Thanks for the update!

PR #3539, opened 2025-11-27
simple-ntnu-report:0.1.2

torsteinnh:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
Fixed a bug in the 0.1.1 release which slipped through the cracks, in which a long report without any appendix would get an extra blank page added to the end.


elegaanz:
Thank you.

PR #3540, opened 2025-11-27
biz-report:0.3.0

garethwebber:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Reduced package size by using svgs and shrinking one remaining png. Improved readme (describe components).

Sorry for three submissions in quick succession but this is done until I add new components.


elegaanz:
Thanks.

PR #3544, opened 2025-11-27
komet:0.2.0

Mc-Zen:
I am submitting
- [x] an update for a package

## Changelog

- Added `kde` function (thanks to @koole). 
- Fixed a bug with `histogram` where the last value was not binned. 

---
The package checker complains a bit about the pseudo Typst syntax that I use for rendering the function signatures but I think this is currently the cleanest way to annotate the types (and I would like to keep the Typst syntax highlighting).


elegaanz:
Yeah, don't worry too much about the README errors, they are mostly here to help us catch silly mistakes, but we can clearly see that it's intentional here.

PR #3546, opened 2025-11-27
ez-today:2.0.0

CarloSchafflik12:
I am submitting

- [ ] a new package
- [x] an update for a package

Description

- Fix: The cs language mode now uses lowercase month names
- Added cs2 language mode for Czech, which capitalizes month names with different declension
- Added support for English ordinal suffixes as text or exponents (e.g. 21st, 21<sup>st</sup>) with the `S` and `s` format options


elegaanz:
Thank you.

PR #3548, opened 2025-11-28
slydst:0.1.5

glambrechts:
I am submitting
- [x] an update for a package

Description: Added a feature (subslide numbering option) and fixed a bug (invisible slides when only a title).


elegaanz:
Thank you.

PR #3555, opened 2025-11-29
put-thesis:0.1.1

RoyalDonkey:
I am submitting
- [ ] a new package
- [x] an update for a package

- made sure the template works with typst 0.14.0 (compiler in manifest still kept at 0.13.0, as no changes were needed)
- updated dependency packages
- updated README 


elegaanz:
Thank you!

PR #3568, opened 2025-12-02
grayness:0.5.0

nineff:
I am submitting
- [ ] a new package
- [x] an update for a package

**Changed:**

- Updated to require Typst 0.14
- WebP Images are nolonger converted to PNG since Typst supports them natively

**Fixed:**

- SVG Filters now use sequential IDs and can therefore be chained together
- `image-mask()` now correctly applies the `args` parameters
- SVG Hue-Rotation nolonger incorrectly specifies the amount as "deg"


saecki:
Thanks for the update!

PR #3580, opened 2025-12-04
twilight-book:0.1.4

CrossDark:
I am submitting
- [x] a new package
- [ ] an update for a package

- [ ] added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys
- [ ] added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package
- [ ] have chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) and added a `LICENSE` file or linked one in my `README.md`
- [ ] tested my package locally on my system and it worked


CrossDark:
Must I use kebab-case names?

elegaanz:
> Must I use kebab-case names?

No this is only a common convention that we try to follow, so that it is consistent in the whole package ecosystem. If you prefer to keep using snake_case it is possible, especially since the few occurrences are for semi-private variables from what I can tell (it would still be nice to be consistent and use the same convention everywhere in your package, but it's not very important).

CrossDark:
I’ve deleted those files

elegaanz:
Okay, everything seems to be in order now. Thank you. Should I close your two other pending PR? If I'm not mistaken they are both for earlier versions of this package, so it doesn't really make sense to merge them too.

PR #3581, opened 2025-12-04
clean-hwr:0.1.5

Testspieler09:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Just bumping up the dependency versions


elegaanz:
Thank you!

PR #3593, opened 2025-12-05
to-stuff:0.4.0

nikmennega:
I am submitting
- [x] an update for a package

Description:
- All use of `eval()` now entirely eliminated.
- Remove reliance on `eval()` from:
	- color
- Color conversion via constructor now checks validity of all arguments.
- Expose new conversions:
	- float
	- int
	- scalar (returns either float or int as appropriate)
- Stroke miter-limit now treated as scalar.


elegaanz:
Thanks!

PR #3609, opened 2025-12-07
quest:0.2.0

kevinlin1:
I am submitting
- [x] an update for a package

Description: Rename the main template function name from 'exam' to 'quest' to better align with the package name!


saecki:
Thanks for the update!

PR #3612, opened 2025-12-08
orange-book:0.6.2

flavio20002:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: bug fixes and Typst version update.


saecki:
Thanks for the update!

PR #3620, opened 2025-12-09
cogsci-conference:0.1.1

daeh:
I am submitting
- [ ] a new package
- [x] an update for a package

Update style and instructions for the 2026 conference.


elegaanz:
Thank you.

PR #3624, opened 2025-12-09
xarrow:0.4.0

loutr:
I am submitting
- [ ] a new package
- [x] an update for a package

This update changes the parameters of xarrow and aligns with the `size` option from the native `stretch` function. It also introduces minor changes to the README.

### DIFF
--- a/packages/preview/xarrow/0.4.0/LICENSE
+++ b/packages/preview/xarrow/0.4.0/LICENSE
@@ -1,0 +1,14 @@
+xarrow: variable-length arrows in typst.
+Copyright (C) 2023-2024 Lucas Tabary-Maujean
+
+This program is free software: you can redistribute it and/or modify it under
+the terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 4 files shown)


elegaanz:
Thanks!

ice1000:
Hi, it seems that the CI checks are failed and I can't find 0.4.0 version of xarrow on typst universe. In the version history, 0.4.0 is not hyperlinked:

image

Any idea what's going on?

elegaanz:
It was published: the link doesn't exist because the default page is already the one for the latest version. If you go to a previous version by clicking its link you will see that it is no longer underlined either, but 0.4.0 will be.

ice1000:
https://packages.typst.org/preview/xarrow-0.4.0.tar.gz doesn't download anything for me (but thanks for the clarification)

elegaanz:
I was able to download it just fine. Do you have a more specific error message? Or do you just receive an empty file?

ice1000:
I am also able to download it now. Interesting...

Issue #3617, opened 2025-12-09
`./assets` in the `exclude` field of `typst.toml` actually matches `**/assets`


YDX-2147483647:
### Description

https://github.com/typst/packages/blob/8785ec6f6da2e7896db08857d38d6ed4cddebce8/docs/manifest.md?plain=1#L68

It looks like _an array of globs_ is the only description for the format of the `exclude` field in `typst.toml`.
This ambiguity cause problems.

https://github.com/QuadnucYard/ourchat-typ/issues/3 is an example.
This package has the following file structure.

- `/typst.toml`
- `/assets/` (to be excluded)
- `/src/`
  - `/src/assets/` (to be included)

The package specifies `./assets`, but surprisingly, it matches both `/assets/` and `/src/assets/`.

```toml
exclude = ["./assets"]
```

## Analysis

The bundler trims the `./` prefix, and uses [a library for `.gitignore`](https://docs.rs/ignore/0.4.25/ignore/overrides/struct.OverrideBuilder.html#method.add) to parse the glob.
As a result, `./assets` becomes `assets`, and matches recursively.
https://github.com/typst/packages/blob/8785ec6f6da2e7896db08857d38d6ed4cddebce8/bundler/src/main.rs#L336

## Workaround

Use `/assets`.
(not confirmed yet)

laurmaedje:
In Typst itself, `assets` and `./assets` are also equivalent, though I can see how in the context of a manifest, the dot could carry extra meaning. Regardless of whether the current behavior is sensible, using `/assets` seems like the clearest option and not like a workaround to me.

HackYardo:
I don't think it's a good idea to write something like "any dir", "assets", "the top level dir".
Perhaps there are no asset in a package. And what is top dir, root or /home/user/ ?
My suggestions are:
1. do not use relative path in `exclude` string array, it's not reasonable in manifest or gitignore;
2. make it clear that glob patterns will be applied **recursively** throughout the entire working tree;
3. attach a link to [Pro Git 2nd - Ignoring Files](https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository).

arepodesir:
> Hello world! _As it would turn out in a all but fair economy there is more free work to be found then paid work for those who then get paid from free work 😄 However out of love for this work of art gifted to humanity I thought I share my suggestion for this issue that I also had to deal with (AuDHD brain can't stand inconsistency :P)..._

[...]

PR #3637, opened 2025-12-11
Document the format of the `exclude` field in `typst.toml`

YDX-2147483647:
~~Resolves~~ Relates to #3617

~~I only provided two examples with a link to the git documentation.~~

I didn't mention details like `trim("./", at: start)`.
For those who have used git, it's sufficient to know that it's the gitignore format. For others, knowing ~~these two examples~~ the key point is enough.

### DIFF
--- a/docs/manifest.md
+++ b/docs/manifest.md
@@ -72,8 +72,8 @@ Optional:
   otherwise unnecessarily increase the bundle size. Don't exclude the README or
   the LICENSE, see [what to exclude]. Globs provided here have the same semantics
   as lines in a [gitignore file](https://git-scm.com/docs/gitignore). They are
-  applied _recursively_ throughout the entire project directory. Prepend a
-  forward slash (e.g., `/assets*/`) to avoid recursivity.
+  applied _recursively_ throughout the project directory. Prepend a forward
+  slash (e.g., `/assets*/`) to avoid recursive matching.
 
 Packages always live in folders named as `{name}/{version}`. The name and
 version in the folder name and manifest must match. Paths in a package are local


YDX-2147483647:
@HackYardo I see [your comment](https://github.com/typst/packages/issues/3617#issuecomment-3677949325) and change my wording.

- I replaced _working tree_ with _project directory_.

- I didn't mention relative paths, because I don't think people can think of `./` in the first place. (In the case of ourchat, the author tried `assets` first, and `./assets` was only a kind of debugging technique.)

  Also, the word _relative_ is ambiguous. All paths here are relative to the project directory.

- I didn't replace the link with [Ignoring Files](https://git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository#_ignoring). That article is too specific to the git workflow. All parts framed in red in the screenshot below do not apply here.
 
图片
.

HackYardo:
Fairly nice, the item you modify is a part of `manifest.md`,
so you'd better edit in the way there. My suggestions are:
1. the same semantics => some semantics
2. [gitignore file](https://...) => [gitignore file]
  - at the bottom of `manifest.md`: [gitignore file]: https://...
3. entire progect directory => `{version}/` folder

YDX-2147483647:
1. _the same semantics_ was copied from [`OverrideBuilder` in `ignore::overrides` - Rust](https://docs.rs/ignore/latest/ignore/overrides/struct.OverrideBuilder.html#method.add). I prefer the original expression here.

2. Okay, I'll change it later.

3. I think that's redundant. The context clearly indicates that this paragraph is talking about a certain version of a package.

    (Not every package needs the `exclude` field. Writing too much details might distract readers' attention.)

YDX-2147483647:
I've unlinked this PR from #3617, becaused Arepo Desir's post https://github.com/typst/packages/issues/3617#issuecomment-3724537521 (implemented in #3813) will still be relevant after this PR gets merged.

saecki (approved):

docs/manifest.md
```diff
@@ -70,7 +70,10 @@ Optional:
   files will still be available on typst universe to link to from the README.\
   To be used for large support files like images or PDF documentation that would
   otherwise unnecessarily increase the bundle size. Don't exclude the README or
-  the LICENSE, see [what to exclude].
+  the LICENSE, see [what to exclude]. Globs provided here have the same semantics
+  as lines in a [gitignore file](https://git-scm.com/docs/gitignore). They are
+  applied _recursively_ throughout the entire project directory. Prepend a
```
saecki:
```suggestion
  applied _recursively_ throughout the package directory. Prepend a
```

docs/manifest.md
```diff
@@ -70,7 +70,10 @@ Optional:
...
+  forward slash (e.g., `/assets*/`) to avoid recursivity.
```
saecki:
```suggestion
  forward slash (e.g., `/assets*/`) to avoid recursive matching.
```

saecki (commented):
What do you think of these changes?

YDX-2147483647:
👍 (and just pushed)

saecki:
Thanks!

PR #3641, opened 2025-12-11
alexandria:0.2.2

ensko:
I am submitting
- [ ] a new package
- [x] an update for a package

Alexandria provides an alternative to the built-in bibliography to allow a single document to have multiple bibliographies. This update upgrades Hayagriva to 0.9, bringing it in line with Typst 0.14 and fixing a few bugs. It also adds support for rendering `$...$` markup in references, and for using footnote bibliography styles.


saecki:
Thanks for the update!

PR #3645, opened 2025-12-12
Pull READMEs out of full index

reknih:
Aside from index.json, the bundler creates a second _full index,_ which is an implementation detail of typst.app. This index has grown to 18MB due to the READMEs included within. This makes it annoying to work with:

- Our CDN, Azure Front Door, does not automatically compress files larger than 8MB
- Downloads from JavaScript and WebKit spuriously fail due to keepalives
- The file occupies a relatively large amount of memory once loaded

Instead, we factor out the READMEs into their own folder that can be loaded on demand.

The new full index without the README field is called `index.extra.json`. To prevent disruption of running infrastructure, `index.full.json` is retained in storage but no longer updated. This exception can be removed soon. The new `readmes` folder is also an implementation detail of typst.app and should not be used by third parties.

### DIFF
--- a/bundler/src/main.rs
+++ b/bundler/src/main.rs
@@ -27,6 +27,7 @@ use self::timestamp::determine_timestamps;
 
 const DIST: &str = "dist";
 const THUMBS_DIR: &str = "thumbnails";
+const READMES_DIR: &str = "readmes";
 
 struct Config {
     out_dir: PathBuf,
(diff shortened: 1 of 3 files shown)


bundler/src/main.rs
```diff
@@ -84,6 +85,8 @@ fn main() -> anyhow::Result<()> {
         let mut package_errors = vec![];
         fs::create_dir_all(namespace_dir.join(THUMBS_DIR))
             .context("could not create output directory")?;
+        fs::create_dir(namespace_dir.join(READMES_DIR))
```
saecki: This should probably also be `create_dir_all` so it doesn't error if the directory already exists.

saecki (commented):
Apart from the one comment LGTM!

PR #3651, opened 2025-12-13
flyingcircus:4.0.0

Tetragramm:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added webassembly plugin to simplify the workflow and allow using plane builder links directly.


Tetragramm:
Hmm, the typst version makes sense, that was released today.  I don't know where the compiled document comes from. That's not in my repository.

elegaanz:
Yes, the CI was not yet updated to use 0.14.2. However, you should not specify this version as `compiler` in your manifest: it corresponds to the oldest version your package supports, so normally at least 0.14.0 should work, since the two patch versions that were released after that didn't change any behavior or introduce any new behavior. You might even be able to specify an older Typst version if you don't use any 0.14 specific feature.

elegaanz:
For some reason, I can't comment on the files directly, so I'll leave my review here:

Can you delete any file that is only useful for development? This includes the whole `.github/workflows` and `scripts` directories, from what I can see. This repository should only contain a copy of the files that are necessary for the package to run (its source code, images and data it imports, etc.) and its documentation (and all documentation files should be linked directly or indirectly from the README, as this is the only file that will be accessible on Typst Universe). Any other file is only relevant in your own repository.

Other than that, this new version looks good to me :)

elegaanz:
Thank you!

PR #3659, opened 2025-12-14
briefs:0.2.0

tndrle:
I am submitting
- [ ] a new package
- [x] an update for a package

### Added
- The position and size of the address box as well as the top margin of the recipient inside the address box are now adjustable. See arguments `address-box` and `recipient-top-margin`.

### Changed
- **Breaking:** The folding marks and hole punch marks are now customizable. The argument `hole-punch-mark` has been renamed to `hole-punch-marks` (with a trailing "s"). Also, the arguments `folding-marks` and `hole-punch-marks` are no longer boolean but can be `none` or a `dictionary`.


elegaanz:
Thanks!

PR #3672, opened 2025-12-16
cogsci-conference:0.1.2

daeh:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Use pseudo-data for example citations and authors, mirroring latest template from the CogSci repo. Make the format-authors() convenience function consistent with manual formatting.


elegaanz:
Thank you!

PR #3676, opened 2025-12-16
sos-ugent-style:0.3.0

th1j5:
I am submitting
- [ ] a new package
- [x] an update for a package

* Updated to typst v0.14.0
* Includes poster template
* Tight integration with guix for bit-reproducibility
* ...


th1j5:
The failing check are only warnings, at the moment, I want to keep on using 'GFM task lists' even if they're not supported on the universe

saecki:
Thanks for the update!

PR #3680, opened 2025-12-17
cogsci-conference:0.1.2

daeh:
I am submitting
- [ ] a new package
- [x] an update for a package

Minor change. Fixing a typo in the template's bibliography `template/bibliography.bib`.

I'm hoping that I can just amend this to the current version `v0.1.2` since the core `cogsci.typ` file hasn't changed. Sorry for trying to update an existing version, I hope that's ok, but if not I'll resubmit this as a new patch. Thanks!


saecki:
We don't allow any package edits anymore to ensure reproducibility of packages, see, https://github.com/typst/packages/issues/2671. You can create a new version with those edits instead

PR #3683, opened 2025-12-17
patstdlib:0.2.0

patrick-kidger:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- Update for typst 0.14 (which introduces `title`, so we don't need to define our own.
- Added `fonts(heading-sizes: ...)` argument.

### DIFF
--- a/packages/preview/patstdlib/0.2.0/README.md
+++ b/packages/preview/patstdlib/0.2.0/README.md
@@ -1,0 +1,54 @@
+<h1 align='center'>patstdlib</h1>
+
+Some standard pieces I find useful: fonts, subfigures, algorithms, labeled enums, etc.
+
+## Documentation
+
+Please see the inline documentation attached to each object. A full list of all available functionality can be found in [`src/lib.typ`](https://github.com/patrick-kidger/patstdlib/blob/main/src/lib.typ).
+
+## Functionality
+
+Here's a few highlights:
+
+**Labeled enums**
+
+```typst
+#show: enable-referable-enums
+
+#referable-enum("step")[
+  + Foo
+  + Bar
+  + Baz <step:baz>
+]
+
+In @step:baz then ...
+// Renders as 'In step 3 then ...'
+```
+
+**Subfigures**
+
+```typst
+#show: enable-referable-subfigures
+#set enum(full: true)
+
+#figure(
+  caption: [Main caption],
+  grid(
+    columns: 2,
+    subfigure(
+      caption: [Caption for subfigure '(a)'],
+      image(...)
+    ),
+    subfigure(
+      caption: [Caption for subfigure '(b)'],
+      image(...)
+    ),
+  )
+)
+```
+
+**Significant figures**
+
+```typst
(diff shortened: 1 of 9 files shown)


packages/preview/patstdlib/0.2.0/README.md
```diff
@@ -0,0 +1,53 @@
+<h1 align='center'>patstdlib</h1>
+
+Some standard pieces I find useful: fonts, subfigures, algorithms, labeled enums, etc.
+
+## Documentation
+
+Please see the inline documentation attached to each object. A full list of all available functionality can be found in `src/lib.typ`.
```
saecki: Could you add a link to the relevant files like this: ``[`src/enum.typ`](src/enum.typ)``, or link the `src` directory? Otherwise the files will be very hard to find on Typst Universe.

packages/preview/patstdlib/0.2.0/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Functionality
+
+Here's a few highlights:
+
+**Labeled enums**
+
+```typst
+#show: enable-referable-enums
```
saecki, line 16: This example doesn't compile without `#set enum(full: true)`

patrick-kidger:
Done.

And whilst I appreciate the feedback, can I ask what the expectations are around Typst Universe? In this case, I don't think either change would affect the ability for my collaborators to use the package – but delaying publication of the package definitely does.

Is the intention that TU be a manually-curated space, and that for sharing work amongst collaborators it would be better to host our own package distribution mechanism somewhere? That's fine if so – it's very kind of you to host our packages at all, after all – I just want to make sure expectations are set correctly. (And that TU shouldn't be treated like PyPI / npm / crates / etc., where there's more of an anything-goes approach to things.)

saecki:
> Done.

Thanks, LGTM!

> And whilst I appreciate the feedback, can I ask what the expectations are around Typst Universe? In this case, I don't think either change would affect the ability for my collaborators to use the package – but delaying publication of the package definitely does.

It's not exactly a curated list of packages, but we do try to ensure some level of quality, mostly in regard to documentation.
One of those expectations is that a package is usable for a reasonable amount of people, and that includes some documentation of how the package works. You already had documentation it just would've been hard to find.
Another thing I try to ensure is that readme examples work out of the box (when reasonable). Just a small change here, which takes us a little bit of time, will save a huge accumulated amount of time and annoyance for other people that would most likely just copy the example to get started.

> Is the intention that TU be a manually-curated space, and that for sharing work amongst collaborators it would be better to host our own package distribution mechanism somewhere? That's fine if so – it's very kind of you to host our packages at all, after all – I just want to make sure expectations are set correctly. (And that TU shouldn't be treated like PyPI / npm / crates / etc., where there's more of an anything-goes approach to things.)

That depends on what you mean by collaborating. It's not meant as a hub for collaborating comparable to a git repository, but that's also not the case for the other package registries you mentioned. You can definitely publish packages to share amongst collaborators, but as you already mentioned there will be a delay due to the review process. See also: https://github.com/typst/packages/blob/main/docs/README.md#package-submission-guidelines

PR #3693, opened 2025-12-19
biz-report:0.3.1

garethwebber:
I am submitting
- [ ] a new package
- [x] an update for a package

This is a in version bug fix. The dropcappara component silently dropped the second character of the paragraph. This was because the array starts at zero not one.

It's a one line fix.

The PR seems to be picking up the original 0.3.0 submission thats already been merged. The actual change is the index change in last commit


elegaanz:
For reproducibility reasons, we do not allow already published versions to be modified. You should publish these changes as v0.3.1.

garethwebber:
Changed as requested.

elegaanz:
Thanks.

PR #3700, opened 2025-12-19
Offload thumbnail processing to the rayon thread pool

saecki:
I was wondering why bundling was so slow, and it turns out most of the time is spent encoding thumbnails.
Offloading the processing reduced the bundling time on my machine from 11 to about 3.5 minutes. Of course in the github actions runner this might not have that huge of an effect, but it will probably still be faster.

For reference here are the numbers from before in minutes:
```
parse_manifest = 0.0067
build_archive = 0.6153
read_readme = 0.0004
validate_archive = 0.0362
write_archive = 0.0079
process_thumbnail = 8.6333
total = 9.2998
```

### DIFF
--- a/bundler/src/main.rs
+++ b/bundler/src/main.rs
@@ -16,7 +16,7 @@ use image::codecs::webp::{WebPEncoder, WebPQuality};
 use image::imageops::FilterType;
 use semver::Version;
 use spdx::LicenseId;
-use typst_syntax::package::{PackageInfo, PackageManifest, TemplateInfo, UnknownFields};
+use typst_syntax::package::{PackageInfo, PackageManifest, UnknownFields};
 use unicode_ident::{is_xid_continue, is_xid_start};
 
 use self::author::validate_author;
@@ -32,6 +32,11 @@ const READMES_DIR: &str = "readmes";
 struct Config {
     out_dir: PathBuf,
     skip_license_validation: bool,
+}
+
+enum Msg<T> {
+    Data(T),
+    Done,
 }
 
 fn main() -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


bundler/src/main.rs
```diff
@@ -186,6 +187,19 @@ fn main() -> anyhow::Result<()> {
         }
     }
 
+    // Send the finished message.
+    thumb_sender.send(Ok(true)).unwrap();
+
+    // Wait for thumbnail processing to complete.
+    let mut thumbnail_errors = Vec::new();
```
elegaanz: Can't that be added to `namespace_errors` directly too?

bundler/src/main.rs
```diff
@@ -186,6 +187,19 @@ fn main() -> anyhow::Result<()> {
...
+    while let Ok(res) = thumb_receiver.recv() {
+        match res {
+            Ok(true) => break,
+            Ok(false) => (), // Ok
+            Err(e) => thumbnail_errors.push(e),
```
elegaanz: It's not too bad because the code is not that long nor complex, but I was a bit confused by the meaning of the messages that are exchanged over this channel. I think a comment somewhere explaining that the boolean in the `Ok` variant is used to tell whether the processing is done would be nice to have.

bundler/src/main.rs
```diff
@@ -186,6 +187,19 @@ fn main() -> anyhow::Result<()> {
         }
     }
 
+    // Send the finished message.
+    thumb_sender.send(Ok(true)).unwrap();
```
elegaanz: Maybe I'm missing something, but is it guaranteed that this message will be sent (and thus received) after all `Ok(false)` messages are queued by the processing tasks?

saecki: No you're totally right, I previously used a scope to block on any outstanding tasks, but messed this up when refactoring.

saecki: I've introduced a scope again

bundler/src/main.rs
saecki: I've added a `Msg` enum.

bundler/src/main.rs
saecki: Done.

elegaanz:
Thank you :)

PR #3701, opened 2025-12-19
presentate:0.2.3

pacaunt:
I am submitting
- [ ] a new package
- [x] an update for a package

Added a new motion function for precise and easy control of the sequence for displaying elements. Users can now define their own element under the `tag` function and control its display sequence using the `controls` argument of the `motion` function.


elegaanz:
Thanks.

PR #3705, opened 2025-12-20
cogsci-conference:0.1.3

daeh:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixing a typo in the example bibliography


elegaanz:
Thank you!

PR #3710, opened 2025-12-20
ox-scholar:0.1.1

fcelli:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixes the appearance of the template's thumbnail in typst universe. Apparently PNGs must have a filled background or the thumbnail is completely black in the universe page.


elegaanz:
Thank you, and sorry we didn't catch that in the previous review.

PR #3713, opened 2025-12-20
meander:0.3.1

Vanille-N:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixes two assertion errors regarding internationalization (languages not supported by hypher) and handling of obstacles that overflow into the margins.


elegaanz:
Thank you.

PR #3715, opened 2025-12-21
simple-cvut-report:0.2.0

firu11:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:

This package is a simple template for ČVUT (Czech Technical University in Prague) students.
It has a neat front page with the university logo, table of contents and citations. Students can use it to have a great starting point for their thesis or term papers.


elegaanz:
Thank you for making this template. The way you decided to structure it is not really conventional. What you would normally do is to have a "template" function in `lib.typ` that takes some content (and some other parameters, here the thesis title for example), display some common content (the title page for example), configures the styles of the document using `show` and `set` rules, and finally renders the provided content using that styling. This is a bit less flexible than letting the user edit everything, but it also lets them focus on their content without worrying about having to understand the provided style nor breaking it by mistake. They can still override your `show` rules if they need to, which gives enough flexibility in general. Could you refactor your template to follow this pattern please?

Also, can you double check that you are allowed to distribute the university logo, and if so can you please state under which terms it can be used in your README, if the university has any specific rules?

Thank you :)

firu11:
Hi,
well yeah, I chose this approach, because I expect the user to want to overwrite the template a bit. Since I did not define some "rules", rather prepared the document as a whole, I wanted the user to be able to modify it fully. I get your point though. I'll try to refactor it and maybe submit it, but I don't know if the "conventional approach" will suit this use-case well.

Regarding the logo, I even added it to the bibliography: https://github.com/typst/packages/pull/3715/files#diff-8d53010fa95fd19d53aa7f6f73fc97a505b3665d7a3dbcce4ddc972dbd1cce78R1-R4
It's from Wikipedia where it states the "Attribution-ShareAlike 3.0 Unported" licence. I'll add it to the readme.

Thank you for looking into this though!

elegaanz:
Hey, did you have a chance to refactor your template? No worries if not, I just wanted to check what the status was here. If you don't have much time to make the changes for now, I will close this PR to keep our review backlog clear (but you can of course open a new one once ready).

firu11:
Hey, I've refactored the template. Would you mind taking another look? Thanks!

packages/preview/simple-cvut-report/0.2.0/typst.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "simple-cvut-report"
+version = "0.2.0"
+compiler = "0.14.2"
```
elegaanz: This field corresponds to the oldest compiler version you support, not necessarily the one you used when developing it. I doubt you need a fix that was introduced in 0.14.2 for the package to work correctly, it should at least be 0.14.0. And from a quick look, you don't use any feature that was introduced in 0.14 so it may even be lower than that. Can you change it please? Otherwise people who were not able to update (because they use a system which doesn't provide the very last Typst version, or because they need some other package that breaks with a more recent compiler) won't be able to install this package.

elegaanz (commented):
Sure, I only have one small remark, otherwise it looks to me good now.

packages/preview/simple-cvut-report/0.2.0/typst.toml
firu11: Got it. I've tried 0.11.0 and it seems to work just fine.

elegaanz:
Thank you!

PR #3719, opened 2025-12-22
basic-report:0.4.0

roland-KA:
I am submitting
- [ ] a new package
- [x] an update for a package

- Date format can be freely configured 
- Content heading uses auto default 
- Workflow for contributions explained in README


roland-KA:
Files haven't been in the correct subdirectory. It's now fixed. So the checks should succeed.

elegaanz:
Thank you.

PR #3743, opened 2025-12-26
unofficial-sdu-lab-report:0.1.1

arshtyi:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Lowered compiler requirement and modified some implementations accordingly.


saecki:
Thanks for the update!

PR #3748, opened 2025-12-27
tcdm:0.0.1

YDX-2147483647:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: HTML generator for best-of lists, especially [Best of Typst (TCDM)](https://ydx-2147483647.github.io/best-of-typst/).

This package is an integration with the third-party tool https://github.com/best-of-lists/best-of-generator, so it looks like a library. However, the actual workflow (see README) will use this package as a template. Therefore, I publish it as a template.

### Notes

> [The thumbnail must depict one of the pages of the template **as initialized.**](https://github.com/typst/packages/blob/3267c746ff49f08dfd77bea0523286a36e0fad69/docs/manifest.md?plain=1#L176C45-L177C45)

I can't follow the above rule. This package loads external data from best-of-generator (see README), so the template itself can't compile. I have to create the thumbnail with additional data files.

An alternative method is to provide example data, but that means ~100 lines of yaml + a 60-column csv.
Moreover, it will create confusion, because the end users are not expected to care about the data format.


YDX-2147483647:
This is an HTML-only template, so I guess these errors are acceptable?

> typst-package-check: The following error was reported by the Typst compiler: unknown variable: html

saecki:
Yeah, the CI won't handle HTML only templates. But the template should still compile out of the box, without any further actions required by the user. That's currently not the case, because the `json` file is missing.

YDX-2147483647:
> the template should still compile out of the box

Is that a strict rule? If so, I can add placeholder contents like this:

```typst
#let (configuration, statistics, assets, body) = load(
  projects-data: json(bytes("[]")),
  projects-yaml: yaml(bytes(
    ```yaml
    configuration:
      category_heading: robust
      …
    categories: []
    labels: []
    ```.text
  )),
  // NOTE: Please replace the above lines with the following in actual use
  // projects-data: json("/build/latest.json"),
  // projects-yaml: yaml("/projects.yaml"),
)
```

(That would make `main.typ` more confusing and introduce an additional setup step in README , though.)

saecki:
> Is that a strict rule?

The readme says this:
> If your package includes a  template, it should compile out-of-the-box. 

I guess there could be exceptions, but I don't see a good reason why that shouldn't apply here.
You could just include example `json` and `yaml` files that the user can overwrite.

YDX-2147483647:
Fine. I’ll update the template.

> You could just include example json and yaml files that the user can overwrite.

(I’ve expressed my concerns at the end of the main PR description.)

saecki:
I don't really see the problem here. If you include example data and a thumbnail based on that example data, you'd follow the rule as far as I'm concerned. Same for the format, In your examples you were already using json.

YDX-2147483647:
I've just updated my package. Please take another look!

(In any case, thanks for your efforts in maintaining the quality of the universe.)

---

> typst-package-check / @preview/tcdm:0.0.1
> packages/preview/tcdm/0.0.1/typ/example-config.typ#L1
> This file seems to be an example, and should probably be added to `exclude` in your `typst.toml`.

This file is intended to be included.

saecki:
> This file is intended to be included.

Yeah, the warning is just a heuristic :+1:

Thanks, everything works on my machine!

PR #3749, opened 2025-12-27
theoretic:0.3.0

nleanba:
I am submitting
- [ ] a new package
- [x] an update for a package

## Styling Changes
The old `fmt-prefix` and `fmt-body` have been removed, instead there is now more fine-grained control available.

There are now styled presets: `basic`, `corners`, `columns`, `fancy`, `bar`, `colorbox`. These can be controlled by a `variant` parameter on the theorem (e.g. `variant: "important"` adds a frame around the theorem in most presets), and a fine-grained `options` dictionary.

The `basic` style closely mimics the "amsthm" style, and there is similar `options` available: e.g. remove the `.` after the title by using `options: (head-punct: none)`.

For full control and defining custom styles, there is a `show-theorem` parameter which allows for completely custom rendering.

## Other Changes
- `title` and `label` arguments can now be passed as positional or named arguments
- `solution` is now a named argument
- It now takes a piece of content as `suffix` (qed), instead of a function (~~`fmt-suffix`~~)
- Nested proofs are now supported (with correct QED placement)
- There is a new `qed-in-equation()` helper function to help with manually placed QEDs after block equations.
- Restated theorems now use the same style as the original automatically.
- Solutions section: You can now customize what environment/style to use for the solutions.

[Please see the manual for full documentation.](https://github.com/nleanba/typst-theoretic/blob/v0.3.0/main.pdf?raw=true)

**Full Changelog**: https://github.com/nleanba/typst-theoretic/compare/v0.2.0...v0.3.0

### DIFF
--- a/packages/preview/theoretic/0.3.0/src/base.typ
+++ b/packages/preview/theoretic/0.3.0/src/base.typ
@@ -1,8 +1,5 @@
 /** base package functionality */
 
-/// -> state(boolean)
-#let _needs_qed = state("_thm_needs_qed", false)
-
 #let _qeds = state("_thm_qed_stack", ())
 
 /// Place the QED mark last pushed to `state("_thm_qed_stack")` here.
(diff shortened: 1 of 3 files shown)


nleanba:
The automatic warning is due to a argument named `_defaults`, it is an argument for advanced usage of a function that is already only for advanced usage, so having the underscore sends the correct signal to users. I would like to keep it this way.

saecki:
Thanks for the update!

PR #3760, opened 2025-12-30
kthesis:0.1.3

RafDevX:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Better support for abstract serialization, courtesy of @diogotcorreia.


saecki:
Thanks for the update!

PR #3766, opened 2025-12-31
frugal-haw-kiel:0.1.0

Arne-Berner:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: The package is supposed to help people at the haw (university of applied science) kiel to start with a thesis template that conforms to the requirements.


Arne-Berner:

Edit: I now use STIX Two Text instead of Times New Roman, because the adobe licensing seems to be a bit more open to change. 
Also some paths in the example were still pointing to @local instead of @preview. Is the reimport in example/abstract.typ possible as seen here? It was possible via the @local directory. 

Thank you for taking the time!

(I did forget to recompile the main.typ in examples, hence the two force pushes..)

packages/preview/frugal-haw-kiel/0.1.0/example/main.pdf
saecki, line 1: Could you remove this file?

packages/preview/frugal-haw-kiel/0.1.0/example-deutsch/main.typ
saecki, line 1: You probably want to add a [`[template]` section](https://github.com/typst/packages/blob/main/docs/manifest.md#templates) to your `typst.toml` manifest. So a user can conveniently initialize a project from this package.

packages/preview/frugal-haw-kiel/0.1.0/README.md
```diff
@@ -0,0 +1,90 @@
+This is a simple HAW Thesis template based off the HAW Hamburg template and changed with the requirements set by Prof. Dr. Lüssem.
+
+## Example
+The [full example](https://github.com/arne-berner/haw-kiel-template) can be seen on github.
+
+### Getting started
+Some parameters are strictly optional, especially if you don't intend to write your thesis in both german and english. It is good practice to seperate your chapters into files.
+
+```ts
```
saecki: For typst syntax highlighting  you can use the `typ` extension:
```suggestion
```typ
```

saecki:
> Is the reimport in example/abstract.typ possible as seen here? It was possible via the @Local directory.

Yes this is the correct way to do it for templates.

packages/preview/frugal-haw-kiel/0.1.0/README.md
```diff
@@ -0,0 +1,90 @@
...
+#import "./abstract.typ":abstract
+#import "./acknowledgment.typ":acknowledgment
+#import "./abbreviations.typ":abbreviations
```
saecki: This isn't wrong per se, but I think it would be more idiomatic to use `include` to include the content from the files directly:

```typ
#show: thesis.with(
  ...
  abbreviations: include "./abbreviations.typ",
  ...
)
```

That way, you can just write normal Typst markup in the included files.

You can of course leave it the way it is, if you like that better for some reason.

packages/preview/frugal-haw-kiel/0.1.0/example/main.pdf
Arne-Berner: Thank you very much for your review! I thought this would be something that would be removed by the typst.toml? Did I get the glob wrong with /exclude/? 
Or does it have nothing todo with the exclude, but that the repo should be free of pdfs? 
I'm just trying to make sure to do it right with the next package :smile:

packages/preview/frugal-haw-kiel/0.1.0/README.md
Arne-Berner: Thank you for pointing this out to me! I know that I wasn't sure about this. :smile: 
I remember now: 
Glossary takes an array as an argument and I wasnt sure how I could include an array with that syntax.

Arne-Berner: I could not get include to work for arrays since it would convert it to content. 
If there is a way, I'd love to learn, how to do it :)

Arne-Berner:
Thanks to you I also managed to fix some issues when a glossary was not provided or when someone tried to use the "en" version. Thanks a lot!
Is it uncommon to push the examples with the template in this repo?

packages/preview/frugal-haw-kiel/0.1.0/example/main.pdf
saecki: Yeah, we try to reduce the amount of somewhat large files in the repo

packages/preview/frugal-haw-kiel/0.1.0/README.md
saecki: Oh, you're right. I managed to pick out the one file where `include` doesn't make sense :sweat_smile:
But I see you found a solution using YAML :)

packages/preview/frugal-haw-kiel/0.1.0/typst.toml
```diff
@@ -0,0 +1,19 @@
...
+authors = ["Arne Berner"]
+repository = "https://github.com/arne-berner/haw-kiel-template"
+license = "MIT OR Apache-2.0"
+
+keywords = ["HAW", "Kiel", "Hochschule für Angewandte Wissenschaften", "University of Applied Sciences", "Hochschule", "University", "Bachelorarbeit", "Bachelor"]
+categories = ["thesis"]
+disciplines = ["computer-science", "engineering"]
+exclude = ["/example/"]
+
+[template]
```
saecki, line 16: The `template` naming is a bit convoluted, a template is more of an example/starting point of how to use the package. In your case it should point to either `example-deutsch` or `example-english` and the entrypoint should be `main.typ`.

Arne-Berner: It took me a while, but I think I understood you now:
Please correct me, but the template described in the toml is not the [template as described in the docs](https://typst.app/docs/tutorial/making-a-template/), but it is [the template described in the README](https://github.com/typst/packages/blob/6fc12cafee561bce6c8dcfc7f439c5d00481a98b/README.md?plain=1#L20C1-L22C8), right?
So if people wanted to have a starting point with stuff in it, they could use that.
I might need to rework it a bit then. I used it as a showcase rather than as a starting point for others. 
Would it make more sense to have the following filestructure:

package/
├─ lib/
│  ├─ lib.typ
│  ├─ pages/
├─ init/
│  ├─ init.typ
├─ examples/
│  ├─ example-deutsch/
│  ├─ example-english/
├─ LICENSE
├─ README

Where:
- init is the template as described in the README 
- examples can be used to see a working example 
- lib is the template as described in the docs
? 

Is it preffered to use empty files or fill them with example text in a template? I'd tend to leave them empty and just create some sort of structure.

Is there some naming conventions for templates (from the readme)? The first example I found used the name template for both: The readme version and the docs version. So a folder named template/main.typ and a template.typ file at the root of the project. 
I could not find an open issue about the naming of template vs template yet. Do you know more about it? :)

Thank you for your reviews and time.

saecki: > Please correct me, but the template described in the toml is not the [template as described in the docs](https://typst.app/docs/tutorial/making-a-template/), but it is [the template described in the README](https://github.com/typst/packages/blob/6fc12cafee561bce6c8dcfc7f439c5d00481a98b/README.md?plain=1#L20C1-L22C8), right?
So if people wanted to have a starting point with stuff in it, they could use that.

Exactly


> Would it make more sense to have the following filestructure:
> 
> package/
├─ lib/
│ ├─ lib.typ
│ ├─ pages/
├─ init/
│ ├─ init.typ
├─ examples/
│ ├─ example-deutsch/
│ ├─ example-english/
├─ LICENSE
├─ README
>
> Where:
    init is the template as described in the README
    examples can be used to see a working example
    lib is the template as described in the docs
    ?

Sure, sounds good


> Is it preffered to use empty files or fill them with example text in a template? I'd tend to leave them empty and just create some sort of structure.

I don't think there really is a preferred way, do whatever feels right to you :)


> Is there some naming conventions for templates (from the readme)? The first example I found used the name template for both: The readme version and the docs version. So a folder named template/main.typ and a template.typ file at the root of the project.

Usually the library code (docs template as you called it) has a `lib.typ` entrypoint and the code is placed in a `src` directory. The starting point (readme template) is placed in a `template` directory with a `main.typ` entry point.
An `examples` directory for any additional content is also quite common. If you include examples, make sure to create links from the readme, otherwise people will have a hard time finding them.

saecki:
Hey, this PR has been stale for quite a while, and to keep things organized we're gonna close it for now. You can pick it back up at any time and reopen it :)

PR #3779, opened 2026-01-04
patstdlib:0.3.0

patrick-kidger:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes
- Breaking change: `fonts` now uses Typst defaults by default.
- Breaking change: `topmatter` no longer sets numberings or page margin.
- Introduced `composition` to set justification, heading around spacings etc.
- Introduced `numberings` to set numberings for page, heading, etc.


saecki:
Thanks for the update!

PR #3787, opened 2026-01-05
zero:0.6.0

Mc-Zen:
I am submitting
- [x] an update for a package

### Version 0.6.0
_Exponent modes and rounding improvements_

Exponents:
- ⚠️ Breaking Change: Removes the parameter `num.fixed` which is now replaced by the exponent mode `fixed` (see below). 
- Adds new exponent modes: 
  - `auto` for standard formatting, 
  - `"sci"` for automatic scientific notation, 
  - `(fixed: n)` for fixing the exponent to a specific value, and 
  - `"eng"` for automatic engineering notation. 
- Improves rendering of exponents for `math: true` to guarantee a look just like the number was generated by hand with an equation.

Rounding:
- ⚠️ Breaking Change: Renames rounding modes `"down"` to `"towards-negative-infinity"` and `"up"` to `"towards-infinity"`. 
- Adds an option `round.ties` to configure the behavior for rounding the digit 5. 
- Adds two new rounding directions `"towards-zero"` and `"away-from-zero"`. 
- Adds support for padding with a minimum number of digits when rounding. 
- Fixes rounding when no integer part is given (like `num[.5]`). 

Units:
- Fixes spacing with angle units (there should be no space between the number and an angle unit). 
- ⚠️ Breaking Change: The unit alternatives with British spelling `zi.metre` and `zi.litre` have been removed for consistency and to avoid confusion. 
- Adds an option `unit.lowercase-liter` that enables a lower case symbol for the unit liter. 

Miscellaneous:
- Fixes parsing of `[#""]` values in formatted tables.

### DIFF
--- a/packages/preview/zero/0.6.0/README.md
+++ b/packages/preview/zero/0.6.0/README.md
@@ -222,7 +222,7 @@ Rounding can be configured with the `set-round()` function.
   - `"towards-zero"`: Round towards zero, e.g., 2.8 → 2 and −2.8 → −2. 
   - `"away-from-zero"`: Round away from zero, e.g., 2.3 → 3 and −2.3 → −3. 
 - `ties: str = "away-from-zero"` : How to round ties, i.e., when the digit at the rounding position is 5. 
-  - `"away-from-zero"`: Always round ties away from zero. This is also the behaviour of `calc.round`. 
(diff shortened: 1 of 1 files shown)


Mc-Zen:
I checked the Readme code syntax warnings from the package checker, they are all fine.

elegaanz:
Thanks for the update :)

PR #3789, opened 2026-01-05
board-n-pieces:0.8.0

MDLC01:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog

- `play` and `pgn` now return the successive moves as well.
- Allow passing an array of pairs instead of an array of strings to `board.arrows`.
- Add `square-stroke` parameter to `board`.


elegaanz:
Thanks!

PR #3813, opened 2026-01-08
Implemented normalize_exclusion_glob on patterns & tests

arepodesir:
I am submitting

* [ ] a new package
* [ ] an update for a package
* [X] **an update for the bundler**

### Description

This PR updates the internal exclusion logic in `src/main.rs` to fix an issue where all exclusions were forced to be recursive.

**The Problem:**
Currently, the bundler strips `./` prefixes via `trim_start_matches("./")` before passing them to the `ignore` crate. This causes specific paths like `exclude = ["assets"]` or `exclude = ["./assets"]` to be treated as global recursive globs. As a result, a user trying to exclude a root `/assets` folder unintentionally excludes deeply nested folders like `/src/assets`. Refer to #3617 

**The Solution:**
I introduced a `normalize_exclusion_glob` function that implements "Jekyll-style" exclusion logic, better matching user intent:

1. **Implicit Anchoring:** Clean paths (e.g., `"assets"`) are now anchored to the root (`/assets`).
2. **Explicit Anchoring:** Paths starting with `./` (e.g., `"./assets"`) are explicitly anchored to the root (`/assets`).
3. **Recursive Globs:** Strings containing wildcards (e.g., `"*.png"`, `"src/**/*.rs"`) remain recursive.
4. ~**Windows Support:** Backslashes are normalized to forward slashes to ensure consistent behavior across OSs.~

**Changes:**

* Modified `build_archive` in `src/main.rs` to use the new normalization logic.
* Unit tests for the glob normalization logic.
* I also added Integration tests that build a real `.tar.gz` archive in a temporary directory to verify that files are correctly included/excluded.

* Updated `src/main.rs` visibility modifiers (`pub(crate)`) to allow the test module to access internal logic.

### Checklist

* [X] Implemented `normalize_exclusion_glob` to handle explicit (`./`) and implicit root anchors.
* [X] Added `src/tests.rs` with unit and filesystem integration tests.
* [] Verified that recursive globs (e.g., `*.png`) still work as expected.
* [] Verified that nested folders (e.g., `src/assets`) are preserved when root `assets` is excluded.
* [X] Ran tests locally using `cargo test`.

### DIFF
--- a/bundler/Cargo.toml
+++ b/bundler/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "bundler"
 version = "0.1.0"
-edition = "2026"
+edition = "2021"
 publish = false
 
 [dev-dependencies]


YDX-2147483647:
As a regular Windows user, I don’t think we should support so-called _Windows-style paths_.

It makes typst.toml possibly platform-dependent, and backslashes need special treatment when entering in a toml file. 

Also, on Windows, forward slashes also work with explorer.exe, powershell, python pathlib, git bash, …

arepodesir:
That sounds reasonable. Contributing to a major repo it just seemed prudent to make a requirements sheet focused on compatibility as I don't know what people use. I can revise the exclusion ~

saecki:
This would be a major breaking change that would also affect existing packages, as they are re-bundled every time the deployment workflow runs. Before we could think about merging any sort of change, that would have to be addressed first.
While I can see that the current behavior is a bit confusing, I think there is some more discussion to be had before taking any concrete action.

elegaanz:
To not let this PR sit forever, I'm closing it for now. If you have proposals to solve the aforementioned issues, please share them in the original issue. Once we have reached a consensus on how we should deal with that, another PR can be opened.

PR #3821, opened 2026-01-09
versatile-apa:7.2.0

jassielof:
I am submitting
- [ ] a new package
- [x] an update for a package

- Authoring logic was reworked
- For simplicity, custom authors/affiliations were removed and other parameters, the title page and abstract were separated out of the template styling to not lock users if they want a different title page or abstract
- Quotations and running head limits are now configurable, quotations are settable via state, running head via the template styling since it affects the same running head.
- Text script for Serbian (Latin and Cyrillic) are now supported, thus scriptted-terms are supported.
- README has a better explanation/guide
- Change of license to Apache v2.0 and explanation of MIT/-0 error in previous versions.
- CSL style comes bundled with the template instead of being initialized with it.
- Annex and Addendum terms were removed as they no had use case, instead if the user wants to override a term, now it's possible as well as having support for custom terms of a non-supported language (not built into the template).
- Major rework on layout for simplicity of rules and compliance.


jassielof:
Sob, wait I found a little bug with my abstract function.
I don't really get what's there to fix by the bot.

packages/preview/versatile-apa/7.2.0/README.md
```diff
@@ -0,0 +1,166 @@
...
+- Serbian
+
+If the user wants to contribute support for more languages, they can contribute to the template, or using the custom terms option in the template.
+
+##### Unsupported languages and custom terms
+
+In the cases where the language is not supported in the template, you can either make a PR or file an issue in the repository, so it can be added and updated in the next template version. Otherwise you can use the `custom-terms` parameter, which expects a dictionary mapping English terms to their translations. For example for Japanese, the expected format is:
+
+```typ
+#let custom-terms: (
```
saecki: I think this should be:
```suggestion
#let custom-terms = (
```

packages/preview/versatile-apa/7.2.0/README.md
```diff
@@ -0,0 +1,166 @@
...
+  "Keywords": "キーワード",
+  "Appendix": "付録",
+  "Note": "注",
+)
+```
+
+This is also useful for overriding specific terms in supported languages. Such as:
+
+```typ
+#let custom-terms: (
```
saecki: And same here

saecki (commented):
Thanks for the update! I've left a comment regarding the warning.

jassielof:
Ohhh, I didn't know it was evaluated. That's cool

PR #3836, opened 2026-01-12
typslides:1.3.2

manjavacas:
I am submitting
- [ ] a new package
- [x] an update for a package

- Fixed bug causing duplicated outline entries.
- Added back-color parameter (global/slides).
- Minor documentation changes.

### DIFF
--- a/packages/preview/typslides/1.3.2/typst.toml
+++ b/packages/preview/typslides/1.3.2/typst.toml
@@ -2,7 +2,7 @@
 name = "typslides"
 version = "1.3.2"
 entrypoint = "lib.typ"
-compiler = "0.14.2"
+compiler = "0.14.0"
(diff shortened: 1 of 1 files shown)


packages/preview/typslides/1.3.2/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "typslides"
+version = "1.3.2"
+entrypoint = "lib.typ"
+compiler = "0.14.2"
```
elegaanz: If you don't depend on any fixes that were introduced in 0.14.2, this number should be lower, as it corresponds to the oldest compiler version your package supports.

manjavacas: Thanks! I've just updated the compiler version to 1.4.0.

elegaanz:
Thanks!

PR #3837, opened 2026-01-12
juti:0.0.3

nafkhanzam:
I am submitting
- [ ] a new package
- [x] an update for a package

Changed the inputs to be string so that:
- the short name can be inferred from the full name, and
- the email `at` symbol does not require escape character `\`.

### DIFF
--- a/packages/preview/juti/0.0.3/juti.typ
+++ b/packages/preview/juti/0.0.3/juti.typ
@@ -201,7 +201,7 @@
     revised: [will be set by the editor],
     accepted: [will be set by the editor],
     online: [will be set by the editor],
-    doi: [will be set by the editor],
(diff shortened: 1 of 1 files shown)


elegaanz:
Thanks.

elegaanz:
Just a question: I can see that this package is linked on the journal website, and that it is named after the journal without any distinctive part (which is normally allowed only for official templates) but it never underwent the official affiliation process. Because it is linked on the website, I can consider it to be officially endorsed without the confirmation verification e-mail, but I would like to double-check with you whether you want it to be displayed as "Officially affiliated" on Typst Universe?

nafkhanzam:
Yes, that'd be great as I'm one of the maintainer of the journal.

PR #3841, opened 2026-01-12
booticons:0.0.1

DavZim:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Provides a single function `#bsicon("airplane")` to display Bootstrap Icons. The package ships all 2000 svg icons (all MIT licensed).


packages/preview/bootstrapicons/0.0.1/README.md
```diff
@@ -0,0 +1,43 @@
...
+Christmas Present: #bsicon("airplane", color: red)
+
+If you want, you can make it large as well or use different color specifications:
+
+#bsicon("check2-square", height: 3em, color: rgb("#ff0000"))
+
+Find the source of this project at #bsicon("github") #link("https://github.com/davzim/bootstrapicons")
+```
+
+Which will generate this [PDF](https://github.com/davzim/bootstrapicons/examples/mwe.pdf).
```
elegaanz: This link is broken.

```suggestion
Which will generate this [PDF](https://github.com/DavZim/bootstrapicons/blob/740fc8923f809a4e4edf8c7836abf6bf568b832a/examples/mwe.pdf).
```

DavZim:
Thank you very much for the feedback!
I completely understand the concern about package size and repository performance - I really like Typst choices regarding lightweight and speed and I appreciate that this is a guiding principle for the package index.

Unfortunately, the bootstrap icons are only available as svgs and not as a font itself.
The current setup makes it highly portable but has the disadvantage of size. I thought about compressing it and the decompressing the values at runtime, but I didnt see typst capabilities for this.

The motivation behind the package was practical: I repeatedly found myself manually copying Bootstrap SVGs between projects because the set of icons in `#sym` often wasn’t sufficient for UI‑like documents, diagrams, or presentations. At some point this became frustrating enough that a proper, reusable solution seemed worthwhile — both for myself, my colleagues, and other Typst users who need a broader icon collection without constant manual asset handling.

With your constraints in mind, do you see an alternative path for making a richer icon set broadly available to users?

DavZim:
Hi @elegaanz,

thank you again for the comments and motivation to keep things small and efficient. I had a thought about how we can make this package work while keeping its functionality.

Here is what I came up with (see also updated files):

1. The raw icons (see [`raw-icons/`](https://github.com/DavZim/bootstrapicons/tree/main) directory, containing the 2058 svg files are omitted from this PR (and thus the typst/packages repository), dramatically reducing the file count of this PR/package.
2. Instead the svg content is included in one "big" `bs-icons.typ` file, with 2058 +2 lines of code, one for each icon, sitting at less than 1 MB. Comparing the first draft, we had 4.6M of data, now we are down to 983K bytes.
3. To decrease filesize of the `bs-icons.typ` file, we only store the svg contents and not the meta `<svg>` tag and its information as they are filled at (PDF-) compile time by typst in `#bsicon()`. See updated [`bsicon` function](https://github.com/typst/packages/pull/3841/files#diff-b1eeaece614a63432a84fabc5253d21b8eb4e084626a16cc50ed4fefcfd5cd37) for details, this means only the minimum information is stored.

This way we reduced the filecount from 2k+ down to 8 (including toml, Readme, License, etc) and total filesize from 4+M down to 1M bytes (all values reported by `du -sh`).
What do you think, is this small enough to be considered as a package for typst? Or do you see any further potential for optimization?

Thanks,
David

elegaanz:
Hello, sorry for the long delay without a response. After discussing this with the rest of the team, we agreed that your new approach can work. One thing I would like to ask you before we merge this package is to pick something else than "bootstrapicons". It is the canonical for that package and we try to avoid attributing those (or more exactly they are reserved for "official" packages), so that alternatives can be built without any of the packages having an advantage over the other.

DavZim:
Hi eleganz, thanks for coming back.
I will look for a fitting name and update the PR.

DavZim:
I have renamed the package to the (hopefully) less canonical name of booticons. It should still be recognizable what it does but far away from the canonical name (I hope).
Let me know if you feel otherwise or have a better name.

elegaanz:
That name works. :) Thank you!

PR #3844, opened 2026-01-13
fh-joanneum-iit-thesis:2.4.0

internettechnik:
I am submitting
- [ ] a new package
- [x] an update for a package

Added feature: draft mode
References: Support for ibid 
Also: Updated dependencies ( glossary to 0.5.10), Updated previews.


packages/preview/fh-joanneum-iit-thesis/2.4.0/typst.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "fh-joanneum-iit-thesis"
```
saecki, line 2: If this should be an official package, we need a staff member to send an email to [hello@typst.app](mailto:hello@typst.app) that endorses this package. Otherwise the name should be changed, for example by adding a non-descriptive prefix, to meet our [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules). We try to avoid too canonical names, so multiple packages serving the same purpose can coexist without one having an unfair advantage.

saecki (commented):
Thanks for this package! I've left a comment regarding the name.

packages/preview/fh-joanneum-iit-thesis/2.4.0/typst.toml
saecki: Sorry, I missed that this package is already `officially affiliated`. In that case you of course don't need to send an email again.

PR #3845, opened 2026-01-13
clean-othaw:0.3.2

ruebe5w:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fix bug with city argument


saecki:
Unfortunately we don't allow any package edits to ensure reproducibility. See #2671
You can publish a new patch version instead

PR #3846, opened 2026-01-13
tcdm:0.0.2

YDX-2147483647:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Add a missing translation in the template.


YDX-2147483647:
As in the previous PR #3748, all warnings and errors that the bot reported here are false positive.

saecki:
Thanks for the update!

PR #3853, opened 2026-01-15
flupke-headstamp:0.1.0

flol3622:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Git-trace reads Git reflog/HEAD metadata (branch, hash,        
  message, timestamp) so documents can show the exact Git state they were     
  built from. It’s useful for reproducible reports and traceable PDFs without 
  running Git commands.


packages/preview/git-trace/0.1.0/src/lib.typ
```diff
@@ -0,0 +1,108 @@
...
+    } else if s.starts-with("commit ") {
+      let parts = s.split(": ")
+      _at(parts, parts.len() - 1, default: s)
+    } else {
+      s
+    }
+  }
+}
+
+#let _git-head(git-dir: ".git") = read(git-dir + "/HEAD").trim()
```
lordkekz: Since the call to the builtin `read` function is located in your package's `lib.typ`, the path will be resolved relative to the location of this file in the local cache, not the user's call your package's functions.

Other packages such as [`alexandria`](https://typst.app/universe/package/alexandria) work around this by asking the user to supply a read function so that the actual call of Typst's builtin `read` is located in the package user's code:
```typ
#show: alexandria(prefix: "x:", read: path => read(path))
```

Edit: Same thing applies to read call in line 34.

flol3622: Thanks for the heads-up! I wired in a caller read so .git gets read from the doc context, not the package cache. Docs updated too.

packages/preview/git-trace/0.1.0/README.md
```diff
@@ -0,0 +1,44 @@
...
+## Features
+
+- Read branch and HEAD hash from `.git`.
+- Read last commit message and date from `.git/logs/HEAD`.
+- Format reflog timestamps as `dd-mm-yyyy hh:mm` (24-hour time).
+
+## Quick start
+
+```typst
+#import "@preview/git-trace:0.1.0": git-branch, git-head-hash, git-last-commit, git-format-date
```
saecki: This fails to compile:
```
error: the argument `read` is positional
   ┌─ @preview/git-trace:0.1.0/src/lib.typ:67:48
   │
67 │   let line = _git-reflog-last(git-dir: git-dir, read: read)
   │                                                 ^^^^^^^^^^
   │
   = hint: try removing `read:`

help: error occurred in this call of function `git-last-commit`
  ┌─ test.typ:4:12
  │
4 │ #let info = git-last-commit(read: git-read)
  │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

packages/preview/git-trace/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "git-trace"
```
saecki: The name feels a little too descriptive, we try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. See: https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules

flol3622: I renamed the package to flupke-headstamp to avoid a canonical name. Folder + typst.toml + README import are updated accordingly.

packages/preview/git-trace/0.1.0/README.md
flol3622: Thanks for the repro! Fixed: the internal helpers now accept read as a named parameter, so git-last-commit(read: ...) compiles cleanly. I also smoke-tested it locally with typst compile.

PR #3858, opened 2026-01-15
fitch:0.1.0

Lipen:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Fitch-style natural deduction proofs.


saecki:
Thanks for this package!

packages/preview/fitch/0.1.0/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "fitch"
```
saecki, line 2: Considering the package does fitch-notation rendering this name seems a little too canonical, could you come up with something a little different follwing the [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)?

Lipen: I do understand that `fitch` looks "canonical", since the description explicitly states that we provide Fitch-styled proofs, however Fitch notation is not the _only_ way to organize deductive proofs. There is also a tree-like notation (e.g. [curryst](https://typst.app/universe/package/curryst) package) which we do not provide, since we only focus on Fitch notation. And I should mention that curryst package is very hard to discover when I need to write the tree-like deductive proofs. Probably, that would be much easier if not that "no canonical names" rule (`tree-proofs` would be more obvious for users, but that is up to the package author anyway).

Canonical name in that case would be `proof` or `natural-deduction`.  That would be "too abstract/generic" and interfere with similar (yet distinct) search patterns regarding different types of logic proofs.

Note that in latex world we also have the [`fitch`](https://github.com/OpenLogicProject/fitch/blob/main/fitch.sty) package. And it was the main source of inspiration for my package (I should include that fact in the readme, when I get to the package documentation). That name is perfect: if I ever need the "Fitch-style proofs", surely I am going to search "fitch" in CTAN, and find exactly the right package for it.

Of course, I am going to think about some nice name for my package that do not explicitly state "fitch" in its name (only in the description), but I think that would worsen the discoverability...

Lipen: just to clarify, do the following names comply with the naming guidelines?
- `fitch-it` (akin to [`derive-it`](https://typst.app/universe/package/derive-it))
- `fitchify`
- `frederic` (Fitch's name)
- `matryoshka` (because of nested subproofs support)
- `deducto`

saecki: > Canonical name in that case would be proof or natural-deduction. That would be "too abstract/generic" and interfere with similar (yet distinct) search patterns regarding different types of logic proofs.

I see where you're coming from, but since the package and Typst mostly deal with the visual representation, I think the Fitch notation is the key part here.

> Of course, I am going to think about some nice name for my package that do not explicitly state "fitch" in its name (only in the description), but I think that would worsen the discoverability...

Ironically that's exactly the reasoning behind our naming rules. If we would allow `fitch` as a name directly, it would invite people to name-squat good package names, because they inherently would have an advantage in discoverability. When every package name is a little different, this "unfair" advantage is mitigated to some degree.

To still ensure discoverability, the keywords list will be used for searching, so someone searching `fitch` will still find you package.

> just to clarify, do the following names comply with the naming guidelines?
>
>    fitch-it (akin to [derive-it](https://typst.app/universe/package/derive-it))
    fitchify
    frederic (Fitch's name)
    matryoshka (because of nested subproofs support)
    deducto

I would say all of these work :+1:

Lipen: Okay, I think I got the idea behind the "squatting prevention".

Can you please also clarify what to do with the current branch and PR after I rename my package? Resubmit the package in a new branch in the fresh PR?

saecki: You can either just push the changes on this existing PR, the bot will automatically update the name of the PR.
Or you can create a new PR using this or another branch, and close this PR.

(also remember to rename the package directory)

PR #3859, opened 2026-01-15
simple-plot:0.2.0

nathan-ed:
I am submitting
- [x] an update for a package

#### Added
- `axis-x-extend` parameter to extend X-axis beyond plot area (symmetric or asymmetric)
- `axis-y-extend` parameter to extend Y-axis beyond plot area (symmetric or asymmetric)
- `show-origin` parameter to control display of "0" label at origin
- `label-side` parameter for function labels with 8 positioning options: "above", "below", "left", "right", "above-left", "above-right", "below-left", "below-right"
- Line segment clipping using Liang-Barsky algorithm for cleaner plot rendering
- New gallery example: `extended-axes.typ` demonstrating axis extension

#### Fixed
- Tick numbering algorithm now generates more intuitive automatic tick intervals
- Function label positioning now correctly uses visible area instead of extended sampling domain


saecki:
Could you rebase onto `main` so that only the change from `simple-plot:0.2.0` are included in this PR?

nathan-ed:
> Could you rebase onto `main` so that only the change from `simple-plot:0.2.0` are included in this PR?

https://github.com/typst/packages/pull/3891

I create a new PR here after modifying my local repo to track each package on a different branch. It was the first time I was working with "grafted" and i messed up. Thank you!

If you merge the other one, we can close this one.

PR #3861, opened 2026-01-15
taskize:0.2.0

nathan-ed:
I am submitting
- [x] a new package

Description: Horizontal columned lists for exercises and tasks, similar to LaTeX's tasks package. Items flow left-to-right across columns with flexible numbering and layout options, based on itemize package.


packages/preview/taskize/0.2.0/typst.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "tasks"
```
elegaanz: The name is too descriptive in my opnion, which we try to avoid. The goal is to avoid a package to have a SEO advantage over another. Can you pick a more unique name, that is less likely to collide with some other package with a similar purpose?

elegaanz (commented):
Looks  good, thank you for building this package! I would just like you to change the name.

packages/preview/taskize/0.2.0/typst.toml
nathan-ed: I used this name to mirror the LaTeX package https://ch.mirrors.cicku.me/ctan/macros/latex/contrib/tasks/tasks-manual.pdf which i am trying to replicate. `taskize` would be ok ?

nathan-ed: Renamed it to `taskize`. Thank you!

elegaanz:
The new name sounds good. Thank you!

PR #3862, opened 2026-01-15
lumen:0.1.3

mo-nonym:
I am submitting
- [ ] a new package
- [x] an update for a package

I added/exposed more variables to the template/main.typ file (giving users the possibility to change fonts and their size) and I updated/fixed errors in the README example. 


elegaanz:
Thanks.

PR #3882, opened 2026-01-18
tidbit-uoc:0.1.1

DanielRamosAcosta:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This release updates tidbit-uoc with improved heading numbering and link styling. The heading numbering scheme has been enhanced from a single-level format to a three-level hierarchical numbering (1.1.1), providing better document structure organization. Additionally, links are now underlined for improved visual distinction and readability.

### DIFF
--- a/packages/preview/tidbit-uoc/0.1.1/README.md
+++ b/packages/preview/tidbit-uoc/0.1.1/README.md
@@ -6,7 +6,7 @@ This template provides a clean and modern layout following UOC's visual identity
 
 ## Preview
 
-<img src="./thumbnail.png" alt="" width="300">
+<img src="./thumbnail.png" alt="Cover page of an academic document with a blue layout, showing the course name, assignment title, author name, date, and the Universitat Oberta de Catalunya" width="300">
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


packages/preview/tidbit-uoc/0.1.1/README.md
```diff
@@ -0,0 +1,131 @@
+# Tidbit UOC Template
+
+A Typst template for academic reports, practices, papers and thesis at the Universitat Oberta de Catalunya (UOC) with professional styling.
+
+This template provides a clean and modern layout following UOC's visual identity guidelines, featuring the university's branding colors and logo.
+
+## Preview
+
+<img src="./thumbnail.png" alt="" width="300">
```
elegaanz: An actual alt text here would be appreciated. It doesn't have to be very long, something like "Front page of an example document, featuring its title, the author name, date of publication and the UOC logo."

packages/preview/tidbit-uoc/0.1.1/README.md
```diff
@@ -0,0 +1,131 @@
...
+1. **Cover Page**: Subject, title, subtitle, author, and date with UOC branding
+2. **Content Pages**: Clean layout with headers, footers, and page numbering
+3. **Proper Margins**: Academic-appropriate spacing for binding and readability
+
+### Assets
+
+Make sure to include the UOC logo in your project:
+
+- Place `logo.svg` in an `assets/` directory
+- The logo will be automatically included in the cover page and headers
```
elegaanz: This is a bit misleading I think. There is no way for the template to always use what's inside of `assets/` in the user's project. What it currently does is that it always uses the logo provided by default, even if the user replaces the file with something else in their own projects.

If you want to give the possibility to change the logo, it should be passed as an argument of type `content` (a call to the `image` function typically) to your template, by the user. [See this page for more info](https://github.com/typst/packages/blob/main/docs/resources.md#paths-are-package-scoped).

packages/preview/tidbit-uoc/0.1.1/README.md
DanielRamosAcosta: Fixed!

packages/preview/tidbit-uoc/0.1.1/README.md
DanielRamosAcosta: Fixed!

elegaanz:
Thank you!

PR #3883, opened 2026-01-18
gb7714-bilingual:0.2.0

lucifer1004:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
Packages cannot read files outside its own folder, so the API is changed to accept bytes instead of file names.


elegaanz:
Thank you.

PR #3922, opened 2026-01-21
ctz-euclide:0.1.0

nathan-ed:
I am submitting
- [x] a new package

Description: A comprehensive Euclidean geometry package for Typst, a port from the LaTeX **ctz-euclide** package.

### DIFF
--- a/packages/preview/ctz-euclide/0.1.0/typst.toml
+++ b/packages/preview/ctz-euclide/0.1.0/typst.toml
@@ -10,4 +10,4 @@ keywords = ["geometry", "euclidean", "mathematics", "drawing", "cetz", "euclide"
 categories = ["visualization", "utility"]
 disciplines = ["mathematics", "education"]
 compiler = "0.11.0"
-exclude = ["tests/*", "docs/*", "latex/*", "*.pdf", "*.md"]
+exclude = ["tests/*", "docs/*", "latex/*", "*.pdf"]


packages/preview/ctz-euclide/0.1.0/src/lib.typ
```diff
@@ -0,0 +1,1040 @@
...
+  ctx.shared-state.at(_points-key, default: (:))
+}
+
+/// Store points in context
+#let _set-points(ctx, points) = {
+  ctx.shared-state.insert(_points-key, points)
+  ctx
+}
+
+/// Custom coordinate resolver - accepts both "A" and "tkz:A" syntax
```
elegaanz, line 33: Any reason to use the `tkz` prefix? I feel like it would be confusing.

packages/preview/ctz-euclide/0.1.0/src/lib.typ
```diff
@@ -0,0 +1,1040 @@
...
+    let points = _get-points(ctx)
+    points.insert(name1, result.at(0))
+    points.insert(name2, result.at(1))
+    ctx = _set-points(ctx, points)
+    (ctx: ctx)
+  },)
+}
+
+/// Create wrapper functions that have cetz modules pre-bound
+#let create-api(cetz) = {
```
elegaanz, line 800: I feel like this pattern of dependency injection is not very common in Typst packages. Here, I feel like it makes your call more complex than it needs to be. The user API is also not so nice to use because all functions have to be wrapped in parenthesis because the module is built dynamically as a dictionary. Is there a particular reason for this approach?

packages/preview/ctz-euclide/0.1.0/src/lib.typ
```diff
@@ -0,0 +1,1040 @@
...
+  let cetz-coordinate = cetz.coordinate
+  let cetz-draw-module = cetz.draw
+
+  // Helper to convert point name to tkz: coordinate
+  let pt(name) = "tkz:" + name
```
elegaanz, line 805: Same remark about the `tkz:` prefix here.

packages/preview/ctz-euclide/0.1.0/README.md
```diff
@@ -0,0 +1,238 @@
+# ctz-euclide
+
+A comprehensive Euclidean geometry package for Typst, inspired by the LaTeX **ctz-euclide** package.
+
+> **Note:** This package is a port of the LaTeX [ctz-euclide](https://ctan.org/pkg/ctz-euclide) package. We are deeply grateful to Alain Matthes for creating the original ctz-euclide package, without which this Typst port would have been impossible.
```
elegaanz: I think you meant tkz-euclide here.

nathan-ed:
> I left a few comments, mostly about some design choices. tkz-euclide is also always called ctz-euclide. If you use AI to code, please review its work before submitting a PR, I think these mistakes could have been caught easily.

My bad. This was more done by me with search and replace when I decided the package name. 
I'll change everything more carefully tomorrow. Thanks for your careful and precious review.

nathan-ed:
I'll work a bit more on it, before submitting again.

PR #3936, opened 2026-01-22
manifesto:0.1.0

l0uisgrange:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package is the first HTML template for Typst, written 100% in Typst using native HTML export. It works the same as a PDF template, but produces a stunning, HTML website ready for online package documentations, tutorials, etc...

### DIFF
--- a/packages/preview/manifesto/0.1.0/typst.toml
+++ b/packages/preview/manifesto/0.1.0/typst.toml
@@ -8,7 +8,7 @@ categories = ["layout", "visualization"]
 disciplines = ["design", "education"]
 keywords = ["documentation", "docs", "html", "website", "site", "static", "online"]
 exclude = [".github"]
-entrypoint = "/src/lib.typ"
+entrypoint = "src/lib.typ"
 authors = ["Louis Grange <https://github.com/l0uisgrange>"]
 license = "MIT"
 


l0uisgrange:
Hey there, since this template is the first for HTML (instead of PDF), what is the way to publish it? Did I do it correctly?
I'm very sorry for the inconvenience 😅, I already have 2 packages but this is my first template.

Also, the checks do not seem to pass because of `html` missing (available since Typst 0.13 at least, but behind a flag) ... is my project not suitable for GitHub Universe because needs users need a flag? 🥲

packages/preview/manifesto/0.1.0/README.md
```diff
@@ -0,0 +1,43 @@
...
+<img width="1232" height="758" alt="Capture d’écran 2026-01-22 à 18 39 15" src="https://github.com/user-attachments/assets/076e3071-6feb-4453-8b44-e5de1a32ec53" />
+
+## Quick usage
+
+The template is very easy to use, and will transform your whole project into a stunning online documentation in just a second.
+
+```typst
+#import "@preview/manifesto:0.1.0"
+
+#show: it => template(it, title: "MyAwesomePackage", toml: "PATH_TO/typst.toml")
```
elegaanz: I didn't test but I think `template` is not in scope with the current import, it would be `manifesto.template`.

packages/preview/manifesto/0.1.0/README.md
```diff
@@ -0,0 +1,43 @@
+# Manifesto for Typst
+
+**Manifesto** 📚 is the first HTML Typst documentation template for packages, built to be very easy to use. It is **100% written in Typst, responsive** and directly uses your `typst.toml` to automatically get your package information.
+
+[Report a bug](https://github.com/l0uisgrange/manifesto/issues?new) — [Forum](https://github.com/l0uisgrange/manifesto/discussions/categories/q-a)
+
+You are free to check either [this example](https://bone.grangelouis.ch) from `bone` or [this one](https://zap.grangelouis.ch) from `zap`.
+
+<img width="1232" height="758" alt="Capture d’écran 2026-01-22 à 18 39 15" src="https://github.com/user-attachments/assets/076e3071-6feb-4453-8b44-e5de1a32ec53" />
```
elegaanz: Can you write an actual alternative text please?

packages/preview/manifesto/0.1.0/typst.toml
```diff
@@ -0,0 +1,18 @@
...
+categories = ["layout", "visualization"]
+disciplines = ["design", "education"]
+keywords = ["documentation", "docs", "html", "website", "site", "static", "online"]
+exclude = [".github"]
+entrypoint = "template.typ"
+authors = ["Louis Grange <https://github.com/l0uisgrange>"]
+license = "MIT"
+
+[template]
+path = ""
```
elegaanz: Your template should be in its own subdirectory. Only files in that directory will be copied to the user project. The "library" part of your package should contain utility and templating functions you don't expect the user to modify, while the template files should contain some default content.

packages/preview/manifesto/0.1.0/typst.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "manifesto"
+version = "0.1.0"
+compiler = "0.14.2"
```
elegaanz: This corresponds to the lowest version number this package works with, so unless you depend on a fix that was introduced in 0.14.2, it should probably be 0.14.0 or even lower.

packages/preview/manifesto/0.1.0/template.typ
```diff
@@ -0,0 +1,191 @@
...
+
+#let template(doc, title: none, toml: none, copyright: true, version: none, notices: (), links: (), description: none, repository: none, universe: none, license: none) = [
+
+    #let title = if toml != none { detoml(toml).package.name } else { title }
+    #let version = if toml != none { detoml(toml).package.at("version", default: none) } else { version }
+    #let universe = if toml != none { detoml(toml).package.at("name", default: none) } else { universe }
+    #let license = if toml != none { detoml(toml).package.at("license", default: none) } else { license }
+    #let repository = if toml != none { detoml(toml).package.at("repository", default: none) } else { repository }
+    #let description = if toml != none { detoml(toml).package.at("description", default: none) } else { description }
+    #html.elem("html", attrs: (lang: "en", class: "scroll-smooth"))[
```
elegaanz: If you are using the untyped `html.elem` API you can support Typst 0.13, but if you want to only support 0.14 you could use [the typed HTML](https://typst.app/docs/reference/html/typed/) API.

packages/preview/manifesto/0.1.0/src/icons.typ
```diff
@@ -0,0 +1,30 @@
+#let qa-icon = html.elem(
```
elegaanz, line 1: I think it would be nice to provide some fallback when compiling to PDF, even if the template is supposed to be HTML-only. PDF export is still the default and it would be a bad experience to get a non-working template by default. Simply displaying a non-editable page that says "This template is HTML-only" (with a bit of styling if you feel like it) would be enough. You would also have to take care of not using `html` at the top-level like you are doing here, but to always wrap it behind a conditional or a function that is lazily evaluated.

If what I'm saying is not clear I can try to give a more explicit example, let me know.

elegaanz:
Hey, do you think you will have time to make the requested updates? No worries if not, but in that case I will close the PR for now, to keep our review backlog clear. You can always re-open a PR later, once you made the requested changes.

l0uisgrange:
Hey @elegaanz, will update this today, I've prepared a new version to fix the issues you mentioned. Thanks!

packages/preview/manifesto/0.1.0/src/icons.typ
l0uisgrange: I totally agree, but I can't seem to be able to display some text only when PDF-exporting (and not display it on HTML)...do you have an example by any chance?
The same for the top-level icons (can't manage to export them only if html export option activated)

elegaanz: What I was thinking about when writing this comment was to wrap these expressions in `context` and use [`target`](https://typst.app/docs/reference/foundations/target/). But maybe that's not ergonomic at all or there are some issues I didn't think of. If that's the case, feel free to ignore this comment.

l0uisgrange: Thanks, didn't know about `target`. Just updated the template to show the warning you mentioned.

l0uisgrange:
I have finished updating the template, but the tests don't pass... am I missing something? 😇

l0uisgrange:
@elegaanz checks now pass, you can merge whenever you want 😇 thanks for the help, appreciate it!

elegaanz:
Thank you :)

PR #3937, opened 2026-01-22
swaralipi:0.1.0

abirbasak:
I am submitting
- [x] a new package
- [ ] an update for a package

Swaralipi is a Typst package for typesetting Indian Classical and Semi-Classical music notations. `Swaralipi` allows you to write music notations and lyrics using regular English letters and renders them into a beautiful rhythmic grid (Taal).

### DIFF
--- a/packages/preview/swaralipi/0.1.0/README.md
+++ b/packages/preview/swaralipi/0.1.0/README.md
@@ -37,12 +37,10 @@ Import the package into your Typst file:
 
 ## Usage
 
-You can use the `render_composition` function or set up a show rule for raw blocks to use the `note` language tag.
+You can use the `apply-swaralipi` function or set up a show rule for raw blocks to use the `note` language tag.
 
 ```typst
-#show raw: it => if it.lang == "note" {
-  render_composition(it.text, it.lang)
-} else { it }
+#show raw: apply-swaralipi
 ```
 
 ## Examples
--- a/packages/preview/swaralipi/0.1.0/examples/song.typ
+++ b/packages/preview/swaralipi/0.1.0/examples/song.typ
@@ -1,7 +1,7 @@
 
-#import "@preview/swaralipi:0.1.0": render-composition
+#import "@preview/swaralipi:0.1.0": apply-swaralipi
 
-#show raw: it => if "note" in it.lang { render-composition(it.text, it.lang) } else { it }
+#show : apply-swaralipi
 
 ```note[taal: kaharba,wrapped : false,matra : false]
                       |                |           | -  - P - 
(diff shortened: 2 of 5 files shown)


packages/preview/swaralipi/0.1.0/README.md
```diff
@@ -0,0 +1,94 @@
...
+- **Lower Octave**: Add a dot `.` after the note (e.g., `S.`, `N..`).
+- **Higher Octave**: Add a single quote `'` after the note (e.g., `S'`, `G''`).
+
+### Ornaments
+- **Meend (Slur)**: Wrap notes in parentheses: `(S R G)`
+- **Kan (Grace Note)**: Wrap grace notes in curly braces before the main note: `{R}G`
+
+## Usage
+
+You can use the `render_composition` function or set up a show rule for raw blocks to use the `note` language tag.
```
elegaanz: If you expect everyone to use this exact show rule, you could provide it as a function directly, so that people can just do `show: apply-swaralipi`. With `apply-swaralipi` defined as something like that.

```typst
#let apply-swaralipi = (body) => {
  show raw: it => if it.lang == "note" {
    render_composition(it.text, it.lang)
  } else { it }

  body
}
```

elegaanz: Also the `render-composition` function has been renamed to use kebab-case and should be updated in this example too.

packages/preview/swaralipi/0.1.0/typst.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "swaralipi"
```
elegaanz, line 2: As far as I understand, "Swaralipi" is the canonical name for that music notation. If so, I would like to ask you to pick another name for the package. We want to avoid such kind of names, to let alternative packages co-exist, without any of them having an advantage over the other by looking more "official".
You could pick a name that is tangentially related to classical Indian music, that would work for example.

If I'm mistaken and "Swaralipi" is actually something else (I'm unfortunately not very knowledgeable on the topic) this name should be fine.

packages/preview/swaralipi/0.1.0/README.md
abirbasak: These changes are done, and  README updated accordingly. Thanks for the suggestion

packages/preview/swaralipi/0.1.0/typst.toml
abirbasak: "Swaralipi" is a descriptive compound word for pitch (swar) and script (lipi) and is not a canonical title; other authors remain free to use alternative terms like "Sargam" (which is the most commonly used term for music notations)  ,"swaramalika","swarabitan"  etc.  As it is already used in several places, including repo name, documentation etc, I would prefer to maintain this name for clarity and consistency unless it violates a specific policy.

elegaanz: Okay, then I don't think it violates any of our rules, you can keep using that name :)

elegaanz:
Looks good, thank you!

PR #3962, opened 2026-01-24
citrus:0.1.0

lucifer1004:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Full CSL + Major CSL-M implementation in pure Typst, supporting both BibTeX and CSL-JSON input.

	- I think `citeproc` is the most suitable name for this package since it is a direct implementation just like `citeproc-js`.


packages/preview/citeproc/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "citeproc"
```
elegaanz: The name is too generic to meet our guidelines in my opinion, as most citation formatting programs are called "citeproc". Can you make it a bit more unique, please? Either by changing it entirely, or by adding a distinctive prefix to "citeproc".

lucifer1004:
> I'm not sure what this package brings compared to the standard CSL support in Typst, I think it would be nice to highlight the differences a bit more in your README. Still, I'm impressed that you implemented all that in pure Typst!

This package has [CSL-M](https://citeproc-js.readthedocs.io/en/latest/csl-m/index.html) support, as I already highlighted.
Also, it can directly take CSL-JSON as input, while the embedded `bibliography` takes `.bib` or `.yaml` only.

It can be seen as a generalization of [gb7714-bilingual](https://typst.app/universe/package/gb7714-bilingual). `gb7714-bilingual` only supports GB/T-7714 standards, while this new package can support the majority of CSL-M styles.

packages/preview/citeproc/0.1.0/typst.toml
lucifer1004: I could not think of a better name.
It was previously called `citeproc-typst` (which shows that it is a pure-Typst implementation of citeproc), but that name could not even pass the CI check.

lucifer1004: Existing packages in other languages:

- https://hackage.haskell.org/package/citeproc-hs
- https://github.com/Juris-M/citeproc-js
- https://pypi.org/project/citeproc-py

So a pure Typst implementation of the CSL spec being called `citeproc-typst` is reasonable, I think.

lucifer1004: What do you think?

elegaanz: The thing is, Typst Universe has its own naming guidelines, and what would make a good name in another package ecosystem does not necessarily work here.

Using `-typst` in package names is considered redundant (it should be obvious that this a Typst package we are talking about, both on the Universe website, or in a Typst document). And as I said, canonical names are discouraged.

To give you an example: a package to manipulate images called `image-edition` would not be allowed. Let's say someone also needs to manipulate images but they don't like the API provided by that package so they build and publish their own. Now, what do they call it? The "obvious" name is already taken, and no matter what they choose, people will more easily find the other package.

To get back to this specific package, `citeproc` is (as far as I know) the name of most citation formatting programs and libraries. This will be the name any package for that purpose will go for by default, and what people looking for that functionality will search for and use. Granting you the right to use that name would give you an unfair advantage over other potential implementations.

I hope this makes my opinion and the problem I'm trying to avoid clearer.

elegaanz:
> > I'm not sure what this package brings compared to the standard CSL support in Typst, I think it would be nice to highlight the differences a bit more in your README. Still, I'm impressed that you implemented all that in pure Typst!
> 
> This package has [CSL-M](https://citeproc-js.readthedocs.io/en/latest/csl-m/index.html) support, as I already highlighted. Also, it can directly take CSL-JSON as input, while the embedded `bibliography` takes `.bib` or `.yaml` only.
> 
> It can be seen as a generalization of [gb7714-bilingual](https://typst.app/universe/package/gb7714-bilingual). `gb7714-bilingual` only supports GB/T-7714 standards, while this new package can support the majority of CSL-M styles.

Sorry, I skipped that part of your PR. Thank you for the clarification.

packages/preview/citeproc/0.1.0/typst.toml
lucifer1004: Rebranded to `citrus`

lucifer1004:
@elegaanz I have renamed the package to `citrus`.

elegaanz:
Nice, the new name sounds good :) Thank you!

zepinglee:
I'm just curious. Did you test this package against the test suite at <https://github.com/citation-style-language/test-suite>?

PR #3963, opened 2026-01-25
nexus-tools:0.1.0

mayconfmelo:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:

For me, this package is important because it's a kind of library for resources shared by all [my other packages](https://typst.app/universe/search/?q=author%3A%22Maycon%20F.%20Melo%22); and for anyone else, this is a handy
development aid with a select collection of ready-to-use and well documented
features.
 
I often find myself having to implement the same internal functionalities in
different packages (e.g., data storage, custom defaults, etc.), always with some
residual differences between each implementation; this dynamic makes code
maintenance difficult, repetitive, and hinders development speed. All of this is
handled by _toolbox_, which centralizes and organizes the development of these
features, while keeping the modifications as documented versions.

I've already updated my packages to use this library from now on, and I'm just
waiting for _toolbox_ to be published on Typst Universe to publish them as well.


packages/preview/toolbox/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "toolbox"
```
elegaanz: Because the package name is bit too generic, many other packages may want to use it. To not give an unfair advantage to any of the possible implementations, our guidelines forbid such generic names. Can you pick another name, please? It could be `min-toolbox` for example, if you want to keep a consistent prefix with your other packages. Adding this prefix would make it distinctive enough. But it can also be anything else, as long as it's not too generic.

mayconfmelo: No problem! Would the name `dev-toolbox` still be too generic? I'm not very creative with names.

The prefix `min-` might give the impression that this package is related only to my templates; my packages don't usually follow this naming convention (e.g., `transl`).

elegaanz: I feel like dev-toolbox would be a bit too generic, yes. Package authors generally pick a unique adjective to put in front of an otherwise generic name, which is an easy way to meet our guidelines. For package names, using a name that is tangentially related with the topic at hand is also a common strategy (for example `grayness` for an image manipulation package), but it may not be easy to find a good one here, since the package is quite "abstract". Otherwise, a pun or an obviously misspelled word can be accepted.

mayconfmelo: I figured it was still too generic, I'm terrible at that. Maybe something like `devhub`, `nexus-tools` or `tool-nexus` would work better?

elegaanz: `nexus-tools` is the best-sounding one according to me, but the two others could work too.

mayconfmelo: I'll trust your judgment then, `nexus-tools` it is. Thank you!

elegaanz:
Thank you!

PR #3967, opened 2026-01-25
eggs:0.4.1

retroflexivity:
I am submitting
- [ ] a new package
- [x] an update for a package

- Footnotes now use a separate counter instead of resetting the main one to prevent non-convergence cases
- More detailed errors for glosses


elegaanz:
Thank you.

PR #3974, opened 2026-01-26
fine-lncs:0.4.0

Jozott00:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Add an optional `running-title` template parameter and an example theorem block.


saecki:
Thanks!

PR #3977, opened 2026-01-26
pc-letter:0.4.0

thatfloflo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

#### Added

- New Italian localisation and example (see [examples/example-it.typ](./examples/example-it.typ)) (thanks to Biagio and Giulia)
- New vertical alignment options (`top` \[default\], `horizon`, `bottom`) for the `address-field`.

#### Changes

- The implementation of vertical alignment for the `address-field` has changed subtly along with the implementation of new alignment options. This should normally not be noticeable when migrating, unless the template style was heavily customised beforehand.
- Improved Readme

--- 


packages/preview/pc-letter/0.4.0/examples/example-cy.pdf
saecki, line 1: Could you remove the example PDF files? They contribute to repository bloat and can easily be reproduced by compiling the Typst files. You also seem to be linking the `examples` directory from your own repository anyway, so this shouldn't affect users :)

thatfloflo: Yes of course, I've removed them now, thanks!

thatfloflo:
Fixed, thanks @saecki !

saecki:
Thanks!

PR #3981, opened 2026-01-27
modern-sjtu-thesis:0.5.2

tzhtaylor:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- Add support for custom fields on the Chinese cover page
- Optimize cover layout spacing and field display logic
- Improve reference processing logic
- Fix an issue where “continued table” labels did not display correctly for tables spanning more than three pages


saecki:
Thanks!

PR #3991, opened 2026-01-28
pf2e-style:0.2.0

WiggyJiggyJedHed:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Pathfinder 2nd Edition Remastered template. Adds styling, functions, and icons to enable third party and independent role-playing game material creation using Typst. **Better optimized functions and styling**

### Changelog

- Created an `encounter` function which acts as a single, more simplistic way to create stat blocks, hazards, complications, and whatever else an adventuring party might run into
- Turned the operation to stylize traits into a single function, `pftraits` to reduce redundancy
- Fixed `chap-header` so that it attaches to the head of the page
- Condensed `feat` and `spell` so that they have similar syntax
- Adjusted the spacing in `aloud` to look cleaner
- Corrected the Paizo license from Fan Content to Community Use

### DIFF
--- a/packages/preview/pf2e-style/0.2.0/LICENSE
+++ b/packages/preview/pf2e-style/0.2.0/LICENSE
@@ -1,0 +1,16 @@
+MIT No Attribution
+
+Copyright 2026 Wexel Wingding
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this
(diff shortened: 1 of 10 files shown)


elegaanz:
Thanks!

PR #4000, opened 2026-01-29
slipst:0.2.0

Wybxc:
I am submitting
- [ ] a new package
- [x] an update for a package

**Changelog**:
*   The generated HTML no longer depends on CDN resources, enabling fully offline use.
*   Progress is now controlled and persisted via the URL hash instead of session storage, allowing slides to be shared or bookmarked directly.
*   Spacing between slips and page margins now scale with the screen size and can be customized in the `show` rule.
*   Added support for gesture navigation (swipe up/down) on touch devices.


elegaanz:
Great, thank you!

PR #4005, opened 2026-01-29
noteworthy:0.3.0

tarunjana:
I am submitting
- [ ] a new package
- [x] an update for a package

Noteworthy is a typst template for creating class notes especially for Mathematics.


elegaanz:
Thank you!

PR #4007, opened 2026-01-29
elegant-culsc-record:0.11.0

SchrodingerBlume:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A template for experiment records of the China Undergraduate Life Sciences Contest (Scientific Exploration).

Version numbering starts at 0.11.0 to match this year's contest (11th session), enabling users to easily determine the correct version for their participation year.


SchrodingerBlume:
Regarding the naming convention warnings, I've chosen to keep the current naming unchanged for the following reasons:

1. This template overrides and modifies the styles of `Re` and `Im`.

2. TeX and LaTeX are proper nouns with specific capitalization requirements. Donald Knuth explicitly defined the typesetting format for TeX, and changing them to `tex` and `latex` would lose their identity.

packages/preview/culsc-record/0.11.0/typst.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "culsc-record"
```
elegaanz: I feel like this name is too descriptive, [which we try to avoid](https://github.com/typst/packages/blob/0488b38bb7fdfbce81981803ebbcf5612f8919f8/docs/manifest.md#naming-rules) for non-official packages. Can you change it to fit our naming rules? You can simply add a distinctive word to the current name, or pick an entirely new one that is less descriptive. The goal is to make it clear that this is not the official template, and to not give you an unfair advantage in case someone wants to publish an alternative template for the same purpose.

elegaanz (commented):
Thank you for this template :) No problem for the naming convention, it makes perfect sense to keep these names as they are in this context. Overall, the package looks good, but the name you choose does not match our guidelines. Once you will have renamed it, it should be good to be published. If you are unsure whether a name would be a good fit, feel free to ask first.

packages/preview/culsc-record/0.11.0/typst.toml
SchrodingerBlume: I've added an "elegant-" prefix, which should fit the guidelines better now! ; )

elegaanz:
Thank you!

PR #4013, opened 2026-01-30
pariman:0.1.0

pacaunt:
I am submitting
- [x] a new package
- [ ] an update for a package

This package allows users to show engineering calculations involving multiple units easily. Each number is packed in an element that can follow the unit cancellation, combination, and can even show the unit conversion effortlessly. 

### DIFF
--- a/packages/preview/pariman/0.1.0/src/quantity.typ
+++ b/packages/preview/pariman/0.1.0/src/quantity.typ
@@ -1,6 +1,6 @@
 #import "utils.typ"
 /// The formatting functionality is provided by zero package.
-#import "@preview/zero:0.6.0" as zero: zi
+#import "@preview/zero:0.6.1" as zero: zi
 
 /// Parser for unit input
 /// The possible patterns:
@@ -77,7 +77,8 @@
   }
   let mantissa = calc.round(num / calc.pow(10, N), digits: figures)
   let formatted = str(mantissa) + if N != 0 { "e" + str(N) }
-  formatted.replace(sym.minus, "-")
+
+  return formatted.replace(sym.minus, "-")
 }
 
 #let info-num(value, figures: auto, places: auto) = {


packages/preview/pariman/0.1.0/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+```typst
+#let a = quantity("1.0e5", "m/s^2")
+Display value and unit: #a.display \
+Display only the formatted value: #a.show \
+Display the raw value verbatim: #a.text \
+Significant figures: #a.figures \
+Decimal places: #a.places
+```
+<img alt="img1" src="https://github.com/pacaunt/pariman/blob/main/docs/image1.png"/>
```
elegaanz: An actual alternative text would be appreciated (doesn't have to be very long, but it should convey the general meaning of the image). Also, it is better to use permalinks when linking to your own repository (i.e use a git commit hash or tag name, and not `main`) so that the link stays valid forever, even if you push changes to your repository in the future.

packages/preview/pariman/0.1.0/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+Pariman's `quantity` takes care of the significant figure calculations and unit formatting automatically. The unit formatting functionality is provided by the [zero](https://github.com/Mc-Zen/zero.git) package. Therefore, the format options for the unit can be used.
+
+```typst
+#let b = quantity("1234.56", "kg m/s^2")
+The formatted value and unit: #b.display  \
+#zero.set-unit(fraction: "fraction")
+After new fraction mode: #b.display
+```
+<img alt="img2" src="https://github.com/pacaunt/pariman/blob/main/docs/image2.png"/>
```
elegaanz: Same applies here and to all other images in the README.

packages/preview/pariman/0.1.0/README.md
```diff
@@ -0,0 +1,168 @@
+# pariman
+A Framework on Engineering Calculation With Units in Typst. Featured with unit and number formatting by [zero](https://github.com/Mc-Zen/zero.git) package.
+
+ 
+# Installation
```
elegaanz: For info, level-1 headings are stripped from READMEs on Typst Universe (because there should be only one of them on the page, and our layout already uses it to display the package title). You probably want to update all your headings to be one level deeper. Keeping the `pariman` one at the beginning as a top-level one is fine, as it allows to show it correctly on GitHub while not duplicating information on Universe — where the package name is already shown.

pacaunt:
Thank you for your suggestion! 
I have updated the alternative text for the images and changed their links to permalinks. The first-level headings also become second-level or lower, leaving only the first one, which displays the package name unchanged.  Along the way, I noticed a bug, so I fixed it and updated the package dependencies.

elegaanz:
Perfect, thank you!

PR #4021, opened 2026-02-01
datify:1.0.1

Jeomhps:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed incorrect parsing of `yy` token, which previously produced 20262026 instead of 2026. Added a new token and updated the .typ file formatting. Added corresponding test cases for the short pattern.


elegaanz:
Thanks!

PR #4022, opened 2026-02-01
navigator:0.1.1

eusebe:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added `top-padding` option to `render-transition` for better control of roadmap vertical positioning.


saecki (commented):
Thanks for the update! Seems like you're missing quotes on the imports inside examples/docs.

eusebe:
Hi @saecki, 
Oh sorry! It should be fixed now.
Thank you very much!

PR #4039, opened 2026-02-04
ambivalent-amcis:0.1.1

rschuetzler:
I am submitting
- [x] an update for a package

Description: Fixes appearance of acknowledgements section if none exists


saecki:
Thanks!

PR #4042, opened 2026-02-04
catppuccin:1.1.0

TimeTravelPenguin:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Minor change to expose `set-code-theme` to allow users to use catppuccin code highlighting without altering any other styles of the document.


saecki:
Thanks!

PR #4050, opened 2026-02-05
decasify:0.11.3

alerque:
I am submitting
- [x] an update for a package

----

Relevant to the Typst plugin, the upstream release adds a fix to avoid recursing into math or non-text contexts when iterating over strings.

In irrelevant news the release highlight is support for Lua 5.5.


elegaanz:
Thank you.

PR #4059, opened 2026-02-07
tieffiction:0.2.0

Tiefseetauchner:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Primarily cleanup tasks, but added the ability for subtitles to be set and more header/footer possibilities

### DIFF
--- a/packages/preview/tieffiction/0.2.0/template/main.typ
+++ b/packages/preview/tieffiction/0.2.0/template/main.typ
@@ -1,8 +1,5 @@
--.
-02l
 #import "@preview/tieffiction:0.2.0": book, start-before-main, start-main
 
-=pl0}'
 #show: book.with(
   title: "The Wonderful House",
   author: ("Jochen Garak", "Lena Tauchner"),


packages/preview/tieffiction/0.2.0/template/main.typ
```diff
@@ -0,0 +1,31 @@
+-.
+02l
```
elegaanz: Is that supposed to be here or is it a mistake?

packages/preview/tieffiction/0.2.0/template/main.typ
```diff
@@ -0,0 +1,31 @@
...
+#import "@preview/tieffiction:0.2.0": book, start-before-main, start-main
+
+=pl0}'
```
elegaanz: Same here.

packages/preview/tieffiction/0.2.0/template/main.typ
Tiefseetauchner: Not intentional, no idea how *that* happened, fixed

elegaanz:
Thanks!

PR #4064, opened 2026-02-08
transl:0.2.0

mayconfmelo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- Internal re-design (using `nexus-tools:0.1.0`)
- Removed: `#fluent` command
- Removed: `#std` command
- Added: Fluent YAML database
- Added: Support for `{$arg}` placeables in standard database values
- Added: `#transl(expression)` can be regular expression strings
- Added: Database fallback system
  1. Searches expression in standard database
  2. Searches expression as regex in standard database
  3. Searches expression as identifier in Fluent database
  4. Panics
- Added: Trick to use Fluent database in `#show` rules
  - Get show rule pattern from standard database
  - Get translation string from Fluent database


elegaanz:
Thank you.

PR #4066, opened 2026-02-08
slate-geometry:0.1.0

towoe:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Collection of screen sizes for tablets. Useful when creating documents for non standard sized devices.

### DIFF
--- a/packages/preview/slate-geometry/0.1.0/LICENSE
+++ b/packages/preview/slate-geometry/0.1.0/LICENSE
@@ -1,0 +1,7 @@
+Copyright 2026 Tobias Wölfel
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the “Software”), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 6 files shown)


packages/preview/paper-specs/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "paper-specs"
```
saecki: This name is a little generic and could be misinterpreted as a list of paper specifications (A4 etc.). We try to avoid too canonical, so multiple packages serving the same purpose can coexist. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

saecki (commented):
Thanks for this package! I've left a comment regarding the name.

packages/preview/paper-specs/0.1.0/typst.toml
towoe: Thanks for the feedback. Would you think something like 'epaper-spec' would still be too generic? 
I will think about some other alternatives as well.

towoe: Would appreciate feedback on the suitability of a few name suggestions:
- "epaper-specs"
- "tablet-specs"
- "digital-sheets"
- "slate-sizes"
First one which is okay I will pick :)
Thanks!

saecki: I would say the "epaper-specs" and "tablet-specs" are too descriptive. With "digital-sheets" I'm also not quite comfortable, it's not *that* descriptive for this package, but I would assume it's a package for creating exercise sheets. "slate-sizes" would be more of the sort of name we want to see, you can pick this one :)

saecki:
Thanks!

PR #4069, opened 2026-02-08
clean-cnam-template:1.6.2

TomPlanche:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

Update `clean-cnam-template` to version 1.6.2.

### Added
- **`my-block` body styling**: New `body-style` parameter for body text customization (accepts `size`, `weight`, `fill`, `font` keys with `auto` defaults)
- **Font option in `title-style`**: Added `font` key to `title-style` dicts in `my-block` and `code` components
- **Math component text styling**: New `title-style` and `body-style` parameters for `definition`, `example`, and `theorem` (each accepts `size`, `weight`, `fill`, `font`
with `auto` defaults)

### Changed
- **Math components refactored**: `definition`, `example`, and `theorem` converted from `.with()` wrappers to proper functions with explicit parameters (`title`, `fill`,
`stroke`, `radius`, `inset`, `breakable`, `title-style`, `body-style`)

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through
normal use.


saecki:
Thanks!

PR #4072, opened 2026-02-09
simplebnf:0.1.2

Zeta611:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Allow multiple syntactic forms in Or (Thanks to @shilangyu)


Zeta611:
I have resolved the "homepage" field error, and I'd prefer to keep the rest of the code as-is.

saecki:
> I have resolved the "homepage" field error, and I'd prefer to keep the rest of the code as-is.

That's ok, they are just warnings :+1:  Thanks for the update!

PR #4077, opened 2026-02-09
clearly-hm:0.1.0

towoe:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Unofficial presentation slide theme for Hochschule München.


packages/preview/hm-polylux/0.1.0/typst.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "hm-polylux"
+version = "0.1.0"
+entrypoint = "hm-polylux.typ"
```
saecki: Since this is an unofficial template, you'll need to add some sort of non-descriptive prefix to the package name. That way multiple packages serving the same purpose can coexist without one having an unfair advantage. See our [naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

towoe: Always the naming :)
I thought it is only forbidden to use the single abbreviation, in this case "hm".
And with the combination of "polylux" this already makes it not the singular case, because "polylux" is nothing standard. (So 'hm-slides' would be not allowed.)
There are other HM related templates like ("scribbling-hm", "supercharged-hm"), so I guess it is not allowed to associate with the "polylux" package?
Would be great if you could let me know where my logical error is so I can avoid it in the future.

saecki: Since the `polylux` suffix is also just a purely descriptive property, it doesn't really help with disambiguation. Say the university would want to create an official package to create presentations with polylux. It would probably be named `hm-polylux` :)
So for non-official packages there should be some sort of non-descriptive part to the name that sets it apart from the official and other packages that serve the same purpose. It could also be something entirely abstract/creative like the `meander` package.

EDIT: Maybe we can expand a little bit on that in the docs

towoe: Thanks for the elaboration, I was just looking for some inspiration in the existing packages.
There are things like "haw-hamburg[-.*]", "sns-polylux-template", "shuimu-touying", which are kind similar to what I proposed.
So I kind of don't really understand why this here is not allowed.

saecki: I just checked in with the rest of the team. The `haw-hamburg` and `sns-polylux-template` packages were merged before the rules were in place/slipped through review, and wouldn't be accepted (intentionally) today. And the `shuimu-touying` package actually changed the originally proposed name.
It's always possible that we miss something during package reviews, but we try to adhere to the rules I've explained above.

towoe: Thanks for getting back with this information.
It might be helpful if the documentation was clearer on this, also mentioning that there was a time when package naming was different. Because the situation is now that earlier packages had an advantage, which is what the guidelines now state to prevent (obviously there is now way around it, but just to know that there are some names which are not good examples for new submissions).
And btw, those were just a few examples, I didn't want to pick on anything specific. I also don't want to fight on this, it was just all a bit unclear. Also I might not have been too familiar in which way `polylux` is descriptive.
Anyway, thanks for the feedback.
I will see if I can come up with a good name which is not just "non" :)

towoe: Would be great if you could comment on the suggestions:
- "polylux-hm-theme"
- "non-hm-polylux"
- "presentable-hm"
- "simply-hm"

saecki: Regarding "polylux-hm-theme", `theme` is too descriptive.
All other would work :+1:

towoe: Inspired from 'simply-hm' I am now going with 'clearly-hm', I hope this is now okay.

saecki: Thanks!

PR #4082, opened 2026-02-10
adequate-iso-7000:0.1.1

gauravmm:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Adds support for ISO 7000 icons (used for every conceivable area of engineering).

I processed all the vector icons off Wikimedia, cleaned and compressed them, and have packaged them up for easy use. (We needed these for some medical device packaging.)

Unfortunately, there doesn't seem to be a way to cleanly compress or pack these SVGs together. If you have any ideas on how I could potentially squeeze these SVGs into fewer files, please let me know!

### DIFF
--- a/packages/preview/adequate-iso-7000/0.1.1/typst.toml
+++ b/packages/preview/adequate-iso-7000/0.1.1/typst.toml
@@ -1,6 +1,6 @@
 [package]
 name = "adequate-iso-7000"
-version = "0.1.0"
+version = "0.1.1"
 repository = "https://github.com/gauravmm/typst-iso-7000"
 license = "MIT"
 entrypoint = "src/lib.typ"


saecki:
Thanks for this package!

> Unfortunately, there doesn't seem to be a way to cleanly compress or pack these SVGs together. If you have any ideas on how I could potentially squeeze these SVGs into fewer files, please let me know!

You could bundle them up into one file, similar to what `octique` and `booticons` do.

The package name is a little too descriptive. We try to avoid canonical names for packages so multiple packages serving the same purpose can coexist without any one having an unfair advantage. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

gauravmm:
I've bundled it similar to octique -- thanks for the pointer. I've also renamed it to `adequate-iso-7000`.

Unfortunately, the automatic check is complaining about the file size, but its necessary to bundle all the ISO icons.

saecki:
> I've bundled it similar to octique -- thanks for the pointer. I've also renamed it to `adequate-iso-7000`.

Perfect, thanks!

> Unfortunately, the automatic check is complaining about the file size, but its necessary to bundle all the ISO icons.

That's ok in this case, it's just a warning.

PR #4084, opened 2026-02-11
se-jyutcitzi:0.2.3

VincentTam:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed logic error about the syllabic nasal sounds.

In the old version, my package won't treat "hm" (the Cantonese jyut6ping3 romanization of the "hmm…" sound in English) correctly.

Apart from that, the vertical alignment of the jyutcitzi of those sounds will be improved.

### DIFF
--- a/packages/preview/se-jyutcitzi/0.2.3/display.typ
+++ b/packages/preview/se-jyutcitzi/0.2.3/display.typ
@@ -1,0 +1,53 @@
+#import "utils.typ": split-jyutping
+#import "data.typ": *
+
+/// Combine 2 parts: combine mode
+/// - "-" for stacking part 1 on top of part 2;
+/// - "|" for combining them side-by-side
+#let combine-parts(part1, part2, combine-mode, tone: none) = {
+  set text(bottom-edge: "descender", top-edge: "ascender")
+  let combined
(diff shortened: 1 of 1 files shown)


packages/preview/se-jyutcitzi/0.2.3/sample.png
saecki, line 1: Could you add this to the `exclude` list? See [what to exclude](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude)

VincentTam: It's linked in README, so from section 2, I suppose I've to include it.

saecki: It belongs to the `Documentation Files` category, which means it should be committed to this repository, but added to the `exclude` list. That way it will be available when viewing the readme on Typst Universe, but users won't download when using the package.

VincentTam: I see thanks for explanation.  I confused "committing to the repo" with "putting into `exclude`".
Looking back I shouldn't put the `test/` files.  I'm still new to `typship`.  I'm gonna close this in favor of newer version.

PR #4085, opened 2026-02-11
simple-plot:0.2.6

nathan-ed:
I am submitting
- [x] an update for a package

#### Added
- **Grid label breaks**: White boxes behind tick labels create elegant breaks in grid lines (enabled by default)
- **Integer ticks by default**: Tick step defaults to 1 for cleaner integer labels
- **Tick label step**: `xtick-label-step` and `ytick-label-step` to show labels only at every N-th tick
- **Unit label only mode**: `unit-label-only: true` shows only "1" on axes for minimal style
- **Axis arrows extend beyond grid**: Axes now extend 0.5 units beyond the grid by default on the arrow side
- Axis labels (x, y) now position at the extended arrow tips

#### Changed
- Default tick label size changed from `0.65em` to `10pt` (same as axis labels)
- Default axis label size changed from `0.8em` to `10pt` for consistent sizing
- Grid lines no longer extend beyond the plot bounds (only axes extend)
- Minor grid step default changed to 5 subdivisions

#### Fixed
- White box positioning for grid label breaks now properly accounts for text dimensions and minus signs

#### Fixed
- Function `label-pos` now respects explicit function domains; when no domain is set, it falls back to the axis range

### DIFF
--- a/packages/preview/simple-plot/0.2.6/README.md
+++ b/packages/preview/simple-plot/0.2.6/README.md
@@ -4,6 +4,10 @@ A simple, pgfplots-like function plotting library for Typst. Create beautiful ma
 
 > **Note:** This package is built on top of [CeTZ](https://github.com/cetz-package/cetz) v0.4.2.
 
+## Manual
+
+A full manual is available in [docs/manual.pdf](docs/manual.pdf), with a Typst source version in [docs/manual.typ](docs/manual.typ)
+
 ## Gallery
 
 Click on an image to see the source code.


packages/preview/simple-plot/0.2.6/docs/manual.pdf
elegaanz, line 1: Because you don't seem to link to that file in your README, it will be very hard to discover for people viewing the package on Typst Universe. Can you please a link?

nathan-ed: Done, thanks for the kind reminder!

elegaanz:
Thank you.

PR #4086, opened 2026-02-11
exercise-bank:0.4.0

nathan-ed:
I am submitting
- [x] an update for a package

#### Changed (Breaking)
- **New display control system**: Replaced `solution-mode`, `fallback-to-correction`, and `append-solution-to-correction` with three clearer parameters:
  - `display`: Controls what to display - `"ex"` (exercises only), `"sol"` (solutions only), `"both"` (default)
  - `corrDisplay`: Controls which content type to show - `"solution"` (default), `"correction"`, `"mixed"`
  - `corrLoc`: Controls where solutions appear - `"after"` (default), `"pagebreak"`, `"end-section"`, `"end-chapter"`

- **New exercise-level flags**:
  - `solInCorr`: If true, correction already contains solution (show only correction, not both)
  - `showCorr`: If true, show correction in "mixed" mode for this exercise

#### Removed
- `solution-mode` parameter (replaced by `show` and `corrLoc`)
- `fallback-to-correction` parameter (behavior controlled by `corrDisplay`)
- `append-solution-to-correction` parameter (use `corrDisplay: "mixed"` instead)
- `solution-in-correction-style` parameter (no longer needed)


saecki:
Thanks for the update!

PR #4104, opened 2026-02-13
aero-dhbw:0.1.1

cocoveal:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
Minor change to how logos are rendered on the titlepage, to make vertical spacing between logos and the title of the paper more consistent. 

Note:
Since I have to upload a whole new folder, you can't really see the change based off of the commits alone.
In `src/titlepage.typ` I changed the following from:
```typ
#let cover(source) = {
  set image(width: 6cm)
  source
}
```
to:
```typ
#let cover(source) = {
  set image(height: 2cm, fit: "contain")
  source
}
```


saecki:
Thanks!

PR #4105, opened 2026-02-13
uastw_thesis:0.1.0

mhorauer:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a thesis templates for students at the University of Applied Sciences Technikum Wien.

- [x ] selected [a name](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) that isn't the most obvious or canonical name for what the package does
- [ x] added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys
- [ x] added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package
- [x ] have chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) and added a `LICENSE` file or linked one in my `README.md`
- [x ] tested my package locally on my system and it worked


mhorauer:
the 2 WARNINGS stem from the upstream CODLY package

packages/preview/uas-technikum-wien/0.1.0/typst.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "uas-technikum-wien"
```
elegaanz: Because this is not an official package of the University, can you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university.

If you want to make this template official to keep the current name, that's also an option. In that case a member of the University staff should send an email to `hello@typst.app` stating that they officially endorse your package. In practice it means that they will accept submission of works built using it, and ideally promote it (for example by making it available along Word and LaTeX templates via the campus information system) and guarantee that it will be maintained on the long term.

packages/preview/uas-technikum-wien/0.1.0/typst.toml
```diff
@@ -0,0 +1,20 @@
...
+  "Martin Horauer <https://github.com/mhorauer/>"
+]
+license = "GPL-3.0-or-later"
+description = "Typst template for bachelor and master thesis at UAS Technikum Wien"
+repository = "https://github.com/mhorauer/typst-templates"
+keywords = ["Technikum Wien", "thesis", "bachelor", "master"]
+exclude = ["*.pdf", "src/fonts/", ".github"]
+categories = ["office"]
+disciplines = ["computer-science", "education"]
+compiler = "0.14.2"
```
elegaanz: Unless you depend on a fix that was released in 0.14.2, this number should be lower: it corresponds to the oldest supported compiler version. For example, lowering it to 0.14.0 should allow people still on 0.14.0 or 0.14.1 to use this template, which I think should be possible. If you don't depend on features introduced in 0.14 you can even make it lower than that.

packages/preview/uas-technikum-wien/0.1.0/template/sections/images/logo.svg
elegaanz, line 1: Under which terms can this logo be distributed and used? Can you double check that you are allowed to ship it in this package, and if not replace it with a placeholder and instructions on how to obtain it? Can you please also specify under which conditions — if the university specifies any — it may be used and explain it in your README?

packages/preview/uas-technikum-wien/0.1.0/template/Makefile
elegaanz, line 1: Ideally, we prefer build files to not be shipped with templates, especially for people using the web app, where they are useless and may cause confusion. Can you please delete that file?

PR #4107, opened 2026-02-13
athena-tu-darmstadt-thesis:0.1.2

JeyRunner:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This is an unofficial thesis template for the University TU Darmstadt (Technische Universität Darmstadt) following the university design guidelines while imitating the existing official latex template. This update improves various small aspects.

### DIFF
--- a/packages/preview/athena-tu-darmstadt-thesis/0.1.2/example/main.typ
+++ b/packages/preview/athena-tu-darmstadt-thesis/0.1.2/example/main.typ
@@ -17,7 +17,7 @@
   ],
 
   // TODO: download actual log and convert according to README,
-  // then replace with image("logos/tuda_logo.pdf")
+  // then replace with image("logos/tuda_logo.svg")
   logo_tuda: image("logos/tuda_logo_replace.svg"),
   accentcolor: "9c",
  


packages/preview/athena-tu-darmstadt-thesis/0.1.2/README.md
```diff
@@ -0,0 +1,134 @@
...
+```bash
+typst compile main.typ --font-path assets/fonts/ --pdf-standard a-2b
+```
+In case this should not yield a PDF which is accepted by TUbama, you can use a converter to convert from the Typst output to PDF/A, but check that there are no losses during the conversion. 
+
+## Development
+
+Create a symlink to more easily develop your package changes:
+```
+DEV_TEMPLATE=not-tudabeamer-2023
```
elegaanz: The template name and version seem wrong here. More generally, I don't think this section is very useful to people looking at the package on Typst Universe, I would advise to simply remove it. Of course, if you think it is beneficial to keep it you also can, I don't have a strong opinion on it.

packages/preview/athena-tu-darmstadt-thesis/0.1.2/example/logos/convert_logo.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh 
+
+# first download the logo from:
+#   https://download.hrz.tu-darmstadt.de/protected/ULB/tuda_logo.pdf
+
+cd "$(dirname "$0")"
+
+# install with: apt get install pdf2svg
+pdf2svg tuda_logo.pdf tuda_logo.svg
```
elegaanz, line 9: Now that Typst supports PDF images, this script may no longer be necessary. You don't have to change that in this version, but I just wanted to let you now in case you missed the new feature, so that you can eventually benefit from it.

JeyRunner: I'll keep the convert script for now, in case someone uses a bit older version of typst. In a future version the conversion script will be removed.

elegaanz:
Thank you.

PR #4127, opened 2026-02-16
eggs:0.5.0

retroflexivity:
I am submitting
- [ ] a new package
- [x] an update for a package

- Added guards against using the functions without the config set
- Richer gloss line error message with example number for easier tracing
- Fixed various bugs with subexample numbering


elegaanz:
Thanks!

PR #4130, opened 2026-02-16
phonokit:0.4.1

guilhermegarcia:
## Summary
- Minor update to phonokit (v0.4.1) where one function (`#consonants()`) has better alignment in column labels and a new argument is introduced (optional): `abbreviate` (`bool`).


saecki:
Thanks!

PR #4135, opened 2026-02-17
pariman:0.2.1

pacaunt:
I am submitting
- [ ] a new package
- [x] an update for a package

Added `explicit-method` property to toggle unit display when accessing with `.method` property.


elegaanz:
Thank you!

PR #4136, opened 2026-02-18
vocabulo:0.2.0

towoe:
The update makes the package more flexible.
Sections can now be freely arranged.
Links are added to both table types.
The dimensions are now pulled in from a separate package.

I am submitting
- [ ] a new package
- [x] an update for a package

Description: Template for vocabulary writing practice.


saecki:
Thanks for the update!

PR #4141, opened 2026-02-18
simple-unimi-thesis:0.1.1

VictuarVi:
I am submitting
- [ ] a new package
- [x] an update for a package

Updated `link()` to account for Typst 0.14 changes for empty links.


packages/preview/simple-unimi-thesis/0.1.1/src/lib.typ
```diff
@@ -0,0 +1,465 @@
...
+
+#let document-state = state("init", "TITLE_PAGE")
+
+#let localization = yaml("utils/locale.yaml")
+
+#let project(
+  // Name of the university
+  university: "Università degli Studi di Milano",
+  // Path of the logo of the university
+  unilogo: "img/unimi.svg",
```
elegaanz: For info this doesn't allow users to import their own logo, only to use the ones shipped with the package (and they have to know the exact name of the file). See [this section of the docs for more info](https://github.com/typst/packages/blob/main/docs/resources.md#paths-are-package-scoped). It doesn't have to be fixed in this version if you just want to ship this patch now, but I wanted to let you know for a future version.

packages/preview/simple-unimi-thesis/0.1.1/src/img/unimi.svg
elegaanz, line 1: Regarding all these logos, it would be nice to specify under which terms they are distributed and can be used. This is usually done at the end of the package README, but can also go in the LICENSE file or some other dedicated file. If the University doesn't allow for their distribution by third parties, they should be replaced with placeholders and instructions to download them.

packages/preview/simple-unimi-thesis/0.1.1/src/img/loghi/unimi.pdf
elegaanz, line 1: This file seems to be duplicated in the parent folder, and also exists in SVG. Is there a reason for that? If not, would it be possible to remove two of them only keep one version please?

elegaanz (commented):
Small things regarding the various logos that I didn't notice in the previous review. It would be nice if you could change that before publishing this new version :)

packages/preview/simple-unimi-thesis/0.1.1/src/img/unimi.svg
VictuarVi: Could it be there isn't a specific license, at all? Should I just write "Unlicensed", in that case?

packages/preview/simple-unimi-thesis/0.1.1/src/img/loghi/unimi.pdf
VictuarVi: It should be resolved in c5f2eb1030e91c9a0e57aaf11c18e40a2d455968.

packages/preview/simple-unimi-thesis/0.1.1/src/lib.typ
VictuarVi: It should be resolved in c5f2eb1030e91c9a0e57aaf11c18e40a2d455968.

packages/preview/simple-unimi-thesis/0.1.1/src/img/unimi.svg
elegaanz: Yes there can be no specific license, but then the conditions under which you may distribute and use the logos depend on Italian copyright law. So you should double check how that works. In that case, you should specify that the logos are under copyright and the property of the university.

elegaanz: Hello, did you find any further information? If not, just specifying that the logo is copyrighted and the property of the university is enough. But in that case, it would also mean that distributing it with your package is not possible and that users should acquire it by their own means (you can still document where it can be found, for example if there is a page on your intranet for that).

VictuarVi: Hi, I could not find any information about the license of the logos. I will update the package accordingly.

VictuarVi:
In c11414ed675af0a4617a5de8ffb2c3f739cb2a2f I removed all the logos, added a placeholder if no logo is provided and updated the README with a link to download them.

Currently, in the github repo, there are all the files.

elegaanz:
Thank you!

PR #4143, opened 2026-02-18
simple-unibo-thesis:0.1.0

Soliprem:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package is a simple (unofficial) template for writing a thesis at unibo (University of Bologna)

- [x] [`excluded`](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude) PDFs or README images, if any, but not the LICENSE


Soliprem:
changed to draft because I found out about *another* template the university has, which includes the logo. Will mark as ready when I have it ported correctly and when I find a nice high-resolution image of the logo

Soliprem:
ok I can't think of anything else that's missing atm. I'm not sure how your policy for merging works, but if you prefer I squash the commits before merging, rather than do a squash merge, I can do that!

packages/preview/simple-unibo-thesis/0.1.0/README.md
```diff
@@ -0,0 +1,110 @@
...
+    dissertation: "Dissertation",
+    graduation-session: "Abschlussprüfung",
+    academic-year: "Akademisches Jahr",
+    abstract-title: "Zusammenfassung",
+  ),
+  ...
+)
+```
+
+## Licensing
```
elegaanz, line 112: Can you add a note about the university logo license here, please? Unless it is also distributed under MIT-0 by the university.

Soliprem: fairly certain it isn't under MIT for the university, my bad. At the same time, I can't find any information about its licensing. Since it was pretty easy to find, I assumed that Unibo was quite lax about it. I'll try to get more precise information. I know for a fact that students are allowed to use it in their dissertations, which is why I added it to the template. If it turns out to be an issue, I can just ship a placeholder image of a rickroll or something so the template file will compile, and leave an instruction comment on how to download the actual logo from the official website

Soliprem: I'm looking at other similar packages, but I might be missing something. How did `simple-unimi-thesis` and `simple-unive-thesis` handle this? I don't see any particular mentions in their files, but again I might have missed something

Soliprem: @elegaanz I think this solution works best. From the uni's website, they seem more restrictive than I thought, so I thought I'd air on the safer side and provide instructions on how to get the logo from their official website, and only bundle a very obvious placeholder (literally just a dark color, so an unmissable eyesore). The thumbnail technically shows the logo, but that should be explicitly fine under Art.15 of the university's rules on the usage of their logo (since it's a descriptive function used to represent the characteristics of a product or service, i.e. in this case of the package. I'm no lawyer, but it feels pretty linear)

elegaanz: > I'm looking at other similar packages, but I might be missing something. How did `simple-unimi-thesis` and `simple-unive-thesis` handle this? I don't see any particular mentions in their files, but again I might have missed something

Yeah, sometimes we miss that this mention is missing during the package review and don't ask package author to add it. It's not too bad, it's more to be extra-safe (both for the Typst team distributing the package and the package authors themselves) that we do it.



> @elegaanz I think this solution works best. From the uni's website, they seem more restrictive than I thought, so I thought I'd air on the safer side and provide instructions on how to get the logo from their official website, and only bundle a very obvious placeholder (literally just a dark color, so an unmissable eyesore). The thumbnail technically shows the logo, but that should be explicitly fine under Art.15 of the university's rules on the usage of their logo (since it's a descriptive function used to represent the characteristics of a product or service, i.e. in this case of the package. I'm no lawyer, but it feels pretty linear)

Sounds good, thanks!

PR #4144, opened 2026-02-18
simple-plot:0.3.0

nathan-ed:
I am submitting
- [x] an update for a package

  ### [0.3.0] - 2026-02-18

  #### Changed
  - **Grid label break**: Replaced white-box overlay approach with gap-based grid drawing. Grid lines are now drawn
  with gaps where labels are placed, instead of overlaying white rectangles. This makes `grid-label-break` work on
  any background color.
  - Internal refactor in tick label placement to consistently reuse the computed `label-offset` value.

  #### Fixed
  - Origin label handling now avoids duplicate `0` tick labels when `show-origin: true` and axes cross at the
  origin.


saecki:
Thanks for the update!

PR #4166, opened 2026-02-21
digidraw:0.9.0

joelvonrotz:
I am submitting
- [x] a new package
- [ ] an update for a package

`digidraw` is a drawing library to create digital timing diagrams and "inspired" by WaveDrom. The goal is to get as close to WaveDrom's syntax as possible. Compared to the Typst package [`wavy`](https://typst.app/universe/package/wavy/), `digidraw` uses `cetz` instead of WASM and thus supports various configuration options. Styling, fonts, colors and sizes can be changed. In other words, I wrote my "own" engine which uses WaveDrom's syntax.

I think this should be enough for `digidraw` to stand on its own. What are your thoughts on this?

### DIFF
--- a/packages/preview/digidraw/0.9.0/examples/example1.typ
+++ b/packages/preview/digidraw/0.9.0/examples/example1.typ
@@ -1,4 +1,4 @@
-#import "../src/wave.typ"
+#import "@preview/digidraw:0.9.0"
 
 #set page(width: 14cm, height: auto, margin: 2mm)
 #set align(center)
--- a/packages/preview/digidraw/0.9.0/typst.toml
+++ b/packages/preview/digidraw/0.9.0/typst.toml
@@ -10,6 +10,5 @@ keywords = ["digital timing","drawing","configurable","wavedrom","timing"]
 categories = ["visualization"]
 exclude = [
     "examples/",
-    "docs/banner.typ",
-    "docs/manual.typ",
+    "docs/",
 ]
--- a/packages/preview/digidraw/0.9.0/examples/example2.typ
+++ b/packages/preview/digidraw/0.9.0/examples/example2.typ
@@ -1,4 +1,4 @@
-#import "../src/wave.typ"
+#import "@preview/digidraw:0.9.0"
 
 #set page(width: 14cm, height: auto, margin: 2mm)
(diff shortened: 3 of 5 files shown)


joelvonrotz:
Also any pre-release feedback is welcome :)

packages/preview/digidraw/0.9.0/examples/example1.typ
```diff
@@ -0,0 +1,6 @@
+#import "../src/wave.typ"
```
saecki: The examples should use `@preview` imports, so users can directly copy them.

packages/preview/digidraw/0.9.0/typst.toml
```diff
@@ -0,0 +1,15 @@
...
+name = "digidraw"
+version = "0.9.0"
+entrypoint = "src/exports.typ"
+authors = ["Joel von Rotz"]
+license = "EUPL-1.2"
+description = "Drawing digital timing diagrams with style and (almost all of) the WaveDrom syntax."
+repository = "https://codeberg.org/joelvonrotz/typst-digidraw"
+keywords = ["digital timing","drawing","configurable","wavedrom","timing"]
+categories = ["visualization"]
+exclude = [
```
saecki, line 11: Could you also exclude the `manual.pdf` here? Probably the whole `docs` directory should be added. It will still be available on Typst Universe, just not when the user downloads the package.

joelvonrotz:
Thank you for your feedback, I've applied the changes. Should I also update the import statement of `.typ` files in the `docs` folder?

joelvonrotz:
Or since the examples are a "direct" file for the user, that only there the `@preview` import statement are required?

saecki:
> Thank you for your feedback, I've applied the changes. Should I also update the import statement of .typ files in the docs folder?
> Or since the examples are a "direct" file for the user, that only there the @preview import statement are required?

Yes, examples that are supposed to be read directly by the user should use `@preview` imports. For the banner or manual it doesn't really matter, since the user is not expected to build those themselves.

joelvonrotz:
Alright, will keep that in mind for the future. Thank you very much for taking the time to review my package! You guys do a good job!

PR #4167, opened 2026-02-21
cycle-ensaj:0.1.0

Achraf-saadali:
I am submitting
- [x] a new package
- [ ] an update for a package

### DIFF
--- a/packages/preview/cycle-ensaj/0.1.0/template/template.typ
+++ b/packages/preview/cycle-ensaj/0.1.0/template/template.typ
@@ -8,7 +8,7 @@
 #cover-page(
   title: "Design and Implementation of a Scalable DevOps CI/CD Pipeline for Modern Microservices Architectures",
 
-  subject-image: image("../template/devops_5266248.png"),
+  subject-image: image("devops_5266248.png"),
 
   module: "Architecture microservice et DevOps",
 
@@ -72,5 +72,6 @@
   "SQL",
 
 )
+
 
 


packages/preview/ensaj-report-template/0.1.0/src/cover.typ
```diff
@@ -0,0 +1,182 @@
...
+    )
+
+    #align(center + horizon)[
+      #text(size: 16pt)[
+        *Module*: #emph(module)
+      ]
+    ]
+
+    #if subject-image != none and subject-image != "" {
+      image("../template/"+subject-image, width: 7cm)
```
elegaanz: This won't work as expected: this path doesn't refer to the copy of the template being made in the user project, but to the original unmodified version of it that you ship with the package. The user won't be able to use their own image with this API. Instead, you should [take `image(...)` as a parameter directly](https://github.com/typst/packages/blob/main/docs/resources.md#paths-are-package-scoped).

packages/preview/ensaj-report-template/0.1.0/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## ⚠️ Disclaimer
+
+This project is **not officially affiliated with, endorsed by, or associated with ENSA El Jadida** or any institutional administrative body.
+
+It is an independent initiative developed voluntarily by a student as a personal academic contribution.
+
+The objective is to support fellow students by providing a structured and professional Typst template adapted to our school’s report requirements.
+
+This work reflects individual effort and initiative in contributing positively to the student academic community.
```
elegaanz: I think a mention of the terms under which the university logos are distributed and can be used by students (or other people using this template) would be in order here. It would also be nice to specify the same information (and/or license) of the "Dev ops" image, unless you made it yourself and want to place it under MIT too.

packages/preview/ensaj-report-template/0.1.0/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+---
+
+## 📌 Usage
+
+1. Import the template file:
+
+#import "lib.typ": cover-page, achraf_code_block
```
elegaanz: I would recommend quoting that code block using backticks, so that it renders better.

packages/preview/ensaj-report-template/0.1.0/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "ensaj-report-template"
```
elegaanz: If this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

Also the "template" part is a bit redundant, it doesn't really bring more to the name. The new name should preferably drop it.

Achraf-saadali:
Sure I will get to correct these  right away

Achraf-saadali:
Hey , I have double checked on the issues you addressed and i think i fullfilled and corrected all of them

elegaanz:
Looks good, thank you!

PR #4174, opened 2026-02-23
elegant-polimi-thesis:0.2.0

VictuarVi:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: fixed figure numbering, added executive summary, article format and other theses types; added examples, thumbnails. Removed localization, argument. Reorganized docs.


saecki:
Thanks for the update!

PR #4317, opened 2026-03-09
phonokit:0.4.6

guilhermegarcia:
## phonokit:0.4.6

A toolkit for creating phonological representations in Typst.

### What's new in 0.4.6

- Adds `intonational.typ` with the `int()` function for ToBI-style intonational annotation (inline above-text labels with stems)
- Numbered examples (`ex()`) now support titled examples and have been updated to accommodate ToBI examples


elegaanz:
Thank you!

PR #4371, opened 2026-03-16
phonokit:0.5.0

guilhermegarcia:
phonokit:0.5.0

Add feature geometry module (`geom.typ`) with support for arrows, delinking, and node highlights. Includes pre-defined phoneme segment presets. Also includes the intonational module (`intonational.typ`) for inline ToBI annotations introduced in 0.4.6.


elegaanz:
Thanks!

PR #4428, opened 2026-03-23
phonokit:0.5.1

guilhermegarcia:
phonokit:0.5.1

Minor improvements to existing functions.

- Constraint-based functions now accept prosodic structure as candidates (i.e., the user can employ the prosodic functions in the package already together inside tableaux); see manual
- `gloss` is introduced as a new parameter to tableaux


saecki:
Thanks!

PR #4471, opened 2026-03-30
phonokit:0.5.3

guilhermegarcia:
## Summary

- OT tableaux: shading is improved; more complex inputs can now be added
- `#consonants()` function now allows for a simplified version of inventories
- Prosody: `distance` parameter now lets the user adjust level distances (the default have also been improved, so chances are the user will never _have_ to use this parameter; it's just there for the sake of customization)
- Autosegmental: added `dash` parameter for association line styles (`"solid"`, `"dotted"`, `"dashed"`); defaults to `dashed`
- `#ipa()` now accepts many more inputs (many-to-one relationships with short and long codes for a wide range of symbols), so coverage is virtually complete
- Minor fixes throughout


elegaanz:
Thank you.

PR #4511, opened 2026-04-04
scoryst:0.1.0

bernsteining:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:

A Typst plugin that allows music engraving in Typst from multiple formats using [Verovio](https://www.verovio.org/), thanks to WASM.

## Features

- **6 input formats**: ABC, MusicXML, MEI, Humdrum, Volpiano, CMME
- **5 [Standard Music Font Layout](https://github.com/w3c/smufl)**: Leipzig (default), Bravura, Gootville, Leland, Petaluma
- **Full Verovio options**: scale, font, page layout, and all [toolkit options](https://book.verovio.org/toolkit-reference/toolkit-options.html)

Check the [documentation](https://github.com/bernsteining/verovio/blob/v0.1.0/test/documentation.pdf) for a full demonstration with examples.


packages/preview/verovio/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "verovio"
```
elegaanz: Since this is not an official port of Verovio for Typst, can you pick another name please? It can still make a reference to the original library, but it should be clear that it is not affiliated with it.

packages/preview/scoryst/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "verovio"
+version = "0.1.0"
+entrypoint = "verovio.typ"
+authors = ["bernsteining"]
+license = "LGPL-3.0-only"
+description = "🎼 Music engraving for Typst via Verovio — render ABC, MusicXML, MEI, Humdrum, Volpiano, and CMME notation"
+repository = "https://github.com/bernsteining/verovio"
+keywords = ["music", "notation", "score", "scoring", "partition", "tablature",  "engraving", "verovio", "abc", "musicxml", "mei", "humdrum"]
+categories = ["visualization"]
+compiler = "0.14.2"
```
elegaanz: This corresponds to the earliest supported Typst version, so unless you depend on a patch introduced in 0.14.2 this can be at least 0.14.0. It may even be lower depending on what features you are actually using. By picking the latest version, people who didn't update (or can't update) won't be able to import your package, even if it may work fine in practice.

packages/preview/verovio/0.1.0/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Renders music notation to an SVG image. `data` is a string in any supported
+format. `..args` are forwarded to Typst's `image()` function (`width`,
+`height`, `fit`, `alt`).
+
+**`music-page-count(data, options: none)`**
+
+Returns the number of pages for the given music data.
+
+## Building
```
elegaanz: This section is not really relevant to people browsing Typst Universe. I would suggest to remove it from this copy of the README. This not a blocker, and if you think it might still be useful or if you don't want to maintain two versions of the README, it is also perfectly fine to keep it.

elegaanz (commented):
Thank you for this package :) A few remarks:

packages/preview/verovio/0.1.0/README.md
bernsteining: True, thanks

packages/preview/scoryst/0.1.0/typst.toml
bernsteining: I would suggest people to update to the latest version ? Especially in the changelog it's suggested that people update 0.14.{0,1} to 0.14.2 (cf https://typst.app/docs/changelog/0.14.2/), so let's stick with 0.14.2 nah ?

packages/preview/verovio/0.1.0/typst.toml
bernsteining: Okay gonna rename to scoryst (score, typst, choriste mashup)

packages/preview/scoryst/0.1.0/typst.toml
elegaanz: Yes, of course it is always better to be on the very last patch version, but people may depend on their distribution package manager and may not be able to use the very last version, especially if it was released recently. They may also need some older packages that are no longer working with more recent Typst versions. So having a bit more flexibility by not requiring the very last version is better.

MDLC01:
I've been trying to compile MuseScore to Wasm to embed it into Typst without success for a while. I knew about Verovio but didn't look into it that much. Very impressive that you managed to do it! I will definitely look into it when I have more time!

bernsteining:
> I've been trying to compile MuseScore to Wasm to embed it into Typst without success for a while. I knew about Verovio but didn't look into it that much. Very impressive that you managed to do it! I will definitely look into it when I have more time!

Thanks! 

I tried to compile Verovio for Typst several times in August without success but then managed to achieve it recently, lots of shenanigans. If you find bugs or have speedup ideas tell me :)

packages/preview/scoryst/0.1.0/typst.toml
bernsteining: Then should I find the lowest version for which my plugin works ?

elegaanz: Ideally, yes. You can use [the changelogs](https://typst.app/docs/changelog/) to see what breaking changes each version brings and when specific features were introduced. But since this is quite a lot of work to properly do, changing the limit to `0.14.0` is also a good compromise: most people should be using the 0.14 series, and it's unlikely that one of the patches in 0.14.1 or 0.14.2 broke your package.

bernsteining: Okay done

elegaanz:
Thank you!

PR #4522, opened 2026-04-06
uepg-tcc-unofficial:0.1.0

ekspy:
## Summary
- New template package for UEPG (Universidade Estadual de Ponta Grossa) monographs following ABNT norms (NBR 14724)
- Includes cover page, title page, dedication, acknowledgments, epigraph, abstract (PT/EN), abbreviation list, table of contents, and ABNT-compliant formatting
- Conditional auto-generation of illustration/table/frame lists based on document content

## Checklist
- [x] `typst.toml` with all required fields
- [x] `lib.typ` entry point with `monografia()` and `citacao-longa()` exports
- [x] `template/main.typ` with complete example document
- [x] `template/refs.bib` with example bibliography
- [x] `thumbnail.png` generated from template output
- [x] `LICENSE` (MIT)
- [x] `README.md` with usage instructions and parameter documentation
- [x] Compiles successfully with `typst compile`


packages/preview/monografia-uepg/0.1.0/typst.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "monografia-uepg"
```
saecki: If this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

saecki (commented):
Thanks for the package!

packages/preview/monografia-uepg/0.1.0/typst.toml
ekspy: Renamed to `abnt-monografia`. Thanks for the feedback!

packages/preview/uepg-tcc-unofficial/0.1.0/README.md
```diff
@@ -0,0 +1,117 @@
+# abnt-monografia
+
```
saecki, line 4: Sorry I didn't catch this in my previous review. It is generally appreciated to have at least an introduction sentence in English, even if the target audience for the template is not native English speakers. This allows the broader Typst community to get an idea of what the package is about and make sure that it isn't of interest to them.

packages/preview/monografia-uepg/0.1.0/typst.toml
saecki: This name has the same problem as the name before, because `abnt` is a canonical name for the entity. An unofficial name should have a non-descriptive prefix or the whole name should be more creative/unique.
Feel free to ask if you have any questions/suggestions :)

ekspy: Renamed to `uepg-tcc-unofficial` following the same pattern as `utfpr-tcc-unofficial`. Also added an English introduction to the README. Thanks!

saecki:
Thanks!

PR #4540, opened 2026-04-08
zen-zine:0.3.1

tomeichlersmith:
I am submitting
- [ ] a new package
- [x] an update for a package

Just a patch to make error messages more helpful when the deduced number of pages might be different from what the user expects.


saecki:
Thanks!

PR #4542, opened 2026-04-09
uniwarn:0.1.0

zral0kh:
I am submitting
- [x] a new package
- [ ] an update for a package

`uniwarn` allows other universe packages to emit namespaced typst compile warnings, which users then can also disable/enable if they so choose.
It is not directly intended for end users but for package authors or rather their packages.

It exposes four functions:
1. `warning(namespace: str, prefix:str, str)` the function that will emit the warning. 
2. `register-namespace(str)` registers a namespace and panics if it has already been registered.
3. `disable-warnings(str)` disables warnings for a specific namespace. By default they are enabled.
4. `enable-warnings(str)` enables warnings for a specific namespace.

The latter two have an alias: `disable` and `enable` which some end user might want to use if a package does not expose their disable, enable method directly.

The namespace has to be chosen carefully, such that it does not collide with the namespace of another package.
The warning is emitted via `text(font: BAD_FONT)` and we use backspaces to remove the original warning.
This of course does not yield perfect warnings, as the shown code instance will not be correct, but it may very well be more helpful than not, the message just has to be even more expressive.

Bc of the way typst handles warnings they are also automatically deduped.

PS: yes i know that laurenz does not really want packages to get the ability to emit warnings, but his concern in some issue was that the user needs the ability to turn them off and this package allows that.

not a template...


zral0kh:
nvm i was dumb we don't need to use the text features, can just use normal state as updates don't require context.

packages/preview/font-warnings/0.1.0/typst.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "font-warnings"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["zral0kh", "OrangeX4"]
+license = "MIT"
+description = "A miniscule package for package authors to emit custom warnings."
+repository = "https://github.com/fornings-typ/font-warnings"
+keywords = ["warning", "illegal"]
+categories = ["utility"]
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the package!

packages/preview/font-warnings/0.1.0/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "font-warnings"
```
saecki: The name feels a bit misleading to me, I initially thought this would somehow deal with font-fallback warnings. Maybe I'm missing something? Otherwise could you change it to something else?
Also see: https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules

packages/preview/font-warnings/0.1.0/typst.toml
zral0kh: so i should it as low as possible then?
will check what still works and fix that.

packages/preview/font-warnings/0.1.0/typst.toml
zral0kh: i was also thinking of `fornings` or some actual synonym for warning like alert or caution, but i want to keep it close to warning really, since that is the syntax typst uses. the prefix `font-` just came from main mechanism being an abuse of the font warning typst emits naturally.

zral0kh: some other ideas: `pack-warnings`,  `uniwarnings`, `uniwarn` as that is the whole point: allowing packages (from universe) to emit compile warnings.

zral0kh: this becomes obsolete the moment we get warnings natively anyway

packages/preview/font-warnings/0.1.0/typst.toml
saecki: Ideally yes, but at this stage of Typst most people will be using the last two minor releases, so you don't need to go back *that* far :)

packages/preview/font-warnings/0.1.0/typst.toml
saecki: From the examples you listed I think I'd prefer `uniwarn`, since it feels the most like a coherent name, not just two different words.

packages/preview/font-warnings/0.1.0/typst.toml
zral0kh: yes okay.

packages/preview/font-warnings/0.1.0/typst.toml
zral0kh: yes true, given other package's names that sounds best,

zral0kh:
@saecki thanks for the feedback, i am actually more happy with the name now, having to think about it more made sense :)

saecki:
Thanks!

PR #4546, opened 2026-04-09
hetvid:0.2.1

itpyi:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: change default for authors (remove the placeholder so if this field is not set by user, this field will not be displayed); add a field called version, which is useful for note.


saecki:
Thanks!

PR #4547, opened 2026-04-09
ori:0.2.5 and theorion:0.6.0

OrangeX4:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- **BREAKING CHANGE: `frame-box` no longer sets `outlined: false`** — unnumbered theorems created via `theorem-box` etc. now appear in outlines by default. Use `#theorem-box(outlined: false)[...]` to opt out.
- **BREAKING CHANGE: natural unnumbered theorem prefix** — `frame-box` (i.e. `frame.with(numbering: none)`) now displays the supplement prefix (e.g. "Theorem", "Example") without numbering, instead of losing the prefix entirely. No need to manually type `#theorem-box[Theorem (Title)][...]` anymore — just use `#theorem-box[Title][...]`.
- **feat: `full-title` parameter** — pass `full-title` to any frame to override the auto-generated full title (prefix + number + title). Works with both numbered and unnumbered theorems, and is respected by `theorion-restate`.
- **feat: unnumbered theorem references** — `@label` on unnumbered theorems now shows "Supplement (Title)" instead of "Supplement " with a trailing space. `@label[-]` shows the title only, and `@label[!!]` shows supplement + title.
- **fix: use stick for indent-repairer to fix empty frame after page break**


elegaanz:
Thank you.

PR #4569, opened 2026-04-12
typographix-polytechnique-reports:0.2.1

remigerme:
I am submitting
- [ ] a new package
- [x] an update for a package

Note that version v0.2.0 was not submitted to typst-packages as it was only a refactor of the internal code. The codebase was split accross several files and is now contained entirely in `main.typ`. This was performed without affecting the behavior of the template. 

## Changelogs
- first line indentation: French mode (all lines are indented) and English mode (the first one is not)
  - smart default based on contextual `text.lang`
  - introduce parameter `first-line-indent-all` to account for user personal preferences
- headings are now inside blocks rather than boxes (caused issues for first line indentation)
- headings do not use dashes in their numbering anymore


elegaanz:
Do you want to rename the function to kebab-case as suggested? It's also fine to publish it as is if you prefer.

remigerme:
Just renamed everything that was still snake_case to kebab-case :)

elegaanz:
Thank you :)

PR #4571, opened 2026-04-12
phonokit:0.5.7

guilhermegarcia:
phonokit:0.5.7

- `#multi-tier()` now also supports automatic labelling, so linking nodes no longer needs explicit coordinates


elegaanz:
Thank you.

PR #4573, opened 2026-04-13
diatypst:0.9.2

skriptum:
I am submitting
- [ ] a new package
- [x] an update for a package

Fix diatypst#33 (references for figures) and diatypst#34 (math block in quotes)

### DIFF
--- a/packages/preview/xwysyy/0.1.0/LICENSE
+++ b/packages/preview/xwysyy/0.1.0/LICENSE
@@ -1,0 +1,22 @@
+MIT License
+
+Copyright (c) 2025 Carlos-Mero
+Copyright (c) 2026 xwysyy
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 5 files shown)


packages/preview/diatypst/0.9.2/diatypst.code-workspace
elegaanz, line 1: Can you delete that file please? It doesn't really make sense to ship it in this repository.

elegaanz (commented):
Thank you for the update. I just have a small remark. It would also be nice to fix the warnings reported in CI (at least the ones about alternative text, the invalid syntax in example makes sense in that context). But that's not mandatory for this new version to be published.

packages/preview/diatypst/0.9.2/diatypst.code-workspace
skriptum: oopsie, dont know how that slipped in

skriptum:
> Thank you for the update. I just have a small remark. It would also be nice to fix the warnings reported in CI (at least the ones about alternative text, the invalid syntax in example makes sense in that context). But that's not mandatory for this new version to be published.

updated the alt texts, should be more descriptive now. I also think that the invalid syntax in the example is better for brevity, so Ill leave it at that. For the linked Example PDFs, I would also like to keep the absolute link, so that it always opens to the up to date one. But maybe thats an incorrect approach?

elegaanz:
The idea of linking to a tag or specific commit instead of branch is that the link will always be valid for the version people are importing on Universe. The goal is to avoid the following scenario:

- you publish v0.9.2
- after that you start working on v0.10.0 which introduces some breaking changes. So you update the example files to use the new API, and push that on `main`.
- someone sees the package on Typst Universe (still at v0.9.2) and wants to try one of the example.
- it doesn't work because what they copied from your repository uses the new API and not the one exposed by v0.9.2.

If you linked to the `v0.9.2` tag, the example would still use the previous API and work fine.

PR #4575, opened 2026-04-13
chordish:0.2.2

soxfox42:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Chordish 0.2.2 contains a fix for `#seq` in `chords: "above"` mode.


elegaanz:
Thank you.

PR #4578, opened 2026-04-14
agregyst:1.2.0

MDLC01:
I am submitting
- [ ] a new package
- [x] an update for a package

Bibliography fixes and automatic summary detection for recap.


MDLC01:
I don't think the warning is relevant here. The link may become dead in the future if the file is moved but that would be surprising since it is already linked from a lot of places.

saecki:
Thanks!

PR #4585, opened 2026-04-14
ab-annotate:0.1.0

cwimpy:
Adds `ab-annotate:0.1.0` — an annotated bibliography package.

Each entry renders as **citation → abstract → user annotation**, reading the standard `abstract` and `annotation` (or `annote` as fallback) biblatex fields from a `.bib` file. Both fields are optional per entry; entries without either render as a normal citation.

## Usage

```typst
#import "@preview/ab-annotate:0.1.0": annotated-bib

#annotated-bib(
  read("references.bib", encoding: none),
  style: "apa",
)
```

The function takes bytes (from the user's own `read()` call) rather than a path, because Typst resolves `read()` relative to the containing file — a package cannot read a path the user gives it. The bytes are passed straight through to `bibliography()`, which accepts bytes since 0.12.

## Checklist

- [x] `README.md` documents usage with a compilable example using the `@preview` import
- [x] `LICENSE` (MIT) matches the `license` field in `typst.toml`
- [x] No binary blobs (no example PDFs included)
- [x] Name: `ab-annotate` — not the canonical name for this functionality (`annotated-bibliography` / `bibliography` would be), uses kebab-case, no "typst" in name
- [x] Compiles cleanly against Typst 0.14.2 (smoke-tested locally via `@local` namespace)

## Cross-format companion

Part of a cross-format toolchain (also provides a biblatex `.sty` and a Quarto extension reading the same `.bib` fields). Full source: https://github.com/cwimpy/ab-annotate

### DIFF
--- a/packages/preview/ab-annotate/0.1.0/typst.toml
+++ b/packages/preview/ab-annotate/0.1.0/typst.toml
@@ -1,0 +1,11 @@
+[package]
+name = "ab-annotate"
+version = "0.1.0"
+entrypoint = "ab-annotate.typ"
+authors = ["Cameron Wimpy <cwimpy@astate.edu>"]
+license = "MIT"
+description = "Annotated bibliographies that render abstracts and user annotations beneath each citation entry."
+repository = "https://github.com/cwimpy/ab-annotate"
+keywords = ["bibliography", "annotated-bibliography", "citation", "biblatex"]
+categories = ["scripting"]
+compiler = "0.11.0"
--- a/packages/preview/ab-annotate/0.1.0/typst.toml
+++ b/packages/preview/ab-annotate/0.1.0/typst.toml
@@ -8,4 +8,4 @@ description = "Annotated bibliographies that render abstracts and user annotatio
 repository = "https://github.com/cwimpy/ab-annotate"
 keywords = ["bibliography", "annotated-bibliography", "citation", "biblatex"]
 categories = ["scripting"]
-compiler = "0.11.0"
+compiler = "0.13.0"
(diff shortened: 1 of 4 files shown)


packages/preview/ab-annotate/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "ab-annotate"
+version = "0.1.0"
+entrypoint = "ab-annotate.typ"
+authors = ["Cameron Wimpy <cwimpy@astate.edu>"]
+license = "MIT"
+description = "Annotated bibliographies that render abstracts and user annotations beneath each citation entry."
+repository = "https://github.com/cwimpy/ab-annotate"
+keywords = ["bibliography", "annotated-bibliography", "citation", "biblatex"]
+categories = ["scripting"]
+compiler = "0.11.0"
```
elegaanz: From what I can tell, passing `bytes` to `bibliography` is supported [since 0.13.0](https://typst.app/docs/changelog/0.13.0/#:~:text=Functions%20that%20accept%20file%20paths%20now%20also%20accept%20raw%20bytes%2C%20for%20full%20flexibility). Your package won't work with 0.11 or 0.12, can you update that field please?

elegaanz (commented):
Just a small thing to fix :) Thank you for this package.

packages/preview/ab-annotate/0.1.0/typst.toml
cwimpy: Thanks for catching that! I bumped to 0.13.0 in 2f650cc. Let me know if anything else needs adjustment.

elegaanz:
Thanks!

PR #4594, opened 2026-04-16
blockcell:0.1.0

daleione:
Composable block-and-cell diagrams for visualizing structured data layouts, memory maps, register fields, protocol formats, and more.

Repository: https://github.com/daleione/blockcell


saecki:
Thanks!

PR #4596, opened 2026-04-16
htlium:1.1.1

DeltaAT:
I am submitting
- [x] a new package
- [ ] an update for a package

This is a template for protocols, that me and my classmates from the HTBLuVA Salzburg (and potentially everyone) can use.
It has useful implementations and also toggles.
Note that this template is german only but there may be a english version in the future.

- [ X added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package


packages/preview/htlium/1.1.1/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "htlium"
+version = "1.1.1"
+entrypoint = "src/main.typ"
+authors = ["Elias Pöschl <@DeltaAT>"]
+license = "MIT"
+description = "A template for school protocols, designed for HTL Salzburg, but adaptable to other schools as well."
+repository = "https://github.com/DeltaAT/htl-protocol"
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

packages/preview/htlium/1.1.1/examples/main.pdf
saecki, line 1: Could you delete this file?

packages/preview/htlium/1.1.1/src/main.typ
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    pagebreak()
+}
+
+#let template(
+  body,
+  author: "Your Name",
+  class-long: "Protokoll",
+  logo-url: "logo.png",
```
saecki: This won't work as expected. Any path passed in here will be resolved withing this packages directory, not the user directory. Instead templates commonly accept an image directly, which is also more flexible, since the user could pass in any content.
See: https://github.com/typst/packages/blob/main/docs/resources.md#paths-are-package-scoped

packages/preview/htlium/1.1.1/typst.toml
```diff
@@ -0,0 +1,17 @@
...
+exclude = ["examples"]
+categories = ["layout", "paper"]
+disciplines = ["computer-science", "education"]
+
+[template]
+path = "src"
```
saecki, line 15: The template should be a directory that uses the package, so essentially your `examples` directory would work here.

packages/preview/htlium/1.1.1/src/logo.png
saecki, line 1: Could you remove this duplicate logo? The logo is already present in the `examples` directory. The template already handles the `logo-url == none` case, so that shouldn't be an issue.

saecki (commented):
Thanks for this package! I've left a few review comments.

packages/preview/htlium/1.1.1/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "htlium"
+version = "1.1.1"
+entrypoint = "src/main.typ"
+authors = ["Elias Pöschl <@DeltaAT>"]
+license = "MIT"
+description = "A template for school protocols, designed for HTL Salzburg, but adaptable to other schools as well."
+repository = "https://github.com/DeltaAT/htl-protocol"
+compiler = "0.14.0"
+exclude = ["examples"]
```
saecki: If the `examples` directory is used for the template, it can't be excluded.

packages/preview/htlium/1.1.1/typst.toml
```diff
@@ -0,0 +1,17 @@
...
+categories = ["layout", "paper"]
+disciplines = ["computer-science", "education"]
+
+[template]
+path = "examples"
```
saecki: This isn't required by any means, but it's common to name this directory `template`. I only say that because the plural seems a bit odd, I think singular `example` would feel more natural as well.

saecki (commented):
Thanks! One final issue and a remark.

DeltaAT:
I can't see what is failing in these checks

saecki:
> I can't see what is failing in these checks

Oh, it's a bug in the package-check. The error failed to be converted into a github annotation.
The template fails to compile, because the directory is missing. It's because of the reason the one i mentioned here: https://github.com/typst/packages/pull/4596#discussion_r3131988085

saecki:
Thanks!

PR #4605, opened 2026-04-17
uniwarn:0.1.1

zral0kh:
I am submitting
- [ ] a new package
- [x] an update for a package

The update changes the state key used internally to be `'warnings::{namespace}`' instead of just `'{namespace}'`, which allows packages that compute on the state `state(package-name)` to use uniwarn.

And `register-namespace` now allows the named parameter `panic:true` which toggles whether it panics, when the namespace already exists. If it does not panic, it will emit a warning instead. 

See #4542 for a description of the package.


elegaanz:
Thanks.

PR #4606, opened 2026-04-17
modiagram:0.1.1

jonathandip:
I am submitting
- [ ] a new package
- [x] an update for a package

It is a package useful for drawing molecular orbitals and energy pathways. The update includes new functions like "ep-annotations" and "ep-difference".


jonathandip:
Is it OK even with those warnings?

saecki:
Thanks! This PR currently also includes the changes from version `0.1.0` of the pacakge. Could you rebase onto `main` so it only contains the newest package version?

saecki:
> Is it OK even with those warnings?

The warnings here are fine. Ideally the alternative descriptions should be a little more elaborate.

jonathandip:
Is it ok now?

saecki:
Yes, thanks!

PR #4610, opened 2026-04-17
unichar:0.4.0

MDLC01:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog

- Properly isolate text in `codepoint` show rule.
- Provide informative aliases from NamesList.txt under `codepoint(..).info.aliases`.
- Math classes now use their [Typst names](https://typst.app/docs/reference/math/class/#parameters-class).
- Use a Wasm plugin.


MDLC01:
The images are auto-generated from the code example. Would a generic "Result of the code example" alt text be sufficient or do you think it would be better to provide the text from the images directly?

Note that as of now, the rendered examples in the official Typst documentation don't have good alt text either ("Preview").

elegaanz:
Yes, the official docs are not great on that either, but it doesn't mean packages shouldn't try to do better. I would go for something a bit more detailed than just a generic sentence. For example

> Result of the code example: the requested information about each codepoint is displayed in the document.

could work for the first example. And

> Result of the code example: each codepoint is shown on a line along with its name and ID.

could be used for the second one.

These are just to give you an idea of what I would consider an appropriate image description here, feel free to rephrase it or add a bit more details. For the first one in particular, it might be insightful to pick one entry and describe it a bit more in depth. You can also argue that what it does is fairly obvious and not go further than a generic description, it's up to you to judge.

Something I'm using as a mental guideline when writing alternative text is to image that I'm reading the README to someone over the phone, and that I have to explain what the image shows. It helps to convey just enough information to be able to picture the image, while avoiding to go too much into details.

MDLC01:
I have added alternative text for both images

saecki:
Thanks!

PR #4615, opened 2026-04-18
canyon:0.1.0

cnaak:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A Book Template That Puts You in Full Control.


elegaanz:
Hello! I understand that the point of this package is to be as flexible as possible by letting people edit absolutely every part of the template, but this is not something we want to accept on Typst Universe. We consider that if your code is supposed to be modified that much, it doesn't really make sense to ship it using a package manager. This architecture also makes updating to a newer version very complicated, unless you have a proper version control system.

What I would recommend instead of publishing to Universe, is to have this template in a Git repository that can be cloned (or downloaded as an archive if the user doesn't use Git). It would be better suited to your use case I think.

cnaak:
I _extensively went through all the requirements_ and tested the package as to _adhere to everything previously stated_. This is evidenced by the fact that all PR tests passed and the merge is clear.

Moreover, I believe I'm bringing a **novel, full customization, approach**—really some fresh air, from the standpoint of end user, _which anyone that had suffered from template-side opaqueness and limited flexibility would readily understand and welcome_.

Further, this PR has been the fruit of _lengthy hard work_ and _voluntary contribution_ for the benefit of end-users.

And the package is **not** "supposed to be modified that much", it only **can**.

Unfortunately, your answer comes as a surprise, as stemming from personal taste—"not something we want to accept"—rather than previously stated and publicly extant requirements. Would you please reconsider your decision—in which case it will token Typst contributors that innovation is not hindered by unexpected factors. If so, thank you, but if not, _it would probably be a good idea to clearly state something to that effect with the other package requirements_.

Thank you and let me know it, ideally in a sensible time frame.

elegaanz:
Hello. [The "Template customization" section of the docs](https://github.com/typst/packages/blob/main/docs/typst.md#template-customization) deals with a very similar concern. On top of that, the way the package/template separation was designed, how it is documented in general, the examples we give, the officially maintained templates as well as the ones built by community all adhere to this architecture, suggesting it is the right way to structure a template. I'm sorry this was not explicit enough and I understand your frustration. If you are open to it, we can discuss how the current documentation can be improved to make it clearer that this architecture is mandatory.

Also, the CI check suite is not exhaustive (both for this repository and in general), there are always edge cases that are not covered by automated tests and a green check mark doesn't necessarily mean a pull request will be merged.

We are aware that the current architecture can feel limiting both for template authors and their users, and we have plans to solve this problem. In particular, custom elements should help a lot here. If you don't know about them and are curious, I would recommend reading [Laurenz's blog post about it](https://laurmaedje.github.io/posts/types-and-context/) (a bit old but still a quite accurate vision of what's to come) and [his presentation at the last community meetup](https://www.youtube.com/watch?v=PElGKQj0Ryc). If you are member of the Typst Discord, there is also [a dedicated discussion thread](https://discord.com/channels/1054443721975922748/1175895383600275516). However, it will still take some time before they are generally available.

Also, my previous suggestion of distributing this template as a Git repository that people can clone and start from still holds. All your work is not lost because this PR is not accepted, and I'm sure it will benefit other people. If you wanted to use Universe to promote the template, I can recommend making a forum and/or Discord post (in the "showcase" category).

cnaak:
Hello @elegaanz , thank you for your kind answer!

Pointing to the "Template Customization" part of the docs _fully answered my request to satisfaction_. So I guess I missed that part of the documentation and there is, indeed such a policy in place. Moreover, I was overly reliant on the CI check suite (both locally by `typst-package-check` and on GitHub).

There's no need for making a forum and/or Discord post about `canyon`, as I'm considering looking for ways to conform `canyon` (or another package) to the current specs.

Thank you for the other linked references, which I've found to be very interesting—I myself wish to be able to define custom `types` in Typst as mentioned in Laurenz's blog.

Best regards, and I'll close this PR with this comment!

PR #4616, opened 2026-04-18
invoice-pro:0.2.0

leonieziechmann:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Overhaul of the API and data model to improve flexibility and future proofing.


saecki:
Thanks!

PR #4631, opened 2026-04-19
mannot:0.3.3

ryuryu-ymj:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixes:
* Fix layout issues when marking math expressions starting with unary operators (ryuryu-ymj/mannot#8)


elegaanz:
Thanks!

PR #4633, opened 2026-04-19
tonguetoquill-usaf-memo:3.0.0

nibsbin:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 3.0.0 Update

- **DAF memo style.** New `memo_style: "daf"` option produces Department of the Air Force–formatted memos with fixed paragraph indentation and DAF-standard date formatting, distinct from the default USAF style.
- **`letterhead_seal_subtitle`.** Optional subtitle rendered beneath the letterhead seal image.
- **Letterhead formatting.** Title and caption are now uppercase; title is bold. Font defaults enforced to NimbusRomNo9L (Times New Roman equivalent) for both letterhead and body.
- **Unified paragraph formatter.** Replaced four near-identical formatter functions with a single `format-par` taking an `indent-fn` parameter, eliminating redundancy.
- **`auto_numbering` parameter removed.** Base-level paragraphs are now always auto-numbered per AFH 33-337 §2. Hierarchical numbering applies to explicit list/enum items only.
- **Indorsement DAF-style propagation fixed.** `memo_style` was previously read as a contextual content value in `indorsement.typ`, causing the DAF branch to never trigger.
- **Shared line-stride measurement.** Blank lines and body paragraphs now share a single cached line stride value for consistent vertical spacing.


nibsbin:
I've been using snake case since v1. Are kebab case names enforced now for published templates?

elegaanz:
It is not, but for some reason CI didn't catch it previously. However, I can see you are using a mix of kebab-case and snake_case, it would be nice to decide on either and stick to it. The Typst standard library and ecosystem generally uses kebab-case, which is why we have this lint, and I would recommend using it too, but if you prefer to have snake_case everywhere that's also possible.

packages/preview/tonguetoquill-usaf-memo/3.0.0/README.md
```diff
@@ -0,0 +1,348 @@
...
+- `Cinzel-Regular.ttf` — optional decorative font for footer taglines
+
+You can either clone the repository to pull all fonts or download just the files you need. All font files are available from the `fonts/` directory in the repo: https://github.com/nibsbin/tonguetoquill-usaf-memo/tree/main/fonts
+
+  - **Note:** *Times New Roman* is a proprietary Microsoft font that I can't distribute legally. After you add the *NimbusRomNo9L* files (see above), the templates use them to approximate Times.
+
+3. Start with one of the template files:
+   - `template/usaf-template.typ` for a standard Air Force memo
+   - `template/ussf-template.typ` for Space Force
+   - `template/starkindustries.typ` for a custom organization example
```
elegaanz, line 47: For info, when creating the template, the root will contain the same files as `template/`, meaning that the `template/` prefix might be confusing, as it won't exist.

nibsbin: Gotcha. This might be a deeper design issue, but I want the README to make sense for the https://github.com/nibsbin/tonguetoquill-usaf-memo repo as well (that's what it was primarily written for).

Will make the change if this is a hard requirement for the update

elegaanz: Yes, I agree that it is not great if you want to keep a unified version of the README, and that maintaining two slightly different ones is a bit of effort. If you think it won't affect users that much and that your README will still be understandable for someone discovering the package via the Typst Universe website, you can keep it as is.

nibsbin: Thanks for the feedback! I'll be back with another iteration for v3.0.0

PR #4645, opened 2026-04-21
quan:0.1.0

SchrodingerBlume:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: `quan` (圈, meaning "circled" in Chinese) is a lightweight package for typesetting circled numbers in Typst. It uses native Unicode circled glyphs (⓪–㊿) when the current font supports them, and automatically draws a circle around the number otherwise — so it works with any font, even those with partial or no circled-digit support. Users declare their font's coverage range via `quan-init`, and can fine-tune the drawn-circle appearance (stroke, radius, inset, size, letter-spacing) globally or per numeric range via `quan-style`. Built-in style rules are tuned for SimSun and cover 0–99 out of the box.

### DIFF
--- a/packages/preview/quan/0.1.0/LICENSE
+++ b/packages/preview/quan/0.1.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 SchrodingerBlume (Andrew Junhao Wu)
+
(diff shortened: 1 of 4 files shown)


elegaanz:
Thank you.

PR #4647, opened 2026-04-21
ioppub:0.1.1

munechika-koyo:
I am submitting
- [ ] a new package
- [x] an update for a package


packages/preview/ioppub/0.1.1/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "ioppub"
```
elegaanz: Because this package is not officially affiliated with IOP Publishing, can you please pick an alternative name, that makes it clear that this is community-built. It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

elegaanz (commented):
Thank you for this template. I just have a remark regarding the name, but otherwise it looks good to me.

packages/preview/ioppub/0.1.1/typst.toml
munechika-koyo: Thank you for your comment!
I updated `unofficial-ioppub` for both v0.1.0 and v0.1.1.
Should I modify only v0.1.1? Is there no need to fix anything except for the package.name in `typst.toml`?

elegaanz: Sorry, I didn't notice this was an update of an already published package. Keeping the previous name is fine then, it was my mistake to accept it during the first review. Sorry for making you change unnecessarily, feel free to revert your latest commit.

If you still want to change it, please only do it for the new version, not 0.1.0. In that case, the folder name should also be changed.

elegaanz:
Thank you.

packages/preview/ioppub/0.1.1/typst.toml
munechika-koyo: I apologize for not recognizing the guidelines as well.

PR #4678, opened 2026-04-24
simple-soc-report:0.1.0

tzyian:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: An unofficial template for Final Year Project reports for School of Computing, National University of Singapore.


preview/soc-fyp-report/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "soc-fyp-report"
```
saecki: If this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

saecki (commented):
Thanks for this package!

saecki:
This package is also in the wrong location. It should be under `packages/preview` instead of `preview`.

tzyian:
Thanks for reviewing!
I have renamed the package and moved it to the correct location.
Have checked it compiles locally.

saecki:
Thanks!

PR #4681, opened 2026-04-24
axiomst:0.2.1

rezaarezvan:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Patch release for `axiomst` with fixes for slide pause handling and counter resets across subslides.


rezaarezvan:
There are some things failing regarding the name of the image and how I link it in the README.

Let me know if this is okay or if I should fix it :).

elegaanz:
The link warning is here to encourage you to link to a version of the file that will always reflect the state of the package for that specific version. You can either replace `main` with a git tag name or a commit hash, or replace the full URL with just `thumbnail.png` since you happen to ship that file with the package anyways. By doing so, even you update your repository in the future (while developing the next version for example) the README will display correctly on Typst Universe and be faithful to what the package actually does in its latest published version.

Regarding the image alternative text, I also feel like the warning is relevant, since it doesn't really convey the same meaning as looking at the image. I would suggest replacing it with something like "Two example slides using axiomst, featuring what could be a math class, with a well spaced layout and colorful boxes for theorems and other notable sections." (feel free to edit that sentence as much as you like, it is only a suggestion to give an idea of what a good alternative text can contain).

Also regarding the template thumbnail: [it should depict a single page of the template](https://github.com/typst/packages/blob/main/docs/manifest.md#:~:text=The%20thumbnail%20must%20depict%20one%20of%20the%20pages%20of%20the%20template%20as%20initialized). Can you please fix that? You can still keep the current image as a nice header for your README.

rezaarezvan:
> The link warning is here to encourage you to link to a version of the file that will always reflect the state of the package for that specific version. You can either replace `main` with a git tag name or a commit hash, or replace the full URL with just `thumbnail.png` since you happen to ship that file with the package anyways. By doing so, even you update your repository in the future (while developing the next version for example) the README will display correctly on Typst Universe and be faithful to what the package actually does in its latest published version.
> 
> Regarding the image alternative text, I also feel like the warning is relevant, since it doesn't really convey the same meaning as looking at the image. I would suggest replacing it with something like "Two example slides using axiomst, featuring what could be a math class, with a well spaced layout and colorful boxes for theorems and other notable sections." (feel free to edit that sentence as much as you like, it is only a suggestion to give an idea of what a good alternative text can contain).
> 
> Also regarding the template thumbnail: [it should depict a single page of the template](https://github.com/typst/packages/blob/main/docs/manifest.md#:~:text=The%20thumbnail%20must%20depict%20one%20of%20the%20pages%20of%20the%20template%20as%20initialized). Can you please fix that? You can still keep the current image as a nice header for your README.

Alright, understood, I think I addressed all the things in the latest commit :)

elegaanz:
Perfect, thank you!

PR #4684, opened 2026-04-24
zebraw:0.6.2

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fix irregular zebra striping and HTML-only style/script injection while keeping the existing Typst compiler compatibility.


hongjr03:
close this pr for skipping this version

PR #4693, opened 2026-04-25
cetz-mechanical-system-34j:1.1.1

34j:
I am submitting
- [ ] a new package
- [x] an update for a package

Support latest cetz


preview/cetz-mechanical-system-34j/1.1.1/README.md
elegaanz, line 8: The import here should use the actual version.

preview/cetz-mechanical-system-34j/1.1.1/renovate.json
elegaanz, line 1: This file is only useful for development, can you please delete it from this copy of the package?

elegaanz (commented):
The package is in the wrong directory, it should be in `packages/preview/` and not `preview/` (this is a bit confusing because the repository itself is called `packages`, so there are two times `packages` in the full path).

PR #4694, opened 2026-04-25
ts-upc-thesis:0.1.0

ttOwwA:
I am submitting
- [x] a new package
- [ ] an update for a package


packages/preview/ts-upc-thesis/0.1.0/README.md
```diff
@@ -0,0 +1,98 @@
+# ts-upc-thesis
+
+中国石油大学（华东）本科毕业设计（论文）Typst 模板，独立框架，开箱即用。
+
```
saecki, line 4: It is generally appreciated to have at least an introduction sentence in English, even if the target audience for the template is not native English speakers. This allows the broader Typst community to get an idea of what the package is about and make sure that it isn't of interest to them.

packages/preview/ts-upc-thesis/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "ts-upc-thesis"
```
saecki, line 2: Is ts part of the official name of the university/program, or is it a non-descriptive prefix? If it's the latter what does stand for?

If this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

saecki (commented):
Thanks for this package!

packages/preview/ts-upc-thesis/0.1.0/typst.toml
ttOwwA: thanks for the feedback.

I've renamed the package from `ts-upc-thesis` to `modern-upc-thesis` to better comply with the naming rules. 
 I've also added an English introduction sentence at the top of the README as suggested.

The updated PR(#4715) will replace this one. Please let me know if anything else needs adjustment.

PR #4703, opened 2026-04-26
eggs:0.7.0

retroflexivity:
I am submitting
- [ ] a new package
- [x] an update for a package

This is eggs' largest release yet. it is now powered by [elembic](https://github.com/PgBiel/elembic), which means a few things:

- It's now somewhat more stable, without non-convergence cases and footnote bugs caused by overly complex system of contexts and state updates;

- All config options can now be passed to the corresponding functions (example, subexample, gloss) directly or set with [elembic's set rules](https://pgbiel.github.io/elembic/elements/styling/set-rules.html) (of course, the old way still works);

- Updates to the config are now properly layered, without options being reset to default; e.g. using `#show: eggs.with(indent: 1em)` and then `#show: eggs.with(spacing: 2em)` will preserve the 1em indent setting;

- The codebase is now overall cleaner and more comprehensible.

As for real features,

- Option `subexample-wrapper` has been added, which allows controlling how list-derived subexamples are placed (e.g. in a grid).


elegaanz:
Thanks for the update! Cool to see another package build on elembic :)

PR #4706, opened 2026-04-27
minimal-note:0.10.1

Michel-Liao:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Bumps `minimal-note` from 0.10.0 to 0.10.1.

- Add `show_outline` named argument to control whether the table of contents is rendered (default: `true`)         
- Add `codly` and `codly-languages` support for code block styling                                        
- Add bibliography support (template now ships with `refs.bib`)                                                    
- Fix README example: corrected `outline:` → `show_outline:` and bumped import version                             


saecki:
Thanks!

PR #4707, opened 2026-04-27
letterloom:2.0.0

nandac:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Enclosures are now specified as dictionaries with a required description field, replacing the previous plain-string format. An optional file field (bytes loaded via read) renders the attachment on a dedicated page after the letter body, with pages controlling how many pages to embed and margin controlling per-page margins for the attachment.

This is a breaking change: existing plain-string enclosure entries must be migrated to the new dictionary format. See the migration guide in CHANGELOG.md.

Before this change, enclosures were just a list of names — the reader had to trust that the documents were physically included. Now the letter and its attachments compile into a single self-contained PDF. That matters for a few reasons:

* No manual PDF merging. Previously, you would write the letter in Typst, export it, then stitch attachments together in a separate tool. Now it's one typst compile, and you're done.

* The list and the pages stay in sync. If a document is named in the enclosures list, its pages are guaranteed to follow — no risk of listing something you forgot to attach, or attaching something you forgot to list.

* Selective page embedding. You can pull just the relevant pages from a larger report rather than dumping the entire thing. That keeps the letter package tight and professional.

* Per-attachment margin control. A scanned image and a typeset PDF may need very different margins. Each attachment can be tuned independently.

* Richer descriptions. Switching from plain strings to dictionaries makes the description field available for Typst content, so you can bold a title, add emphasis, or link to a section — all within the enclosure list itself.

In short, it turns letterloom from a letter template into a complete correspondence package.


nandac:
Closing pull request to fix errors.

PR #4710, opened 2026-04-27
cap-able:0.0.2

SchrodingerBlume:
I am submitting
- [ ] a new package
- [x] an update for a package

Bug fixes:
• `captab-style` / `capfig-style` / `cap-style` switched to patch semantics. Every parameter
now defaults to `auto`, and only the fields you explicitly pass get written back to
`state`. This means you can call `captab-style.with(...)` repeatedly to override one
dimension at a time (e.g. enabling outline-bilingual) without resetting omitted
fields (e.g. cell-inset) to their function defaults. Show / set rules now read the
latest merged values via context { `state.get()` } at render time.
New features:
• `captab` and `captab-style` gain `top-rule` / `middle-rule` / `bottom-rule` parameters for customising
the three-line table’s top, middle, and bottom rules. Accept any
Typst stroke value, e.g. `2pt + red`, `stroke(thickness: 1pt, dash: "dashed")`. Defaults
remain `1.5pt` / `0.5pt `/ `1.5pt`.
• captab gains a `columns` parameter as the preferred alias of `cols`, matching Typst’s
native table(`columns: ...`) naming. cols is still accepted as a back-compat alias.
Documentation updates:
• Complete parameter tables include `top-rule` / `middle-rule` / `bottom-rule` / `columns`.
• Added patch-semantics callouts before captab-style and capfig-style sections.
• Added `== Three-Line Strokes` subsection with usage examples.
• All `cols:` examples rewritten as `columns:`.

### DIFF
--- a/packages/preview/cap-able/0.0.2/LICENSE
+++ b/packages/preview/cap-able/0.0.2/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 SchrodingerBlume (Andrew Junhao Wu)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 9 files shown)


elegaanz:
Thanks.

PR #4711, opened 2026-04-27
deal-us-tfc-template:1.1.1

dayala1:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed figure numbering to correctly update counters to 0. Slightly adjusted default colors


elegaanz:
Thank you.

PR #4712, opened 2026-04-27
Disable non-local `ignore::Walk` features

saecki:
This fixes a potential issue, where non-local `.gitignore`s, git `exclude`s, or the global `.gitconfig` might change how packages are bundled. I've rebundled all packages on my machine with and without this change, and the output was the same.


elegaanz:
Thank you!

PR #4718, opened 2026-04-27
minerva-thesis:0.2.2

lvandevelde:
I am submitting
- [ ] a new package
- [x] an update for a package

- some bug fixes
- additional arguments for configuring the figures, subfigures and captions 


packages/preview/minerva-thesis/0.2.2/typst.toml
```diff
@@ -0,0 +1,26 @@
...
+    "Ghent University",
+    "UGent",
+    "Universiteit Gent"
+]
+categories = [
+    "model",
+    "layout",
+    "thesis",
+]
+compiler = "0.14.2"
```
saecki: Apparently I didn't catch this last time, but the `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the update!

packages/preview/minerva-thesis/0.2.2/typst.toml
lvandevelde: Indeed, the `compiler` field was updated to the version I currently use, which is probably not needed. I will check if it runs under 0.14.0.

lvandevelde: It runs under 0.14.0 and updated the `compiler` field accordingly.

PR #4722, opened 2026-04-28
classic-thesis-uct:0.1.1

james-hepworth:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update to version 0.1.1. Includes a fix for bold text, layout updates to tables and equations, and adds an affiliation statement to the readme.


saecki:
Thanks!

PR #4742, opened 2026-04-30
simple-hust-report:0.1.1

kkkkkkeng:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: it's an update.Add description about fonts in readme and cross-platform fonts support when missing specific fonts.


saecki:
Thanks!

PR #4743, opened 2026-04-30
modern-xmu-thesis:0.2.2

HPCesia:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fix page number type of abstracts in outlines.


elegaanz:
Thank you!

PR #4752, opened 2026-05-01
arnoptical:0.1.0

adam-rocska:
Adds the initial `@preview/arnoptical:0.1.0` package.

This package selects Arno Pro optical text faces from the current Typst text size according to Arno Pro's documented optical-size ranges. It does not bundle font files; users must install Arno Pro separately under their own font license.

Validation run locally / by CI:

- `typst-package-check check @preview/arnoptical:0.1.0`


adam-rocska:
@saecki this is my first contrib to this ecosystem, so please tell me if anything is not ok, or should be done better.

One thing that's weird for me but apparently it's the Typst Universe way of doing things is the copy-paste of code. For now I decided to play along and do the copy-pasta; but the original lives on https://github.com/adam-rocska/arno-pro-optical-text-typst and my tech-instinct would have been to simply submodule the origin repo here, commit push pr and call it a day :-/ 

My discomfort with the current approach is the usual wrath of copy-pasta duplications. It's easy to drop the ball.

Your two cents on this?

packages/preview/arno-pro-optical-text/0.1.0/typst.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "arno-pro-optical-text"
```
saecki: This name is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

saecki (commented):
Thanks for this package!

> One thing that's weird for me but apparently it's the Typst Universe way of doing things is the copy-paste of code. For now I decided to play along and do the copy-pasta; but the original lives on https://github.com/adam-rocska/arno-pro-optical-text-typst and my tech-instinct would have been to simply submodule the origin repo here, commit push pr and call it a day :-/

> My discomfort with the current approach is the usual wrath of copy-pasta duplications. It's easy to drop the ball.

> Your two cents on this?

The fact that Typst Universe currently uses a github repository to store packages is more incidental and for convenience than for anything else. The usual approach is to version your code in a separate repository and treat opening a PR to the `packages` repository as part of your release process, just like you would when publishing a cargo or npm package. The files that should be published to typst universe also differ from the ones you might want to have in the development repository, see: https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude
A common way to do this is with a release script or external tools, but at some point we want to have better built in tooling for publishing packages.

packages/preview/arno-pro-optical-text/0.1.0/typst.toml
adam-rocska: Thanks for the review.

I understand the intent of the naming rule, and I agree that names like `optical-text`, `optical-sizing`, or `font-optical-sizing` would be too canonical because they would describe a general package category.

However, I think `arno-pro-optical-text` is materially different. This package is intentionally not a general optical-sizing package. It only targets Arno Pro, and specifically maps Arno Pro’s documented optical-size text faces — Caption, SmText, Regular, Subhead, and Display — according to point size. It will not grow into a generic solution for other fonts.

So the descriptive part here is not claiming the general category; it is constraining the scope. A user looking for optical sizing in general should not infer that this package handles arbitrary fonts, while a user looking specifically for Arno Pro optical text faces should be able to identify it unambiguously.

My reading of the naming rule is:
- `optical-text` would be canonical and unfair.
- `arno-pro-optical-text` is specific to one typeface and one behavior.
- The name matches the package’s intentionally narrow API and avoids suggesting broader functionality.

Would that interpretation be acceptable, or do you consider any fully descriptive name canonical even when it is this narrowly scoped?

adam-rocska: I want to avoid giving this package a name that makes it look more general than it is.

A generic optical-sizing package would imply that it can make correct typographic decisions across arbitrary fonts. This package cannot and should not claim that. It is only correct for Arno Pro because it relies on Arno Pro’s documented optical-size ranges and face names.

So the current name is intentionally specific: it communicates both the supported typeface and the supported behavior. Renaming it to something more creative would make it less searchable and less accurate, while renaming it to something broader would overstate the package’s scope.

That is the main reason I would prefer to keep `arno-pro-optical-text`.

saecki: I agree with most of what your saying, `optical-sizing` or `optical-text` would indeed be to broad of a name for a package like this. `arno-pro-optical-text` is indeed a very fitting name for this package, but that's precisely what the naming rules are trying to circumvent. A package name shouldn't be completely descriptive. There should be either a personal, otherwise non-descriptive part, that would make the name less canonical, [often times an adjective](https://typst.app/universe/search/?kind=templates). Or the whole name should be more creative, like a related word (e.g. `meander`, `alchemist`), or a combination of words, we could just now think of two examples for this package: `arno-opt` or `arnoptical`. 

To address your searchability concerns: the description will be displayed right next to the name, and the `keywords` array is also searched, so anyone searching for `arno-pro` or `optical-sizing` will find the package, no matter how it's named.

adam-rocska: `arnoptical` is great! If you’re fine with the name, I’m happy with it.
I’m more used to the dry side of software, so pardon me for the lack of creativity.

If `arnoptical` gets the green light, I’ll do the update and we’re good to go. Your call 🩵

saecki: Feel free to pick `arnoptical` :)

adam-rocska: Renamed.

adam-rocska:
@saecki I'm done with the rename. The diff tools say it's fine ¯\\_(ツ)_/¯

saecki:
Thanks!

adam-rocska:
Thank you, it was my pleasure.

PR #4754, opened 2026-05-01
synkit:0.0.3

guilhermegarcia:
Bug fix release.

- Greek letters in subscripts/superscripts now render correctly.
- `tree()` now has a `show-refs` argument to help with labels


packages/preview/synkit/0.0.3/typst.toml
```diff
@@ -0,0 +1,13 @@
...
+categories = ["utility", "text", "visualization"]
+disciplines = ["linguistics"]
+name = "synkit"
+version = "0.0.3"
+exclude = ["gallery/"]
+keywords = ["linguistics", "syntax", "semantics", "morphology"]
+entrypoint = "lib.typ"
+authors = ["Guilherme D. Garcia <guilherme.garcia@lli.ulaval.ca>"]
+license = "MIT"
+description = "A toolkit to create syntactic representations in Typst"
```
saecki: Seems like we've missed this before, but the description generally shouldn't contain the word Typst, since it somewhat redundant. See: https://github.com/typst/packages/blob/main/docs/manifest.md#writing-a-good-description

saecki (commented):
Thanks for the update!

PR #4758, opened 2026-05-02
cap-able:0.1.0

SchrodingerBlume:
I am submitting
- [ ] a new package
- [x] an update for a package

# Added

## Tables (`captab` / `captab-style`)

- **Breakable tables** — new `breakable` parameter (default `true`). Long tables now flow naturally across pages instead of being clipped on a single page. Pass `breakable: false` to restore the old "atomic" behavior.
- **Repeat header on page break** — new `repeat-header` parameter (`true` / `false` / `<int>`, default `true`). Re-emits the markdown header row at the top of each continuation page using Typst's native `table.header(repeat: ...)`. `<int>` repeats only on the first N continuation pages.
- **Repeat caption on page break** — new `continued-caption` parameter (bool, default `false`). When enabled, each continuation page gets a "Cont. Table X.Y caption" row at its top, in the same format as `refer-to` mode. Number is anchored to the main table; no figure re-registration. Caption row and markdown header live in two separate `table.header`s (level 1/2), so `continued-caption: true` and `repeat-header: false` can coexist.
- **`width` parameter** — new `width` (accepts `auto` / `<length>` / `<ratio>`, e.g. `8cm`, `80%`, `80.5%`). Available on `captab` per-call and `captab-style` global. Replaces the previous fixed-100% behavior.
- **`three-line-table` toggle** — set to `false` to drop the manual three rules and use Typst's default table stroke instead. Custom `hlines` / `vlines` still work.
- **`cell-inset` / `inset` aliases** — both names now work in both `captab` and `captab-style`; `cell-inset` wins when both are passed.
- **`captab` accepts `align` and arbitrary `..extra-args`** — passes through `fill` / `stroke` / `gutter` / `column-gutter` / `row-gutter` / `rows` etc. to the underlying `table(...)`, matching tablem's advanced-usage surface.

## Captions (`cap-style` / `captab-style` / `capfig-style`)

- **`caption-position`** (`top` / `bottom`) — controls whether the caption sits above or below body. Defaults to `top` for tables, `bottom` for figures. Configurable globally via `cap-style` / `captab-style` / `capfig-style`, or per-call via `captab(caption-position: ...)` / `bicap(position: ...)`.
- **`caption-align`** — controls horizontal alignment with five values: `"center"` (default) / `"left"` / `"right"` (table-local) / `"text-left"` / `"text-right"` (text-area-local). Also accepts a dict `(main: ..., continued: ...)` to set main and continuation captions independently.
- **`caption-text`** — pass-through dict for the caption's `text(...)` rendering. Accepts any Typst native `text()` parameter (`font` / `fill` / `tracking` / `spacing` / `stretch` / ...). Two forms:
  - *Flat dict* (`(font: "Times")`) — applies to the whole caption.
  - *Nested dict* — uses any of `whole` / `prefix` / `supplement` / `number` / `body` as keys for layered overrides; inner overrides outer.
  Example: `caption-text: (whole: (font: "Helvetica"), number: (fill: red, weight: "bold"), body: (font: "SimSun"))`. Lets users style supplement, number and body independently — solving the "show figure.caption rule doesn't take effect" problem (cap-able's caption is manually rendered, not a real `figure.caption` element).
- **`placement`** — Typst-style float positioning. Four values: `none` (default, in-flow) / `top` / `bottom` (float to top/bottom of current or next page) / `auto` (Typst picks `top` or `bottom` — whichever edge is closer). Hidden figure (cross-ref anchor) rides along inside the float wrapper, so `@ref` resolves to the float's landing page.

## bicap

- **`bicap()[body]` form** — wrap a caption with a body in a single non-breakable block, conceptually a cap-able-flavoured `figure(body, caption: ...)`. The trailing-block form and the `body: [...]` named-arg form are equivalent.
- **`bicap(continued-caption: true)`** — when `kind: "table"`, every native `#table()` inside body gets its caption repeated on each continuation page. Same mechanism as `captab(continued-caption: true)`.
- **`bicap(repeat-header: ...)`** — overrides the markdown header `repeat` setting of any `#table()` in body. Independent of `continued-caption`.
- **`bicap(breakable: ...)`** — controls whether the outer block may break across pages (default `true`, so a breakable body flows naturally).

## Subfigures

- **`label-style` accepts `str | dict`** — passing a string keeps overlay label and subcaption prefix coupled (default behavior). Passing a dict `(overlay: "a)", subcaption: "(a)")` decouples them; missing keys fall back to package default `"(a)"`.
- **Per-subfig overlay style override** — `label-style-override` dict now accepts four new keys for individual overlays: `style` (the format string), `font`, `size`, `offset`. The previously supported visual keys (`text-color`, `stroke`, `bg`, `bg-shape`, `bg-radius`, `bg-inset`) still work.
- **`subcaption-number-title-spacing`** (default `auto`) — controls the gap between the subcaption number prefix and its body. `auto` inherits from the main caption's `number-title-spacing`. Configurable on `capfig-style` and per-call on `capsubfig`.
- **Visibility-aware cross-reference letter** — when `label-style` differs between overlay and subcaption (dict form), `@ref` follows whichever side is actually visible. If both are visible (typical), subcaption wins; if only the overlay is drawn, overlay wins.
- **Per-subfig `style` cross-ref consistency** — when ref follows overlay (subcaption hidden), `@ref` renders that subfig's own format so the letter matches what's stamped on the figure.

## Numbering (`numbering-format`)

- **Function-form support** — `numbering-format` now accepts `str | function`, matching Typst's native `numbering()` first argument. Function form receives all heading levels + the figure/table number when `use-chapter: true`, just the figure/table number when `use-chapter: false`.

# Changed

- **Default table width: `100%` → `auto`** — tables now size to content by default, not the full text width. Set `set-table-width(width: 100%)` or `captab-style.with(width: 100%)` to restore the old behavior. Tables with explicit `fr` columns still expand to fill their container.
- **Default subcaption number-text spacing** — changed from a single space to inheriting the main caption's `number-title-spacing` (e.g. `: ` or `\u{3000}`). Affects only users with `show-subcaption: true` and `show-subcaption-label: true`. To restore the old single-space, set `subcaption-number-title-spacing: " "` explicitly.
- **`use-chapter: false` now honors `numbering-format`** — previously this branch hard-coded `numbering("1", num)`, silently ignoring custom formats like `"(A)"` or `"App. 1"`. Both branches now respect the configured format. *Bug fix; no real-world content depended on the old behavior.*

# Fixed

- **`captab` align parsing** — passing an already 2D alignment (e.g. `align: horizon + center`) no longer panics with "cannot add a vertical and a 2D alignment". The auto-`+ horizon` is only applied when `.y == none`.
- **`numbering-format` UTF-8 boundary** — multi-byte literals in format strings (e.g. `"附 1"`) no longer panic on byte-level indexing. Format scanning now uses grapheme clusters (`.clusters()`).

# Known Limitations

- **`continued-caption: true` + `caption-position: bottom`**: incompatible. Continuation repetition relies on `table.header(repeat: true)` to emit the caption inside the table; using `table.footer` instead would lock the caption inside the column boundaries — visually "embedded in the table" — which contradicts what `caption-position: bottom` means. When both are set, `continued-caption` is silently disabled. Use `caption-position: top` if you need every-page repetition, or stick with `continued-caption: false` and let the original caption appear once at the bottom of the last page.
- **`continued-caption: true` + `caption-align: "text-left"` / `"text-right"`**: the continuation-side text-anchored values silently degrade to `"left"` / `"right"`. The continuation caption is structurally locked inside `table.header`'s column width and can't extend out to the text-area edges. For true text-anchored continuation captions, fall back to manual `refer-to` splitting (one `captab(refer-to: <main>, caption-align: "text-left", ...)` per continuation segment).
- **`continued-caption: true` + multiple tables in one `bicap`**: the same caption is injected above every native `#table()` in body. Convention is one table per `bicap`; for multi-table layouts, use multiple `bicap` calls or manage each table's caption with `captab` directly.
-  **`placement: top` / `bottom` / `auto` forces `breakable: false`** — Typst floats can't break across pages. A long table that doesn't fit a float will overflow; keep `placement: none` for those. Also incompatible with `continued-caption: true` (which depends on page breaks).

### DIFF
--- a/packages/preview/cap-able/0.1.0/LICENSE
+++ b/packages/preview/cap-able/0.1.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 SchrodingerBlume (Andrew Junhao Wu)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/preview/cap-able/0.1.0/README.md
+++ b/packages/preview/cap-able/0.1.0/README.md
@@ -1,0 +1,188 @@
+<p align="center">
+  <strong>cap-able</strong><br>
+  <em>Make it more able to caption — now that is cap-able</em>
+</p>
+
+<p align="center">
+  <img src="https://img.shields.io/badge/version-0.1.0-blue" alt="version: 0.1.0">
+  <img src="https://img.shields.io/badge/license-MIT-green" alt="license: MIT">
+  <img src="https://img.shields.io/badge/typst-0.13.0+-orange" alt="minimum typst version: 0.13.0">
+</p>
+
+<p align="center">
+  <a href="#english">English</a> | <a href="#中文">中文</a>
+</p>
+
+<p align="center">
+  📖 <a href="https://github.com/SchrodingerBlume/typst-cap-able/releases/download/v0.1.0/doc-en.pdf">English Documentation</a> · 📖 <a href="https://github.com/SchrodingerBlume/typst-cap-able/releases/download/v0.1.0/doc-zh.pdf">中文文档</a>
+</p>
+
+---
+
+## English
+
+A comprehensive Typst package for creating professional three-line tables and figures with bilingual caption support, continued tables/figures, subfigures, and flexible customization options for academic documents.
+
+### Features
+
+- **Three-line tables** — Academic-standard tables (top, middle, bottom rules) with Markdown syntax
+- **Bilingual captions** — Auto-formatted bilingual captions (Chinese/English, German/English, Spanish/English, etc.)
+- **Continued tables/figures** — Multi-page tables and figures with automatic numbering
+- **Subfigures** — Flexible multi-image grid layouts with overlay labels and subcaptions
+- **Notes** — Auto-width-matched notes for tables and figures
+- **Multilingual support** — 25+ languages including RTL languages and Traditional/Simplified Chinese
+- **Unified configuration** — Configure both via `cap-style`, or override per-type with `captab-style` / `capfig-style`
+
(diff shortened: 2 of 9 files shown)


saecki:
Thanks!

PR #4759, opened 2026-05-02
basic-report:0.5.0 and clean-dhbw:0.5.0

roland-KA:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- Contains the new declaration for AI usage
- Update of the Declaration of Authorship and the Confidentiality Statement according DBHW-Leitlinien as of 25.11.2025


roland-KA:
Some tests failed. But it is not clear to me what the reasons are.

elegaanz:
Some are explained [here](https://github.com/typst/packages/pull/4759/checks?check_run_id=74050941955). If this explanation doesn't make sense to you, we can discuss it.

The "Test / Test (pull_request)" check failed because of the `basic-report` package that was likely added by mistake. It is in the wrong sub-directory (there is no version in the path) and it makes the test workflow fail. I think the fix here is just to remove it.

roland-KA:
Ah thanks! The main reason was that the new 0.5.0-version of `basic-report` wasn't in the respective sub-folder but in the main folder of that template (I have an update for `basic-report` as well as for `clean-dhbw` inlcuded within this PR). 

I have fixed that, hoping that everythink works now.

roland-KA:
... and the 0.5.0-folder for clean-dhbw was somehow not complete; I've fixed that too

roland-KA:
The remaining warnings can be ignored since I didn't change the content where these URLs point to.

elegaanz:
Thank you for fixing these issues. Regarding the remaining warnings, the idea is that you may change the content of these files in the future, for example while developing a new version of the package on the `main` branch, but before it gets released to Typst Universe. Then, it would be a bad experience to click on one of those links and see a PDF document with a different layout or an example file that uses a new API.

For the design documents, I think it can make sense to link to `main`. Or at least, it is not too bad if the user reads a more recent version of the document.

If you really don't expect to make such breaking changes in the future and/or think it's fine to link to `main` in your case, it is also possible to just ignore these warnings, but I wanted to give a bit more context to make sure you understand what they stand for, so that you can make a conscious choice about this issue. So, let me know what you prefer :)

roland-KA:
Hi @elegaanz, thanks for the explanations! I think for some of the links it makes sense to adapt them for upcoming versions.

PR #4760, opened 2026-05-02
eqrun:0.1.0

snlxnet:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: `eqrun` runs math equations: it inserts variables into them, calculates the result, displays it and saves it into a variable for use in other equations. It makes it so you don't have to write the equation at least twice: in code mode to calculate and in math mode to display.


packages/preview/eqrun/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "eqrun"
+version = "0.1.0"
+keywords = ["equation", "runner", "evaluation"]
+categories = ["scripting", "visualization", "utility"]
+entrypoint = "lib.typ"
+authors = ["<@snlxnet>"]
+repository = "https://github.com/snlxnet/eqrun"
+license = "MIT"
+description = "Evaluate equations, use the results in further equations."
+exclude = ["*.svg", "example.typ"]
```
saecki, line 11: The example images should also be added to the `exclude` list here. They will still be available for display on Typst Universe, but won't need to be downloaded when importing this package. See https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

packages/preview/eqrun/0.1.0/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```typst
+#eqrun($A = w dot h$)
+```
+![Output of the example block that uses variables from init](./example_1.svg)
+
+Now you can use the new variable in other equations:
+
+```typst
+Half of that is:
```
saecki: I think some of the code snippets here are outdated.

saecki (commented):
Thanks for this package!

packages/preview/eqrun/0.1.0/typst.toml
snlxnet: Doesn't `*.svg` already exclude them?

packages/preview/eqrun/0.1.0/README.md
snlxnet: Thank you!

packages/preview/eqrun/0.1.0/typst.toml
saecki: Oh, yes of course. I missed that :sweat_smile:

PR #4762, opened 2026-05-03
ntust-thesis-unofficial:1.0.1

8LWXpg:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

- Move bibliography to function argument.


packages/preview/ntust-thesis-unofficial/1.0.1/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "ntust-thesis-unofficial"
+version = "1.0.1"
+entrypoint = "src/lib.typ"
+authors = ["8LWXpg"]
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the update!

packages/preview/ntust-thesis-unofficial/1.0.1/typst.toml
8LWXpg: I have changed the required compiler version to 0.14.0

PR #4765, opened 2026-05-03
chordx:0.7.0

ljgago:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

- Added `inner-gap` and `outer-gap` parameters in `single-chord` to adjust the chord name offset.


elegaanz:
Thank you!

PR #4773, opened 2026-05-04
non-unlabeled:0.1.0

ravasioluca:
I am submitting
- [x] a new package
- [ ] an update for a package

This package prevents unlabeled objects from being numbered. This helps keeping the look of the pdf much cleaner. Example:
```typst
#import "@preview/non-unlabeled.typ": *

// You need to do this to enable the package
#show math.equation: dont_number_unlabeled(math.equation)
#show figure: dont_number_unlabeled(figure)

// You also need to define how you number your objects (you can do this also without using headcount)
#import "@preview/headcount:0.1.0": *
#set ref(supplement: [Eq.])
#set math.equation(numbering: dependent-numbering("(1.1)", levels: 1))
#set figure(numbering: dependent-numbering("(1.1)", levels: 1))

#set heading(numbering: "1.")

= Suspicious additions

This equation is unimportant: it will not be numbered:
$
3 + 3 = 4
$

This one, on the other hand, will be used later:
$
3 + 3 = 7
$ <label>

Therefore it will be numbered. The following two equations instead won't.

$
3 + 3 = 42
$

$
3 + 3 = 24
$

This also needs to be numbered
$
3 + 3 = 777
$ <label2>
 ```

### DIFF
--- a/packages/preview/non-unlabeled/0.1.0/LICENSE
+++ b/packages/preview/non-unlabeled/0.1.0/LICENSE
@@ -1,0 +1,22 @@
+MIT License
+
+Copyright (c) 2026 Luca Ravasio
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+"Software"), to deal in the Software without restriction, including
+without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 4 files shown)


elegaanz:
Thank you.

PR #4775, opened 2026-05-05
letterloom:3.0.0

nandac:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
#### Letterhead sender layouts (sender-position, sender-valign, bottom-gap)

Before v3.0.0, a letterhead image was purely decorative — it occupied space at the top of the page, but the sender's address was always placed in the body of the letter. Many real letterheads integrate the sender address directly inside the branded zone (left column, right column, or beneath the logo). These three keys make that possible without manual workarounds.

#### Signature bin-packing

The previous version could not handle multiple signatures with long affiliations and would squash it making it unreadable. This version calculates the available content width and lays out the signatures either side-by-side if there is enough space or on the next row.

#### em-based spacing

All internal spacing values (signature row gap, name-to-affiliation gap, cc and enclosure indentation) were previously fixed lengths. If a user changed `main-font-size`, those gaps no longer scaled proportionally, breaking the visual rhythm of the letter. Converting to em units automatically ties spacing to the chosen font size.

#### Removal title field from signatures

The title and affiliation fields were redundant — both conveyed a role or organisational line beneath the signatory's name, and having two separate fields forced users to choose which to use. Merging them into a single affiliation field that accepts free-form content (including multi-line blocks) simplifies the API and gives users full control over how that information is formatted.


elegaanz:
Thank you.

PR #4783, opened 2026-05-05
sanor:0.2.1

pacaunt:
I am submitting
- [ ] a new package
- [x] an update for a package

This update introduces a new API for the animation specification that I think is more flexible and useful. Content reveal control is now done by `s.push(rule())` where `rule` can be `apply`, `once`, `revert` etc., the rule that indicates how to show the content. This method allows modification to integrate with the conventional `pause` functionality, which I added as an alternative for any untagged content to participate in the animation sequence. The tagged element is now has their own state that can be reset, restore, or modify through the rules, which means the rule of showing content can be reversed, or redo again and again.


elegaanz:
Thank you.

PR #4787, opened 2026-05-06
ansi-render:0.9.0

8LWXpg:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

* Added support for `CSI 22`

### DIFF
--- a/packages/preview/ansi-render/0.9.0/CHANGELOG.md
+++ b/packages/preview/ansi-render/0.9.0/CHANGELOG.md
@@ -1,0 +1,102 @@
+# Changelog
+
+## [0.9.0] - 2026-05-06
+
+### Added
+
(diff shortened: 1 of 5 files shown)


elegaanz:
I would appreciate if you could fix the warnings, at least the ones about image descriptions. The alternative text doesn't have to be very long, but it should explain what the image illustrates to someone using a screen reader (or other assistive technology), or someone with a slow internet connection.

The ones about linking to the main branch are less important, but using a commit hash or git tag in the URL instead would guarantee that the links and image stay the same, and correspond to the latest published version, even if you push updates to your repository (while developing the next version for example).

8LWXpg:
I have changed the image alt text and updated URL to tagged version

elegaanz:
Thank you!

PR #4788, opened 2026-05-06
eqrun:0.1.1

snlxnet:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: fixed double result when the equation is a statement (like `$a = 1$`), fixed missing value insertion, fixed fractions, added the unit field.


saecki:
Thanks!

PR #4793, opened 2026-05-06
deal-us-tfc-template:1.2.0

dayala1:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: New version of the template adds fancy referenciable TODOs with optional TODOs outline, makes all secondary outlines optional, and removes unnecessary style changes (including the use of an external library). 


elegaanz:
Thank you :)

PR #4800, opened 2026-05-07
ornamentalyst:0.1.0

thuiop:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: this is a Typst port of the [pgfornament](https://ctan.org/pkg/pgfornament) LaTeX package, using only standard library functions. The primary entry point is the `ornament` function, which simply draws the ornament based on embedded .pgf files; it features easy resizing, mirroring and coloring.

Note: I think I have checked all the conditions, with a caveat on the first one. I went with `ornament` to mirror the name of the LaTeX package, but it might be too general. Let me know if that is the case and I will figure out another one.


packages/preview/ornament/0.1.0/typst.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "ornament"
```
saecki: This name is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

saecki (commented):
Thanks for this package!

packages/preview/ornament/0.1.0/typst.toml
thuiop: Yes, this is what I was worried about. Would something like "ornamentalyst" work?

saecki: That works :+1:

packages/preview/ornamentalyst/0.1.0/README.md
saecki, line 1: I didn't think of it before, but maybe the readme should mention how the included pgf files are licensed?
https://github.com/typst/packages/blob/main/docs/licensing.md

packages/preview/ornamentalyst/0.1.0/ornaments/am/am1.pgf
saecki, line 1: Shipping lots of small singular files is also not ideal, maybe you could combine them into file. See for example how `octique`, `booticons`, or `scienceicons` are storing their icon data.

saecki (commented):
Thanks!

saecki:
Thanks!

thuiop:
> Thanks!

Thanks to you for the review!

PR #4803, opened 2026-05-07
retrofit:0.2.0

xkevio:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog

- update crate versions for WASM blob
- update minimum typst version to v0.14
- update grid-show-rule to work with v0.14 by changing it to a grid-cell show-rule


saecki:
Thanks!

PR #4805, opened 2026-05-07
unofficial-eirb-report:0.1.1

floriandelage:
I am submitting
- [ ] a new package
- [x] an update for a package

In this new version, I fixed reported bugs, mainly related to figure numbering, which became inconsistent when using subfigures. To address this, I integrated the package [subpar](https://typst.app/universe/package/subpar/?utm_source=chatgpt.com), which provides exactly the functionality I needed, and removed the custom figure captions and counters, as they were unnecessary and unreliable.

I also added support for a second logo on the title page and in the footers (for example, to include a company logo in an internship report). Finally, users can now override the default template assets, including the school logo, and optionally disable the abstract section, which is not always useful for shorter papers.


saecki:
Thanks!

PR #4808, opened 2026-05-07
easy-hgb-thesis:0.1.0

TimerErTim:
I am submitting
- [x] a new package
- [ ] an update for a package

I really want to start establishing Typst as a viable alternative to LaTeX at my University. As we are in the progress of starting our bachelor theses, I think a major stepping stone is a quick good to go thesis template. This should be that template, providing a sane set of defaults while being highly customizable for powerusers.


packages/preview/easy-hgb-thesis/0.1.0/typst.toml
saecki, line 1: This package is currently in the wrong location, it should be under `packages/preview` not just `preview`. Moving the package will trigger some automatic checks, so there might be more feedback :)

saecki (commented):
Thanks for this package!

packages/preview/easy-hgb-thesis/0.1.0/typst.toml
TimerErTim: Haha, was already wondering why the `new` label was missing. Stupid me...😁

packages/preview/easy-hgb-thesis/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "easy-hgb-thesis"
+description = "Opinionated bachelor's, master's thesis or protocols for the FH Upper Austria, especially campus Hagenberg."
+version = "0.1.0"
+compiler = "0.14.0"
+entrypoint = "lib.typ"
+authors = ["Tim Peko <timerertim@gmail.com> @TimerErTim"]
+repository = "https://github.com/TimerErTim/hagenberg-thesis-typst"
+license = "CC0-1.0"
```
saecki: Just wanted to make sure, did you read the licensing docs, especially the part about using the creative commons license: https://github.com/typst/packages/blob/main/docs/licensing.md

Even the  creative commons license itself recommends not using it for software: https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software

If you're interested in a more restrictive license GPLv3 might be of interest

TimerErTim: Thanks for pointing that out, I have read the licensing docs. 

I decided to continue with CC0 because even the least restrictive, globally acknowledged licenses require license and copyright preservation (at least in source form). The Boost License would be an option, but I am unsure about Typst packages counting towards source or binary distributions. In case of the former, license and copyright preservation would be mandatory for consumers. I do not want any of that. The Unlicense is not legaly recognized by every nation, especially problematic because Austria is one of these and the university the template is intented for is in Austria.

So CC0 seems to be the only one left doing what I want, apart from the recommendation not to apply it to software but otherwise functional.

I am happy for any other input though :)

saecki: Would the `MIT-0` license work?

TimerErTim: Have changed to `BSL-1.0`. That should cover most cases I think.

saecki: To me it reads like it would still require copyright preservation, since the Typst code would not be considered `machine-executable object code generated by a source language processor`. I'm not sure what other cases it covers. The "MIT No Attribution" or `MIT-0` license wouldn't require that.
Just making sure, if you've decided on the `BSL-1.0` license, that's fine :)

TimerErTim: I only knew and stumbled upon `MIT` license, but `MIT-0` indeed seems to be a better fit. Thank you :)

PR #4809, opened 2026-05-07
approximate-acmsmall:0.1.0

gasche:
I am submitting
- [x] a new package

### Description

This package imitates the [`acmart`](https://github.com/borisveytsman/acmart/) LaTeX document class from ACM, with the `acmsmall` option.

This document class is used by several conferences organized by the ACM (I mostly know of the SIGPLAN conferences). It is also a reasonable default choice of class to use for a new paper in computer science in any field.

### Checkboxes


gasche:
I fixed all the CI failures, except for the fact that the "package check" fails to compile the file because it needs `--features html` to work -- the template support specialized code generation for a HTML backend. I don't know how to tell the CI to use `--features html`.

gasche:
I used the package [bullseye](https://typst.app/universe/package/bullseye) to avoid requiring the `html` feature flag when the HTML output is not requested, so hopefully the CI will now pass.

saecki:
Thanks! Since the `html` target is still experimental there currently isn't any for it in the package check and no way to specify a HTML only package.

PR #4811, opened 2026-05-08
noteworthy:0.4.0

tarunjana:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Noteworthy is a Typst template designed for creating class notes, especially for mathematics.


packages/preview/noteworthy/0.4.0/typst.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "noteworthy"
+version = "0.4.0"
+entrypoint = "lib.typ"
+authors = ["Tarun Kumar Jana <https://tarunjana.in>"]
+license = "MIT"
+description = "A Typst template for creating class notes especially for Mathematics"
```
saecki: I guess we didn't catch this earlier, but the description generally shouldn't contain the word Typst, since it is somewhat redundant. See: https://github.com/typst/packages/blob/main/docs/manifest.md#writing-a-good-description

saecki (commented):
Thanks for the update!

packages/preview/noteworthy/0.4.0/typst.toml
tarunjana: Thanks for pointing that out. It's fixed now.

saecki: Thanks

PR #4814, opened 2026-05-08
simple-csp:0.1.0

DaraMac:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: The package adds operators for the CSP language. In practise what this means is defining maths operators and specific classes on existing maths symbols to get the correct spacing and semantics. Although CSP is a niche language this package is useful to avoid redefining these any time someone wants to use it. I used to use a similar package in LaTeX so I wanted to gain the utility in Typst.


packages/preview/simple-csp/0.1.0/lib.typ
```diff
@@ -0,0 +1,17 @@
...
+#let cspleft = math.class("normal", "left")
+#let extchoice = math.class("binary", sym.square.stroked)
+#let forward = math.class("normal", "forward")
+#let hide = math.class("binary", sym.backslash)
+#let interleave = math.class("binary", sym.bar.v.triple)
+#let process = math.class("unary", "P")
+#let seq = math.class("relation", ";")
+#let then = math.class("relation", sym.arrow.r)
+#let tick = math.class("normal", sym.checkmark)
+#let traceRefinedBy = math.class("relation", $attach(subset.eq.sq, br: T)$)
```
saecki: Typst and its ecosystem usually uses kebab-case for variables. Though if you strongly prefer camelCase you can keep it this way.
I only comment on this variable, because for the other ones their case seemed to be related to the content they produced.

packages/preview/simple-csp/0.1.0/typst.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "simple-csp"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Dara MacConville <https://macconville.ie>"]
+license = "MIT"
+description = "Common CSP operators."
+keywords = ["csp", "processes"]
+categories = ["components"]
+disciplines = ["computer-science", "engineering"]
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for this package!

packages/preview/simple-csp/0.1.0/typst.toml
DaraMac: Thanks for the tip! Have dropped this down now in the next commit.

packages/preview/simple-csp/0.1.0/lib.typ
DaraMac: I lowercased the other ones with capitals to make it more consistent. I wanted to make this one kebab-case but I found it didn't play as nicely in math mode then and I couldn't figure out way around that. So for now I've just made it lower case and simpler.

saecki: Thanks!

PR #4815, opened 2026-05-08
ugm-presentation-unofficial:0.1.0 and ugm-presentation-unofficial:0.1.1

fepfitra:
Update to v0.1.1 with margin fixes and improved usage syntax


packages/preview/ugm-presentation-unofficial/0.1.1/typst.toml
saecki, line 11: Could you also add the `src/previews` directory to the `exclude` list. The images will still be available on Typst Universe for display, but wont need to be downloaded when the packge is imported. See https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

saecki (commented):
Thanks for this package!
There seems to be an issue with the automatic package check, because it found the `0.1.0` version of the package in the git history, but can't find the `typst.toml` now. I've run the checks locally, and everything seems fine though :)
Just one remark regarding the preview images.

saecki:
Hey, this PR has been stale for quite a while, and to keep things organized we're gonna close it for now. You can pick it back up at any time and reopen it :)

PR #4821, opened 2026-05-09
unofficial-ouc-bachelor-thesis:0.3.0

hongjr03:
I am submitting
- [ ] a new package
- [x] an update for a package

### Fixes

- set numbering text in math.equation to 五号 and western font
- don't display pagenum of chapter in outline
- wrong numbering in appendix figure
- weird counter


saecki:
Thanks!

PR #4828, opened 2026-05-10
lambdabe:0.1.0

iuy1:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Parse, reduce, and visualize lambda calculus expressions.


packages/preview/lambda/0.1.0/examples/basic.typ
```diff
@@ -0,0 +1,40 @@
+#import "/src/lib.typ": *
```
saecki: Since these examples are linked from the readme, they should use package specification imports (`@preview`)

packages/preview/lambda/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "lambda"
```
saecki: This name is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

saecki (commented):
Thanks for this package!

PR #4832, opened 2026-05-10
approximate-acmsmall:0.2.0

gasche:
I am submitting
- [x] an update for a package

[approximate-acmsmall](codeberg.org/gasche/approximate-acmsmall-typst-template/) provides a Typst template for the 'acmsmall' option of the 'acmart' documentclass provided by the ACM (Association for Computer Machinery), which is used as the base package to write research papers in various sub-disciplines of computer science.

Version 0.1.0 was the first version of the template. The new version 0.2.0 includes a variety of improvements coming from a careful port of the entirety of the "sample" file of the LaTeX class, exercising most LaTeX constructs that the authors considered in their class definition. In particular, it includes the following changes:

- an `acknowledgments` function to include acknowledgments
- an `appendix` function to include appendices
- more faithful rendering of figures and tables

The API of the template was reworked a bit (deviating from the API originally chosen by @michel-steuwer), it exposes a single functions that returns a dictionary with in particular a `show_` field installing the template style, and a `make-title` field to produce a title page following ACM conventions.

This means that users upgrading from 0.1.0 to 0.2.0 need to adapt their code to this new API:

```typst
/* Before: */
#import "@preview/approximate-acmsmall:0.1.0": acmart
#show: acmart.with(...template parameters...) // template style + title page

...user content...


/* After: */
#import "@preview/approximate-acmsmall:0.2.0": acmart
#let acmart = acmart(...template parameters...)
#show: acmart.show_ // the template style
#(acmart.make-title)() // the title page

...user content...
```

---


saecki:
Thanks!

PR #4838, opened 2026-05-11
niram-css:0.2.0

nandac:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
**Support different CSS color name conventions.**

CSS color names appear in the wild in several forms — the spec uses concatenated lowercase (mediumvioletred), web tooling often uses camelCase or PascalCase (mediumVioletRed, MediumVioletRed), design tools and CLIs prefer kebab-case (medium-violet-red), and prose naturally reads as spaced Title Case (Medium Violet Red). Previously, `niram-css` only accepted the canonical lowercase form, forcing users to memorize or look up the exact spelling. Inputs are now normalized internally, so any common casing or separator (including snake_case and SCREAMING_SNAKE_CASE) resolves to the same color — letting authors write color names in whatever form reads most naturally in context.

**Restructured and improved the manual and README to add more information about the package.**

The initial release shipped with minimal documentation, which left common questions unanswered: why `green` in Typst differs from `green` in CSS, why the list contains 147 names but only 138 unique hex values, and where the names came from historically. The manual now covers usage examples across fills, strokes, underlines, and shapes; a side-by-side comparison of Typst's built-in palette against the CSS equivalents; the X11/CSS history behind the names; and a complete reference table. The `README` mirrors the introduction, so users land on the same framing whether they arrive from GitHub or the PDF.


packages/preview/niram-css/0.2.0/typst.toml
```diff
@@ -0,0 +1,28 @@
+# for a description of available keys, see https://github.com/typst/packages/?tab=readme-ov-file#package-format
+
+[package]
+name = "niram-css"
+version = "0.2.0"
+entrypoint = "src/lib.typ"
+authors = ["nandac"]
+license = "Unlicense"
+description = "Brings standard CSS color names to Typst."
```
saecki: I guess we didn't catch this before, but the description generally shouldn't contain the word Typst, since it is somewhat redundant. See: https://github.com/typst/packages/blob/main/docs/manifest.md#writing-a-good-description

saecki (commented):
Thanks for the update!

packages/preview/niram-css/0.2.0/typst.toml
nandac: > I guess we didn't catch this before, but the description generally shouldn't contain the word Typst, since it is somewhat redundant. See: https://github.com/typst/packages/blob/main/docs/manifest.md#writing-a-good-description

@saecki, I have updated the description as per your guidance to "Access all 147 standard CSS color names."

saecki: Thanks!

PR #4845, opened 2026-05-12
cap-able:0.1.1

SchrodingerBlume:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

# Fixed

- User-supplied `hlines` `row` index was incorrectly shifted relative to the markdown table when `continued-caption: true` (SchrodingerBlume/typst-cap-able#8). The injected caption row at y=0 was not compensated for, so `row: 2` ended up between the caption and the markdown header. After the fix, `row` is always indexed against the markdown table itself, regardless of `continued-caption`. ⚠️ Compatibility: if you used `row: N+1` as a workaround, switch back to `row: N` — otherwise you'll get two overlapping lines.
- `subcaption-number-title-spacing` with a length value (e.g. `0.3em`) raised `"cannot join string with length"` because the length was joined into the content directly; this also suppressed hidden-figure registration, causing the chain failure where `@subfig` reported "label not exist" (SchrodingerBlume/typst-cap-able#9). Fix routes the value through `handle-spacing`, which wraps lengths/relatives into `h(...)` and passes content through.

# Added

- `subref-style` config on `capfig-style` / `capsubfig` (default `"letter"`, new `"full"`) — controls the letter style of `@subfig` cross-references. `"letter"` shows just the letter (backward-compatible, `Fig. 1a`); `"full"` keeps the `label-style` decorations (`Fig. 1(a)`). In `"full"` mode `label-sep` defaults to empty (decorations already separate visually). SchrodingerBlume/typst-cap-able#10.
- `extra-rule` config on `captab` / `captab-style` (default `0.5pt`) — default stroke for `hlines` / `vlines` entries that omit `stroke`, so you don't repeat the same stroke per line. Accepts a single value (shared by h & v) or `(h: ..., v: ...)` dict for per-axis. Per-line `stroke` still wins.
- `hlines` / `vlines` entries now accept an *int shorthand*: `hlines: (2, 3, 4)` ≡ `((row: 2,), (row: 3,), (row: 4,))`. Can mix with full dicts: `hlines: (2, (row: 5, stroke: 1pt), 7)`.

### DIFF
--- a/packages/preview/cap-able/0.1.1/LICENSE
+++ b/packages/preview/cap-able/0.1.1/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 SchrodingerBlume (Andrew Junhao Wu)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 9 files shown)


SchrodingerBlume:
Edited to fix cross-refs. Sorry for the noise!

saecki:
Thanks!

PR #4857, opened 2026-05-13
clean-cnam-template:1.6.6

TomPlanche:
I am submitting
- [x] an update for a package

**clean-cnam-template:1.6.6**

What changed:

- **`#no-numbering()`** — new function that suppresses the numbering prefix on the immediately following heading (any level). The heading retains its normal styling; only the counter is omitted. Mirrors the existing `#no-big-title()` API.

- **`#code()` `numbering: auto`** — the `numbering` parameter now accepts `auto` in addition to `true`/`false`. `auto` hides line numbers when the block contains only a single line.

- **`#code()` title API refactor** — the four separate title params (`title`, `title-align`, `title-style`, `title-inset`) have been merged into a single dict `title: (:)` with keys `font`, `size`, `fill`, and `weight`. Reduces noise at the call site and makes the API consistent with `text-style` and `number-style`.

- **Heading layout fix** — sub-headings (level 2+) are now rendered in an explicit `block` with per-level font sizes (`1.2em` down to `0.9em`) instead of relying on Typst's built-in hanging indent, which produced inconsistent indentation at deeper levels.


saecki:
Thanks!

PR #4869, opened 2026-05-15
glotter:0.1.0

rice8y:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: 

**glotter** is a Typst package for detecting the language of text fragments and applying language-aware settings to content.

>[!NOTE]
>Please note that the bundled `glotter.wasm` file is about 1016 KB.
>
>This is because the WebAssembly plugin embeds data derived from fastText's compressed language identification model, `lid.176.ftz`, in order to provide language detection without requiring external files or network access at compile time.
>
>I apologize that this is another relatively large package after `auto-jrubby`, but the size is intentional and comes from the embedded language identification model.


rice8y:
The package check reports that the `license` field should be OSI-approved for the following value:

```toml
license = "MIT AND CC-BY-SA-3.0"
```

However, the current licensing guidelines say that packages may be licensed under an OSI-approved license or a version of CC-BY, CC-BY-SA, or CC0.

In this package, the source code is MIT-licensed, while the embedded language identification data/model component is derived from CC-BY-SA-3.0 material. The README and NOTICE file clarify which license applies to which part.

Could you confirm whether `MIT AND CC-BY-SA-3.0` is acceptable here, and whether this check failure is due to the package-check license validation being stricter than the current licensing guidelines?

saecki:
Thanks for this package!

> The package check reports that the `license` field should be OSI-approved for the following value:
> 
> ```toml
> license = "MIT AND CC-BY-SA-3.0"
> ```
> 
> However, the current licensing guidelines say that packages may be licensed under an OSI-approved license or a version of CC-BY, CC-BY-SA, or CC0.
> 
> In this package, the source code is MIT-licensed, while the embedded language identification data/model component is derived from CC-BY-SA-3.0 material. The README and NOTICE file clarify which license applies to which part.
> 
> Could you confirm whether `MIT AND CC-BY-SA-3.0` is acceptable here, and whether this check failure is due to the package-check license validation being stricter than the current licensing guidelines?

I agree this is a false positive by the package check and the licensing of this package is fine.

PR #4874, opened 2026-05-16
ilm:2.1.0

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog
- `date` option will default to `datetime.today()`
- Refactored to use helper functions to improve legibility.
- Fix abstract block paragraph set rule
- Fix raw-text customization logic
- Remove deprecated options

---


elegaanz:
Thank you!

PR #4886, opened 2026-05-18
econ-working-paper:0.4.0

statzhero:
Update econ-working-paper to 0.4.0. Adds optional epigraph parameter and comments out JEL codes in the starter template.


elegaanz:
Thank you!

PR #4888, opened 2026-05-19
ankify:0.1.0

nvlang:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package, together with its [ankify crate](https://crates.io/crates/ankify) counterpart (a CLI tool), allows users to generate and sync Anki flashcards from their Typst documents.

### DIFF
--- a/packages/preview/ankify/0.1.0/LICENSE
+++ b/packages/preview/ankify/0.1.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025-present N. V. Lang
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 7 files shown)


nvlang:
By the way, I just want to say that I suspect the name `ankify` may be a bit borderline with regards to the naming rules, since it does strike me as rather canonical. The fact that there's also a bunch of different software named "Ankify" (or similar) already out there that's unrelated to this project substantiates my suspicion:

- https://www.ankify.app/
- https://ankify.krxiang.com/
- https://chatgpt.com/g/g-I9DTFOKxS-ankify
- https://notesankify.com/
- etc. etc.

However, two things made me PR this as-is:

- There's a [`typ2anki`](https://typst.app/universe/package/typ2anki/) Typst package out there, and the name doesn't strike me as all that much less canonical (maybe a little bit though)
- I couldn't think of another good name I like

Of course, if you think I should change the name, I'll happily oblige and close this PR and open a separate one once I've figured out a new name.

elegaanz:
I would say the name is okay-ish, we can accept it as is. There are already quite a lot of package with `-fy` as a suffix in their name. "Ankify" is already used by other projects, but it is not a registered trademark as far I can tell, and not really a well known name in general, so I would say it's fine. If you prefer to have a bit of time to find a better name, we can also wait a bit before merging this PR, there is no need to rush.

elegaanz:
Aside from that, I think the package looks good, so just let me know what you want to do regarding the name.

nvlang:
I think if it's okay with you I'd rather keep the name as is then. Thank you!

PR #4906, opened 2026-05-21
beautitled:0.2.0

nathan-ed:
I am submitting
- [x] an update for a package

beautitled

  v0.2.0 — 2026-05-21

  - Section number decimal-comma fix — subsection numbers now use Unicode word-joiner (U+2060) around the separator dot, preventing French/Swiss locales from
   rendering 1.2 as 1,2


packages/preview/beautitled/0.2.0/typst.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "beautitled"
+version = "0.2.0"
+entrypoint = "src/lib.typ"
+authors = ["Nathan Scheinmann"]
+license = "MIT"
+description = "Beautiful title styling for Typst documents with 19 distinctive print-friendly themes for chapters, sections, and table of contents"
```
saecki: I guess we didn't catch this before, but the description generally shouldn't contain the word Typst, since it is somewhat redundant. See: https://github.com/typst/packages/blob/main/docs/manifest.md#writing-a-good-description

packages/preview/beautitled/0.2.0/typst.toml
```diff
@@ -0,0 +1,13 @@
...
+repository = "https://github.com/nathan-ed/typst-package-beautitled"
+keywords = ["headings", "titles", "chapters", "sections", "styling", "themes", "toc", "outline", "education", "textbook"]
+categories = ["components", "layout"]
+disciplines = ["mathematics", "education"]
+compiler = "0.14.0"
+exclude = ["docs/*", "gallery/*.typ", "gallery/examples", "gallery/examples/*", "*.pdf"]
```
saecki: I think the whole `gallery` directory should be excluded here. The purpose of exclude is to have documentation files that are still available for display on Typst Universe, but won't have to be downloaded when importing the package. See https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

If you're using the `exclude` list to filter the files published to Typst Universe, you should probably find another mechanism for that, since it defeats the purpose of the having it in the manifest.

saecki (commented):
Thanks for the update!

packages/preview/beautitled/0.2.0/typst.toml
nathan-ed: Ok, correcting it now. thanks

packages/preview/beautitled/0.2.0/typst.toml
nathan-ed: Ok ,thanks

nathan-ed:
All done, thanks

packages/preview/beautitled/0.2.0/typst.toml
saecki: I think you misunderstood, it was a good thing that the `gallery/examples` weren't included in the PR. Since they aren't linked from the readme, they shouldn't be committed to this repository. It just seemed like you were using the `exclude` list to filter which files were added to the PR.
So could you please remove the `gallery/examples` again?

nathan-ed: Oh okay, yes I misunderstood what you said.

saecki:
Thanks!

PR #4914, opened 2026-05-22
ntust-thesis-unofficial:1.0.2

8LWXpg:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
- Remove supplement on level 1 heading.
- Remove all first line indent if lang is set to `en`


saecki:
Thanks!

PR #4916, opened 2026-05-22
eggs:0.8.0

retroflexivity:
I am submitting
- [ ] a new package
- [x] an update for a package

This release brings a bunch of user-facing improvements.

- Gloss lines are now split on any number of spaces, including one. Just like in TeX. This is a **breaking** change: if you used to use a single space as a non-breaking space, it will now break (not a pun).

- `print-abbreviations` now sorts alphabetically by default (**breaking**) and accepts three options: `sorted-by` (controls how abbreviations are sorted), `wrapper` (controls how every entry is printed), and `separator` (controls what is printed between the entries).

- also **breaking**: abbreviations _top_ and _int_ have been renamed to *top_* and *int_* to not conflict with Typst's built-ins. There are also several new abbreviations available.

- There's a new function, _trailing_, which mimics expex's `\\trailingcitation`: typesets a text on the right,on the next line if it doesn't fit.

- Auto judges are now a lot smarter, with content parsing instead of a regex show rule. This mainly means that they are now properly only recognized at the beginning of an example / gloss word, and not after styled content as before.

- Docs now include a chapter on how-tos that might be useful to a linguist.

- Errors about gloss line length mismatches now trace up to the document again (regression in 0.7.0).

**Full Changelog**: https://github.com/retroflexivity/typst-eggs/compare/0.7.0...0.8.0


saecki:
Thanks!

PR #4920, opened 2026-05-23
clean-math-paper:0.2.7

JoshuaLampert:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added some clarifications in the README.md and made "Proof of" translatable. Thanks @Horhik for the contribution!


JoshuaLampert:
I see CI is failing because I use a non-permalink in my README, but I want to direct a user to the most recent version of that file, i.e., I usually do not want to have a link with a specific tag/release/commit and I do not want to update the link each time I make a change. Is there a recommended way how to deal with this?

packages/preview/clean-math-paper/0.2.7/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+[Typst](https://typst.app/home/) paper template for mathematical papers built for simple, efficient use and a clean look.
+Of course, it can also be used for other subjects, but the following math-specific features are already contained in the template:
+
+- theorems, lemmas, corollaries, proofs etc. prepared using [great-theorems](https://typst.app/universe/package/great-theorems)
+- equation settings
+
+## Set-Up
+
+The [template](https://github.com/JoshuaLampert/clean-math-paper/blob/main/template/main.typ) is already filled with dummy data, to give users an impression what it looks like and to showcase some of its features. The paper is obtained by compiling `main.typ`.
```
saecki: > I see CI is failing because I use a non-permalink in my README, but I want to direct a user to the most recent version of that file, i.e., I usually do not want to have a link with a specific tag/release/commit and I do not want to update the link each time I make a change. Is there a recommended way how to deal with this?

In this case you can just use a link to the local file. Which is generally preferred if possible.
```suggestion
[template](template/main.typ)
```

saecki (commented):
Thanks for the update!

packages/preview/clean-math-paper/0.2.7/README.md
JoshuaLampert: Ah, I didn't know it was smart enough to recognize the local link. That's good to know. Thanks!

PR #4931, opened 2026-05-26
catiche:0.1.0

gzordrai:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: An unofficial template for University of Lille internship reports.


packages/preview/ulille/0.1.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "ulille"
```
elegaanz: If this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

packages/preview/catiche/0.1.0/assets/ulille.png
elegaanz, line 1: Can you please double-check that you are allowed to distribute these logos? If so, can you specify in your README (or some other file) under which terms they can be used and/or under which license they are placed. Otherwise, can you replace them with a placeholder image and instructions on how to manually download and import the logos?

packages/preview/ulille/0.1.0/typst.toml
gzordrai: Sure, I wasn’t sure if the name would be okay, so I’m going to change it :)

packages/preview/catiche/0.1.0/assets/ulille.png
gzordrai: Well, regarding the license, there is no specified license. The logo can be downloaded publicly from the university website or from other sites such as Wikipedia. Is it okay, or would a placeholder be preferable to avoid any risk?

elegaanz: The Wikipedia version of the logo [seems to be in the public domain](https://commons.wikimedia.org/wiki/File:Logotype_Universit%C3%A9_de_Lille_2022.svg?uselang=fr#Conditions_d%E2%80%99utilisation), but the University website seems to suggest that [it is a trademark of the University](https://www.univ-lille.fr/mentions-legales#:~:text=publication%2E-,Les,usage). In that case, I think adding a disclaimer in the README, stating that the files you are using come from a public domain source (assuming you copied them from Wikipedia) and that it is the responsibility of users to check that they are allowed by the University to use them should be enough to be safe.

gzordrai: That sounds like a good idea, I'll do that.
Thanks for the advice :)

elegaanz:
Looks good, thank you :)

PR #4933, opened 2026-05-26
exercise-bank:0.5.0

nathan-ed:
I am submitting
- [x] an update for a package

- **g-exam integration** — the `exam-question`, `exam-question-many`, and `exam-select` wrapper functions have been removed, along with the unconditional `@preview/g-exam` import. The `display-mode: "exam"` flag (which shows points in the badge via `exo-show`) is retained and requires no external dependency.


nathan-ed:
Updated the package as per the warnings. Checks are not triggered again for a reason I don't understand.

elegaanz:
The reason probably is that GitHub Actions are having issues at the moment (see https://www.githubstatus.com/).

elegaanz:
Thank you for the update :)

PR #4940, opened 2026-05-27
typed-smiles:0.1.1

GeronimoCastano:
I am submitting

- [ ] a new package
- [x] an update for a package

Changes:

- Fixed layout for molecules with separated ring systems. Previously, a later ring could be placed as if it still belonged to the first ring, causing long crossing bonds between unrelated ring closures.
- Fixed vertical reaction-arrow labels so `above` appears to the right of an up/down arrow and `below` appears to the left.
- Cleaned up README examples so code snippets match the rendered images, including removing redundant raw text captions.


saecki (commented):
Thanks for the update!

PR #4941, opened 2026-05-27
meander:0.4.3

Vanille-N:
I am submitting
- [ ] a new package
- [x] an update for a package

Fixed two bugs:
- Meander would try to split tables but the algorithm was inadequate. Table splitting is disabled until I find a heuristic that works.
- A check to detect containers that would make `measure` not work was too strict, leading to content failing to go to the next page.


elegaanz:
Thank you! :)

PR #4944, opened 2026-05-27
isc-hei-bthesis:0.7.9, isc-hei-document:0.7.9, isc-hei-exec-summary:0.7.9, isc-hei-poster:0.7.9, isc-hei-report:0.7.9 and isc-hei-tb-assignment:0.7.9

pmudry:
I am submitting
- [x] a new package
- [x] an update for a package

Description: Update of the isc-hei-* templates. The templates were updated for this new academic year and the code has been vastly refactored. Added a new isc-hei-poster package, which is also an official one.


pmudry:
Sorry, I just noticed a problem in my toolchain for releasing the package that was force-pushing my changes in the fork. 

Those updates caused a lot of trigger in the updates. The files are still sound and can be reviewed as is. I am terribly sorry for the unnecessary noise. The problem is solved now.

packages/preview/isc-hei-bthesis/0.7.9/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "isc-hei-bthesis"
+version = "0.7.9"
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
> Sorry, I just noticed a problem in my toolchain for releasing the package that was force-pushing my changes in the fork.
>
> Those updates caused a lot of trigger in the updates. The files are still sound and can be reviewed as is. I am terribly sorry for the unnecessary noise. The problem is solved now.

No problem :)

Thanks for the update!

packages/preview/isc-hei-bthesis/0.7.9/typst.toml
pmudry: Thanks for pointing that out. Compiler version field has been now set to `0.14.0` (and screenshots size reduced along the way)

saecki: Thanks!

PR #4960, opened 2026-05-29
aero-dhbw:0.2.0

cocoveal:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: `aero-dhbw` is a lightweight thesis template for DHBW Ravensburg. This update (0.2.0) adds support for **multiple authors** (up to six): the `author` argument now accepts either a name string (single author, unchanged) or an array of `(name, mat-number, course-acronym)` dictionaries. Multiple authors are rendered on the title page (each with their own student-ID row) and in the Declaration of Authorship (plural wording with a signature per author, laid out in a two-column grid). The title page and declaration were also de-duplicated into a single shared bilingual (German/English) layout, and the cover now uses flexible spacing so it fits on one page regardless of author count.


saecki:
Thanks!

PR #4966, opened 2026-05-31
distro:0.1.0

stochastical:
I am submitting
- [x] a new package
- [ ] an update for a package

A pure Typst library for common probability distributions. Each distribution module provides a constructor, PMF/PDF and CDF. Some modules also include inverse-transform sampling and `mean`/`variance` fields. Allows for literate programming for statistics education.

  - Explanation:
    
   `distro` is a riff off the word `distribution`, but it's not the canonical name for distributions, or even a common contraction. Think "bistro", but for maths instead of food :)
- [x] [`excluded`](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude) PDFs or README images, if any, but not the LICENSE


stochastical:
I've fixed the relative imports for `../../lib.typ` for the example docs and replaced them with package imports (I think that's the correct pattern here?).

I can see that the remaining warnings are on image captions in the `README.md`, however, I think the captions are sufficient (just flagging as warnings due to length). But happy to take on board any feedback! Thank you!

saecki:
Thanks for this package!

> I've fixed the relative imports for ../../lib.typ for the example docs and replaced them with package imports (I think that's the correct pattern here?).

Yep, that's correct.

> I can see that the remaining warnings are on image captions in the README.md, however, I think the captions are sufficient (just flagging as warnings due to length). But happy to take on board any feedback! Thank you!

The alternative text should describe the images in a way that makes them accessible to visually impaired people. The single word captions would not be sufficient for that. They are currently warnings because it's just a heuristic, but it would be appreciated if you could come up with a short (one sentence) description for them :)

stochastical:
> Thanks for this package!
> 
> > I've fixed the relative imports for ../../lib.typ for the example docs and replaced them with package imports (I think that's the correct pattern here?).
> 
> Yep, that's correct.
> 
> > I can see that the remaining warnings are on image captions in the README.md, however, I think the captions are sufficient (just flagging as warnings due to length). But happy to take on board any feedback! Thank you!
> 
> The alternative text should describe the images in a way that makes them accessible to visually impaired people. The single word captions would not be sufficient for that. They are currently warnings because it's just a heuristic, but it would be appreciated if you could come up with a short (one sentence) description for them :)

Thanks! Just updated the captions :)

saecki:
Thanks!

PR #4969, opened 2026-06-01
simple-research-poster:0.2.1

aneziac:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: After seeing someone fork the repo to make changes to parts that are currently inaccessible to the end user (and receiving similar feedback from my friend), I realized that I needed to expose every Typst file that's used. Now, every Typst file is cloned locally, making more extensive modifications possible.


elegaanz:
The way you originally designed the package was better fitting Typst current approach to templates. It is expected that not everything is customizable, and that some parts of the template should not be editable by the user, letting them focus on content. However, we are also aware that this can be fairly limiting, especially for more visual templates such as this one. [As explained in the docs](https://github.com/typst/packages/blob/main/docs/typst.md#template-customization) this will become better when custom elements will be implemented. In the meantime, a middle ground can be found by providing more customization parameters and/or by re-exposing smaller building blocks that can then be imported and composed together by the user. You may also want to have a look at `elembic`, which allows to get a preview of what custom elements could do, even if the syntax is a bit strange at times, and that it's API won't necessarily match what will be implemented natively in the compiler.

In any case, exposing all of your package as a template that can entirely be modified is not really possible in the current state of things.

aneziac:
Thanks for your comment. I was initially thinking about doing what you suggested and just including more parameters, but I concluded that it would be impossible to cover all potential use cases, so exposing everything was most logical. For a highly visual project like this, total control seems to be the best user experience.

I am aware of the coming custom element improvements and generally how to approximate these capabilities now via `elembic`. In fact, I've been actively refactoring the other package I wrote using elembic (which is an extremely impressive project, hats off to PgBiel), and will publish an update soon. To me, that doesn't seem like the right way to write this particular template.

I'm wondering if you could clarify what exactly the technical limitations with the approach I took here are. `typst init` seems to work fine locally, and intuitively I don't see a reason to enforce a fetch to `@preview/simple-research-poster`, especially when it seems like I can move code arbitrarily between which files are cloned and which are not. But I guess from the failing test and your comment, something in Typst internals prevents this. Ideally, I would like to expose as much as possible.

elegaanz:
The CI errors come from the fact that the `package.entrypoint` doesn't exist. But you could make it point at an empty `lib.typ` file and it would work fine. The reason we don't allow that is not so much technical than "logical": the point of using a template is to have some abstractions handled for you, and to not have to worry about layout and styles too much, allowing you to focus on content. Making every file part of the template also makes it impossible to update from version to version. It can also be a bit confusing or overwhelming to have all those files as a user, given that files in your project generally *have to* be edited, which is not necessarily true here.

If your goal is to provide a quick start project setup that is supposed to be arbitrarily tweaked, a Git repository that people can clone or a ZIP archive would be better suited tools. The only disadvantage is that it won't appear on Typst Universe, but with a bit of advertising in `#showcase` on Discord and/or on the forum it can still get visible.

You are not the first to ask for fully editable templates, so we might reconsider this rule at some point. However, this will likely be after we introduce custom types to see how they help to solve these kind of issues in practice. For now, the rule exists, and I encourage you to either distribute the template by some other means, or to refactor it again to still provide some abstraction to the user.

aneziac:
Thank you for taking the time to explain this. I don't necessarily agree with Typst's position on this, but I see the logic and coherence.

Not to belabor my point, but I just want to briefly make an argument in favor of fully editable templates:
- Most pressing is the practical aspect. At the end of the day, my goal with this project is to make it easy for people to create high quality posters, even if they haven't used Typst before. Several people at my university used this template this year, and they asked me how to do things like add a second logo, or slightly decrease the height of the footer, or add trim. I realized that adding more and more parameters to handle this was unsustainable, and poor programming practice besides. By not allowing me to just expose the source file, the best solution to make these types of adjustments is probably to just fork my repo, like this guy did: https://github.com/LukasManinger/simple-research-poster. This significantly raises the barrier to entry. So, while I agree in principle that templates should handle some abstraction for you, I think for more complex use cases, the current structure of templates becomes an obstacle, rather than a convenience.
  - As another example: my friend recently had an issue with a certain font in the `amsarticle` template. The issue was caused by a set rule in the `lib.typ`, which he found only after being forced to dig through the source, and was solved by just copying that file locally and removing the line. This is bad user experience.
- The versioning argument also seems a bit dubious to me. I agree with you in the case of packages, but don't think the logic follows with Typst templates. To me the purpose of versioning, besides being able to update your package, is to concretize code as a black box. Typst templates ship a mixture of immediately customizable code, and non-customizable code. Because part of the source code is modifiable, the "black box" aspect breaks down, in my opinion. One consequence is that updating becomes significantly less desirable in many cases.
- I do think your point on confusion for the user makes a lot of sense. I'm just spitballing here, but I would favor something like an opt-in flag to `typst init`, or a checkbox on `typst.app`, that allows you to clone every file. This way, more advanced users can make tweaks, while the normal experience is unaffected.

This should be solvable with custom types when they arrive, but think it would add unnecessary complexity. I definitely want to distribute this via Typst Universe.

Since I can't make this change, I will just close this PR, because I think the template is in a good state otherwise. Thank you for your patience.

PR #4971, opened 2026-06-01
min-book:1.5.0

mayconfmelo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
This version focuses on implementing a "draft mode" with tools that assist in writing and revising books.
These tools only work when the `#book(cfg.draft)` option is active; otherwise, they will be ignored in the
final output.

- Added: `#book(cfg.draft)` option
- Added: `#comment` for draft annotations
- Added: `#mark` highlights text in draft mode
- Added: `#event` declares an event in the story (draft)
- Added: `#scene` declares a scene in the story (draft)
- Added: Comment outline in draft mode
- Added: Event timeline in draft mode
- Added: List of scenes in draft mode


packages/preview/min-book/1.5.0/src/themes/elegance/cover-image.png
mayconfmelo, line 1: I used `oxipng --strip safe --alpha cover-image.png` to try to reduce the image size, but it's still over 1 MB.

elegaanz: At some point, you would have to reduce the resolution to get a smaller file. It's not high over the limit, we can accept it as is, no worries :)

mayconfmelo: I would do that, but since it's the default cover image for one of the themes, a low resolution would be very detrimental.

Perhaps in the future I'll change the image to something smaller. Anyway, thank you very much.

PR #4975, opened 2026-06-02
htlium:1.6.0

DeltaAT:
I am submitting
- [ ] a new package
- [x] an update for a package

- added more styling to the main page
- added custom and premade boxes
- added codeblocks
- made styling of e. g. tables better
- updated and fixed README


elegaanz:
Thank you.

PR #4979, opened 2026-06-02
verseatile:0.1.1

switchlex:
I am submitting
- [ ] a new package
- [x] an update for a package

- New features
  - Made the distance between verse numbers and the poem configurable via `#verse-number-distance.update()`.
- Fixes
  - Reworked verse numbers to prevent them causing issues with indentation in certain constellations.
- Documentation
  - Updated the manual.


elegaanz:
Thank you.

PR #4981, opened 2026-06-02
echarm:0.4.0

lublak:
I am submitting
- [ ] a new package
- [x] an update for a package

- added theme support with build-ins
- add language support with build-ins
- update ctxjs to 0.5.0
  - eval-later is now value.eval
  - improved performance


lublak:
The last error is still like the old one: The one big file. Can be ignored

elegaanz:
Thank you.

PR #4984, opened 2026-06-03
deixis:0.1.1

inspiros:
I am submitting
- [ ] a new package
- [x] an update for a package

**Description:**
- Added a geometric path pullback calculation for routed links when `link-marks` is enabled, ensuring thick lines stay hidden inside the arrowheads.
- Improved the `#deixis-note-outline` rendering engine. It now allows multi-line note bodies to wrap naturally without clipping, and aligns page numbers to the bottom.
- Fixed a panic caused by backlinks on celibate (mark-only) endnotes.
- Fixed an issue where multi-page region marks would calculate incorrect vertical boundaries on non-A4 paper sizes by dynamically resolving Typst's native `auto` margins.
- Add a showcase and link to source code on all example images when clicked.

### DIFF
--- a/packages/preview/deixis/0.1.1/CHANGELOG.md
+++ b/packages/preview/deixis/0.1.1/CHANGELOG.md
@@ -1,0 +1,23 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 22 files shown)


saecki:
Thanks!

PR #4985, opened 2026-06-03
ilm:2.1.1

talal:
I am submitting
- [ ] a new package
- [x] an update for a package

## Changelog
- Only add external link highlighting if `type(it.dest) == str`.

---


elegaanz:
Thank you!

PR #5010, opened 2026-06-08
to-stuff:1.0.0

nikmennega:
I am submitting
- [x] an update for a package

Description:

### Changed

- **Breaking:** less strict, more natural handling of integers and floats.
	- Integer and float conversions may accept integers, floats or decimals.
	- Float and generic number conversions may default to an integer, float, or decimal.
	- Integer conversions may only default to an integer (or `none`).

### Added

- New general-purpose function `stuff` to guess most appropriate conversion.
- Expose new conversions:
	- bool
	- decimal
	- version

### Removed

- **Breaking:** remove deprecated `quiet` argument from all direct conversion functions.
- **Breaking:** remove deprecated `scalar` function alias.


saecki:
Thanks!

PR #5019, opened 2026-06-09
touying-au-community:0.1.0

A-CHRI:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Aarhus University presentation template built with the touying package

  - Explanation: Used the standard naming convention for touying templates
    

touying-au/0.1.0/typst.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "touying-au"
```
elegaanz: If this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. `touying` is not considered to be a distinctive prefix. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

elegaanz (commented):
The package should go in `/packages/preview/`, not directly at the root. Can you move it please? Once you will have done that, the CI will run and will likely report a few warnings that will have to be fixed before the package can be merged. Some of these warnings can also be ignored if you consider they don't make sense: for example the version shield will likely get one because of the short image alternative text, but in that context it's fine and it can be skipped.

Anyways, thank you for submitting this template :)

touying-au/0.1.0/typst.toml
A-CHRI: I have marked the package as 'unofficial' - hope this is okay

A-CHRI: After browsing some of the other touying themes, I have elected to use the keyword 'community' instead, due to the authors being enrolled and active at the university

elegaanz:
Looks good, thank you!

PR #5023, opened 2026-06-09
beautitled:0.2.6

nathan-ed:
I am submitting
- [x] an update for a package

#### Added
  - `part-fullpage: true` (new default): LaTeX-style dedicated part page, content
    vertically and horizontally centred — each style has its own coherent part renderer
  - `#part(fullpage: false)`: per-call override for inline parts
  - `#part(image:, image-caption:, image-position:)`: optional cover image with
    native figure/caption support, placeable above or below the title
  - New `gallery/parts/` showing 4 representative styles


elegaanz:
Thank you!

PR #5029, opened 2026-06-11
clean-othaw:0.3.5

ruebe5w:
I am submitting
- [ ] a new package
- [x] an update for a package

Fix some bugs


packages/preview/clean-othaw/0.3.5/template/main.typ
```diff
@@ -0,0 +1,136 @@
+#import "@preview/clean-othaw:0.3.4": *
```
elegaanz: Can you update the import here please? Otherwise new projects will still import the previous version.

elegaanz (commented):
Thank you for this update, just a small thing:

packages/preview/clean-othaw/0.3.5/template/main.typ
ruebe5w: Oh yes, thank you! Always forgetting :D

elegaanz:
Thank you.

PR #5044, opened 2026-06-12
bhs-school-bundle:0.2.0

ststolz:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This template provides a standardized layout for diploma theses (Diplomarbeiten) at the vocational high school HAK/HAS and Kolleg Imst in Tyrol, Austria. It simplifies the writing process for students by ensuring all formal and regional formatting requirements are met out of the box.

  - Explanation:
    The name `hak-imst` contains a non-descriptive regional prefix and suffix, specifically identifying the targeted educational institution rather than generic terms like "thesis" or "diploma-template".


packages/preview/hak-imst/0.2.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "hak-imst"
+version = "0.2.0"
+entrypoint = "lib.typ"
+description = "Diplomarbeiten an der HAK/HAS und dem Kolleg Imst normgerecht verfassen."
```
saecki: Could you either add an, or replace the current one with an english description?

packages/preview/hak-imst/0.2.0/template/typst_media/logos/Logo_HAK_Imst.png
saecki, line 1: Do you have the rights to distribute the logos? If so, can you please state that explicitly in your README, and say that they are not covered by the MIT license. If not, can you replace them with placeholders and instruct people on how to import the actual logo themselves, please? See https://github.com/typst/packages/blob/main/docs/licensing.md

packages/preview/hak-imst/0.2.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "hak-imst"
```
saecki: >  The name hak-imst contains a non-descriptive regional prefix and suffix, specifically identifying the targeted educational institution rather than generic terms like "thesis" or "diploma-template".

The name `hak-imst` is still quite canonical, especially considering most users will be studying at the school.

saecki (commented):
Thanks for this package!

packages/preview/hak-imst/0.2.0/typst.toml
ststolz: Thanks for support. I tried to find a less canonical name that says what the package is supposed to be.

ststolz:
Thanks for support. I resolved all requests. Is it necessary to resolve the failing check?

packages/preview/bhs-school-bundle/0.2.0/README.md
```diff
@@ -0,0 +1,92 @@
...
+- `fontsize` (length): Base font size
+- `sectionnumbering` (string or `none`): Heading numbering format
+- `paper` (string): Paper format (e.g. `a4`)
+- `margin` (dictionary): Page margins
+- `responsible_default` (content): Default responsible person for the footer if not set per chapter in content.typ
+- `eidesstattliche_erklaerung_text` (content or `none`): Declaration of Authorship
+- `abnahmeerklaerung_text` (content or `none`): Acceptance Certificate
+- `vorwort_text` (content or `none`): German abstract
+- `abstract_text` (content or `none`): English abstract
+- `project_partner_logo_path` (string or `none`): Path for the left cover-page logo
```
saecki: I just noticed the documentation here is outdated, the function seems to take the images as content directly instead of the paths. Note that with Typst `0.15` and the introduced `path` type, the function could take an image path. Though accepting content is arguably more flexible.

packages/preview/hak-imst/0.2.0/typst.toml
saecki: The current name would work.
I initially only replied to your message and didn't give it much thought. You seem to be a teacher at the institution, which would allow making this an official package and keeping the original name. In that case, you or another staff member should send an email to hello@typst.app stating that you officially endorse this package. In practice it means that you will accept submission of works built using this template, ideally promote it, and guarantee that it will be maintained in the long term.

If this shouldn't be an official package we can just continue with the new name.

I'm sorry for the back and forth.

saecki:
> Thanks for support. I resolved all requests. Is it necessary to resolve the failing check?

Most diagnostics are worth fixing, though some warnings might be false positives and don't need to addressed. Thanks for taking the time to fix all of them :)

packages/preview/hak-imst/0.2.0/typst.toml
ststolz: Yes I am a teacher at this school. But I would prefer to reduce bureaucracy and stay with the new name. At all I think it has more value to develop on a more general level so that more schools can use the template.

packages/preview/bhs-school-bundle/0.2.0/README.md
ststolz: Thanks, fixed it - also variable names.

PR #5045, opened 2026-06-12
grayness:0.7.0

nineff:
I am submitting
- [ ] a new package
- [x] an update for a package

Updated for Typst 0.15 to also accept `path` in addition to `bytes` as arguments.


packages/preview/grayness/0.7.0/typst.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "grayness"
+version = "0.7.0"
+compiler = "0.15.0"
```
saecki: The `compiler` field is a minimum version, are the changes for supporting the `path` type breaking? Otherwise leaving it at `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the update!

packages/preview/grayness/0.7.0/typst.toml
nineff: there's nothing really breaking, I added a version check and panic message if somebody tries to use paths with versions earlier than 0.15

saecki: Thanks!

PR #5051, opened 2026-06-14
rivet:0.3.1

LordBaryhobal:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Minor bug fixes with separator lines and adding missing parameter documentation


elegaanz:
Can you fix the issues reported in CI please? The ones about outdated packages are more here as an information and can be ignored. Same for the relative package import in the manual, in this context it makes sense.

To give a bit more context about the other warnings: the image alternative texts are appreciated because they enable people with vision impairment or just a limited internet connection to still understand what images are about. They don't have to be very long but should briefly describe what can be seen on each illustration. If you are unsure of what should go in that description or not, you can imagine describing the picture to someone over the phone, it generally helps to describe just enough information without drowning into details.
Regarding the relative imports in the examples, replacing them with `@preview/…` imports allows users to just copy/paste the example and get a working document, without having to understand that they have to change the import.
And for the "Typst" word in the package description, it's just that it is considered redundant: it is obvious that this is a Typst package since it is published on Typst Universe.

LordBaryhobal:
Thanks for the detailed review :+1: 
I fixed the imports in examples to use `@preview/...`, added alt-texts to the images in `README.md`, and fixed some typos and a small bug
The older imports are left on purpose to keep this patch compatible with Typst 0.13.1 and minimize changes.
I will probably make another release later on to update everything and switch to 0.14.x, or even 0.15.x

elegaanz:
Thank you!

PR #5056, opened 2026-06-14
pillole:0.1.0

adrfantini:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Renders compact, multi-line bi-colored "pills" (GitLab-style scoped labels) — two-part key/value labels with rounded end-caps and a square-cornered, breakable middle that flows naturally across line breaks.
Useful for inline status/priority/environment markers in reports, slides, and documentation where regular badges would feel out of place or wouldn't break across lines.

  - Explanation: "pillole" is the Italian word for pills/tablets. The name avoids the canonical English terms (`pills`, `labels`, `badges`) while staying faithful to the visual metaphor (pill-shaped).


adrfantini:
I'm sorry for the mess, I decided to change the name of the package after sleeping on it. All should be final now.

Bots are flagging the `manual.typ` file, but this file is excluded so it should not matter.

packages/preview/pillole/0.1.0/typst.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "pillole"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Adriano Fantini <@adrfantini>"]
+license = "MIT"
+description = "Multi-line bi-colored GitLab-style scoped labels (pills)."
+repository = "https://github.com/adrfantini/typst-scoped-pills"
+keywords = ["label", "badge", "pill", "gitlab", "scoped-label"]
+categories = ["components"]
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the package!

> I'm sorry for the mess, I decided to change the name of the package after sleeping on it. All should be final now.

Don't worry, this just goes to show you care about the package :)

> Bots are flagging the manual.typ file, but this file is excluded so it should not matter.

Yes that's fine, the false positive is a little unfortunate, but often times excluded files are examples, so it's hard to differentiate.

packages/preview/pillole/0.1.0/typst.toml
adrfantini: Thank you! I downgraded to 0.14.0, looks fine to me. It's all right with an older version, as well as with a newer one. I could also probably set an even older version, but I cannot test it wight now

adrfantini: Actually, v0.15 even fixes a small annoying misalignement! yay!

PR #5067, opened 2026-06-15
verseatile:0.2.1

switchlex:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Already discovered an issue with nesting in the restructured code that caused issues with verse numbering when using interjections or split verses. Sorry, that I overlooked this when testing for the last version! It should be fixed now.


elegaanz:
Thanks.

PR #5069, opened 2026-06-15
json-resume:0.5.0

smur89:
I am submitting
  - [x] a new package
  - [ ] an update for a package

  `json-resume` is a strict loader for canonical [JSON Resume](https://jsonresume.org/) data. It reads a `resume.json` file, validates it against the [published
  schema](https://jsonresume.org/schema), and returns a normalised Typst dict with free-text fields coerced to `content` — ready to hand to any compatible Typst CV template.

  Useful because the JSON Resume promise of one `resume.json` across many renderers currently requires every Typst CV template to hand-translate the JSON into a Typst dict;
  this package centralises that work into a single dependency so renderers can stay focused on layout. The canonical schema is **derived at module-load time from the vendored
   upstream JSON Schema document** — the published spec is the single source of truth, not a hand-maintained mirror.

  For templates that need their own fields (theme colours, label overrides, header decorations, …) the package exposes the underlying engines (`validate` / `coerce`) and a
  lens-style API (`lens`, `lens-put`, `lens-over`, `add-field`, `remove-field`) so an extension schema can be built on top of the canonical one without forking the loader —
  and `schema-from-json-schema` lets callers translate their own JSON Schema documents into Typst schemas if they prefer that source of truth over hand-rolled combinators.

    - Explanation: The name mirrors the [JSON Resume](https://jsonresume.org/) project name because the package's sole purpose is to load documents conforming to *that
  specific standard*. It identifies the protocol the package implements, not a generic category — common JSON-Resume tooling across other ecosystems follows the same naming
  pattern (e.g. `jsonresume-cli`, `npm:resume-cli`, `python:json-resume`). Generic terms like `resume` or `cv` are deliberately avoided so they remain available for template
  packages.

### DIFF
--- a/packages/preview/json-resume/0.5.0/README.md
+++ b/packages/preview/json-resume/0.5.0/README.md
@@ -1,0 +1,377 @@
+<h1 align="center">json-resume</h1>
+
+<p align="center">
+  <a href="https://typst.app/universe/package/json-resume"><img alt="json-resume on Typst Universe" src="https://img.shields.io/badge/dynamic/xml?url=https%3A%2F%2Ftypst.app%2Funiverse%2Fpackage%2Fjson-resume&query=%2Fhtml%2Fbody%2Fdiv%2Fmain%2Fdiv%5B2%5D%2Faside%2Fsection%5B2%5D%2Fdl%2Fdd%5B3%5D&logo=typst&label=Universe&color=%23239DAE&style=flat-square"></a>
+  <a href="https://github.com/smur89/typst-json-resume/releases"><img alt="Latest GitHub release version of json-resume" src="https://img.shields.io/github/v/release/smur89/typst-json-resume?style=flat-square"></a>
+  <a href="https://github.com/smur89/typst-json-resume/actions/workflows/build.yml"><img alt="GitHub Actions build workflow status on the json-resume main branch" src="https://img.shields.io/github/actions/workflow/status/smur89/typst-json-resume/build.yml?style=flat-square"></a>
+  <a href="LICENSE"><img alt="MIT license badge linking to the json-resume LICENSE file" src="https://img.shields.io/github/license/smur89/typst-json-resume?style=flat-square"></a>
+  <a href="https://github.com/smur89/typst-json-resume/stargazers"><img alt="Number of GitHub stargazers for json-resume" src="https://img.shields.io/github/stars/smur89/typst-json-resume?style=flat-square"></a>
+</p>
+
+<p align="center">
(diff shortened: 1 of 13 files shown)


packages/preview/json-resume/0.5.0/README.md
```diff
@@ -0,0 +1,377 @@
...
+`meta`. The `$schema` top-level metadata field is also accepted. See
+[jsonresume.org/schema](https://jsonresume.org/schema) for every field.
+
+## Usage
+
+`parse` is the one-call entry point. It accepts either a parsed dict
+or a Typst-root-relative path string:
+
+```typst
+#import "@preview/json-resume:0.5.0": parse // x-release-please-version
```
elegaanz, line 65: I suppose the `x-release-please-version` comment is for an automated tool to check that the version number is not outdated, but I'm not a big fan of it as it will appear in the final code snippet, which will not look so nice and may confuse some users. Other tools such as [scrybe](https://github.com/Vanille-N/scrybe) allow for similar workflows without polluting the final output. Can you please configure your current tool differently, switch to something else or just stop using it, so that this comment can be removed?

packages/preview/json-resume/0.5.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "json-resume"
```
elegaanz, line 2: I feel like the name doesn't conform to our guidelines. To me it is the canonical name for that package: if another package wanted to validate JSON-resume compliance, they would be disadvantaged because you already reserved that name. Also, `json-resume` seems to be an actual project and format, so using that name without being officially affiliated with them feels a bit shady, even if not forbidden.
For these reasons, can you please pick something a bit more unique ?

packages/preview/json-resume/0.5.0/README.md
smur89: Yes indeed, this is from googles [release-please](https://github.com/googleapis/release-please) tooling, which drives the full release pipeline for the project. I would keep this over scrybe, but adopt a style that hides the markers in the output. Agreed that it's a bit ugly in the README.

packages/preview/json-resume/0.5.0/typst.toml
smur89: Thanks for the review, I agree - I will rename and submit again as `gairm-import`

smur89: Resubmitted [here](https://github.com/typst/packages/pull/5085)

PR #5070, opened 2026-06-15
beautitled:0.2.7

nathan-ed:
I am submitting
- [x] an update for a package

  v0.2.7 — What's new

  Added
  - beautitled-header(level: N) — drop-in running-header function that queries headings by page number; avoids the
  duplicate-heading problem when using hydra with beautitled-init 
  - Custom heading numbering patterns via a numbering-pattern parameter (e.g. "1.", "A.", "I.") — supported across
  all 19 themes and via beautitled-init
  - Unnumbered headings (numbered: false) — skips the display counter but still triggers page breaks

  Fixed
  - Counter doubling in beautitled-init: intercepted headings no longer increment the native heading counter twice,
  so #set heading(numbering: "1.1") works correctly
  - French colon spacing (\u{00A0}:) applied consistently to chapter and part titles in the outline
  - Multi-level counter undo now handles variadic args correctly (no "unexpected argument" error when sections are
  present alongside chapters)


packages/preview/beautitled/0.2.7/typst.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.2.7"
+entrypoint = "src/lib.typ"
+authors = ["Nathan Scheinmann"]
+license = "MIT"
+description = "Beautiful title styling for documents with 19 distinctive print-friendly themes for chapters, sections, and table of contents"
+repository = "https://github.com/nathan-ed/typst-package-beautitled"
+keywords = ["headings", "titles", "chapters", "sections", "styling", "themes", "toc", "outline", "education", "textbook"]
+categories = ["components", "layout"]
+disciplines = ["mathematics", "education"]
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the update!

packages/preview/beautitled/0.2.7/typst.toml
nathan-ed: Thanks and you already told me that previously. Changing now.

PR #5086, opened 2026-06-16
tudelft-clean-dissertation:0.1.0

NemoAndrea:
I am submitting
- [x] a new package
- [ ] an update for a package

Unofficial TU Delft dissertation template. Designed with clean, sans-serif monochrome sensibilities and clear opinionated directory structure.

  - Explanation: the 'clean' implies personal opnion and leaves the 'tudelft-dissertation' namespace open for any official representatives of the TU Delft University

  - Possible limitation might be the TU Delft logo, which is technically not covered by the repo license


saecki:
Thanks for this package! We are aware of the limitations that come with the split of code into template and package code and want to improve things in the future. But having template packages without any actual library code is something we don't want to accept on Typst Universe. While this approach is very flexible, there is no clear path of updating to a newer version of the package, once files were modified, which somewhat defeats the point of a package manager.

This has been previously been discussed to some length here: https://github.com/typst/packages/pull/4615#issuecomment-4303086405

NemoAndrea:
I see, that makes sense; I had not considered the updating aspect. I agree that it then doesn't make sense to have it on Typst universe with the current vision. I will close the pull request for now; maybe if at some point the template becomes robust enough to not require a lot of user editing, it may become useful as a package with everything in lib.typ.

Thanks for taking the time to look at it :)

PR #5098, opened 2026-06-17
presio:0.2.0

benedict-armstrong:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Switches `media`'s file-source input to Typst 0.15's file `path` type. v0.1 required callers to pre-read bytes (`read(..., encoding: none)`) and pass a separate `placeholder: image(...)` because path arguments inside packages historically resolved relative to the package directory, not the caller. With #7555 landing in Typst 0.15, the package can now `read()` and `image()` the source on the caller's behalf, so the file-mode call collapses from four arguments to one:

```typ
#media(path("figures/demo.gif"), width: 60%)
```

Also bumps `compiler` to `0.15.0`. URL / YouTube / Vimeo handling is unchanged. v0.1.0 stays available for users on Typst 0.13–0.14.

Bug fix: `speaker-notes` deduplication is now counter-based instead of y-position-based, so multiple stacked calls on the same slide no longer race into a duplicate-attachment error.

### DIFF
--- a/packages/preview/presio/0.2.0/LICENSE
+++ b/packages/preview/presio/0.2.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Benedict Armstrong
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


elegaanz:
Thank you.

PR #5101, opened 2026-06-18
econ-working-paper:0.5.1

statzhero:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update to v0.5.1 of the **econ-working-paper** template. Adds a
`table-fontsize` option (default 10pt) that sets the table body text size,
alongside the `note` / `source` helpers for table and figure notes from the
0.5.x line. Tables now render at 10pt by default; override with `table-fontsize`.


elegaanz:
Thank you!

PR #5105, opened 2026-06-18
presio:0.2.1

benedict-armstrong:
Bugfix release of [`presio`](https://github.com/benedict-armstrong/presio-typst-package) (speaker notes + embedded media for the [presio.xyz](https://presio.xyz) viewer). No API changes.

### Fix
`media` (and, for consistency, `speaker-notes`) now key their slide sidecars on the **physical page** via `here().page()` instead of `counter(page)`.

Under touying, `counter(page)` returns the *same* value for every subslide of an overlay group (`#pause`). An element visible on multiple subslides therefore produced two `pdf.attach` calls with an identical filename, which Typst rejects with *"attempted to attach file … twice"* — making any touying deck that combined `media` with `#pause` fail to compile. `here().page()` advances per physical PDF page (which is what the viewer flips through), so each subslide page gets its own sidecar. In polylux and plain decks the two counters already agree, so behaviour there is unchanged.

This is a separate, never-published version directory; `0.2.0` is untouched.

### DIFF
--- a/packages/preview/presio/0.2.1/LICENSE
+++ b/packages/preview/presio/0.2.1/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Benedict Armstrong
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


elegaanz:
Thanks.

PR #5111, opened 2026-06-19
yak:0.1.0

lublak:
I am submitting
- [x] a new package
- [ ] an update for a package

  - Explanation:
    

packages/preview/jq/0.1.0/typst.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "jq"
```
saecki: This name is too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

saecki (commented):
Thanks for this package!

packages/preview/jq/0.1.0/typst.toml
lublak: @saecki Oh I though it was Fine because something Like https://typst.app/universe/package/jsonpath/ already exists.
But yeah I can change it

saecki: Thanks!

PR #5121, opened 2026-06-20
to-stuff:1.1.0

nikmennega:
I am submitting
- [x] an update for a package

### Added

- New `skip-strings` argument allows more selective general conversions.
- New `skip-dictionaries` argument allows more selective general conversions.

### Changed

- Deprecate converting RGB/A hex codes to color without initial hash character.
- Allow `none-from` to take a single string argument, as well as an array.


saecki:
Thanks!

PR #5124, opened 2026-06-20
sheetwise:0.1.0

Hyperrick:
This adds `sheetwise:0.1.0`, a Typst-native imposition helper for arranging print work on press sheets.

We need this kind of package quite often in our own print workflows, and I did not find one Typst package that covered the whole workflow: repeated Typst-authored designs, finished PDF input, saddle-stitch booklet spreads, cut-and-stack numbering, asymmetric margins, explicit mark-only regions, and the usual print guides such as crop, registration, bleed, safe, fold, color, file header, and page-border marks.

A few details that are especially important for the package:

- one public renderer, `impose(job, ...)`, with small job constructors such as `repeat`, `variants`, `sequence`, `pdf`, `booklet`, `marks-only`, and `calibration`
- two documented input modes: typeset content directly in Typst, or impose an already-finished PDF
- side-specific margins for asymmetric printer or finishing needs
- independent mark regions, so multiple crop-mark regions can exist on one sheet
- registration marks default to 4C process black, separate from normal crop mark color
- saddle-stitch page planning with blank-page handling, binding direction, fold marks, and creep compensation

The README is written as usage documentation, not only as a short package summary, because the terms matter for people who are not doing print imposition every day.

Validation run locally:

- `cd authored/sheetwise && sh tests/run.sh` with Typst 0.15.0
- `cd authored/sheetwise && PATH=/tmp/sheetwise-typst-0.14.0/typst-aarch64-apple-darwin:$PATH sh tests/run.sh` with Typst 0.14.0
- `cd package-index/typst-packages && git diff --check upstream/main...HEAD`
- `cd package-index/typst-packages/packages && typst-package-check check @preview/sheetwise:0.1.0`


saecki:
Thanks!

PR #5126, opened 2026-06-20
shuxuejuan:0.2.0

VWWVVW:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
- Refactored most of the code
- Added `sxj-term` for better vertical alignment of `tag` and `body`
- Added answer control


packages/preview/shuxuejuan/0.2.0/typst.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "shuxuejuan"
+version = "0.2.0"
+compiler = "0.14.2"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the update!

packages/preview/shuxuejuan/0.2.0/typst.toml
VWWVVW: Thanks for the reminder, I've changed it into `0.14.0`.

packages/preview/shuxuejuan/0.2.0/typst.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "shuxuejuan"
+version = "0.2.0"
+compiler = "0.14.0"
+entrypoint = "/src/lib.typ"
```
saecki, line 5: Using an absolute path here is quite uncommon, but valid and work with the compiler. The bundler fails to properly resolve the path, but that's a false positive. I've opened #5158 to fix this, but we'll have to wait for that to be merged before accepting this package.

EDIT: It's merged now. Could you rebase onto `main`, so the CI passes?

packages/preview/shuxuejuan/0.2.0/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Manual (First Page)
+
+![Compiled `manual.typ` (first page only).](./docs/outputs/manual-p1.png)
+
+Complete [source code to this manual](./docs/manual.typ), [pre-compiled pdf](./docs/outputs/manual.pdf?raw=true).
+
+## Update Strategy
+
+To avoid unnecessary version bumps[^1], all updates are promptly pushed to the [origin repository](https://github.com/VWWVVW/shuxuejuan), and synced to [typst/packages](https://github.com/typst/packages/blob/main/packages/preview/shuxuejuan) after they're tested. Small updates, including documentation fixes and code changes that don't affect the output, are typically synced with a multi-week delay.
```
saecki: That the packages are stored in this location in the git repository is more of an implementation detail. We might move away from github as a backend completely at some point, so I'm not sure this should be actively promoted here. Could you instead just say it'll be published on Typst Universe?

saecki (commented):
I didn't notice the CI was failing before. There are some more issues.

packages/preview/shuxuejuan/0.2.0/README.md
VWWVVW: I’ve replaced `[typst/packages](https://github.com/typst/packages/blob/main/packages/preview/shuxuejuan)` with `[Typst Universe](https://typst.app/universe/package/shuxuejuan)`. Hopefully, this change meets the requirement.

packages/preview/shuxuejuan/0.2.0/typst.toml
VWWVVW: Thank you XD. (I use absolute path here because I want to keep path format consistent in this package.)

packages/preview/shuxuejuan/0.2.0/README.md
saecki: Thanks!

PR #5141, opened 2026-06-23
bypst:0.4.0

jemus42:
I am submitting
- [ ] a new package
- [x] an update for a package

This release is almost exclusively focused on internal refactoring to make better use of Touying's internals rather than reinventing various wheels I realized too late didn't need reinventing. The user-facing API changes very little, but notably the theme is now more accurate to the institutional style and content overflowing to a subsequent slide now works correctly.


saecki:
Thanks!

PR #5150, opened 2026-06-23
inv-cmarker:0.1.0

ck3d:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: `to-commonmark` converts Typst content to CommonMark strings, written entirely in Typst script with no external tooling required. It provides a single `to-commonmark(content) -> str` function that performs a recursive walk over Typst's content tree, supporting headings, paragraphs, lists (unordered and ordered), block quotes, code blocks (with language info strings), inline code, strong/emphasis, links, images, and line breaks. Unsupported elements (tables, math, figures, footnotes) emit HTML drop comments rather than failing silently.

This is useful for:
- Exporting Typst-generated documentation to Markdown for publishing on GitHub, Wikis, or static site generators
- Building documentation pipelines that originate in Typst but need to be consumed in Markdown-only environments
- Content migration from Typst to Markdown-based systems

The package is written in pure Typst script with zero dependencies, making it portable and easy to integrate.

  - Explanation: The canonical name would be `typst-to-commonmark`. We use `2` as a shorthand convention common in package ecosystems (similar to `json2yaml`, `png2svg`, `csv2json`), making it a stylistic neologism rather than the most obvious English phrasing.

### DIFF
--- a/packages/preview/inv-cmarker/0.1.0/LICENSE
+++ b/packages/preview/inv-cmarker/0.1.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Christian Kögler
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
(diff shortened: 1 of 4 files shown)


packages/preview/to-commonmark/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "to-commonmark"
```
saecki: This name and the initially submitted name are not ideal. The original shouldn't really have to contain the word `Typst` and this one is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

packages/preview/to-commonmark/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
...
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Christian Kögler"]
+license = "MIT"
+description = "Convert content to CommonMark strings."
```
saecki: I see you've changed the description, probably because of the warning. But because this is a somewhat of a meta usecase, I would say it's ok to specifically say `Typst content`, which makes things a bit clearer.
See: https://github.com/typst/packages/blob/main/docs/manifest.md#writing-a-good-description

saecki (commented):
Thanks for this package!

packages/preview/to-commonmark/0.1.0/typst.toml
ck3d: The typst-package-check fails, when I add Typst to description.

packages/preview/to-commonmark/0.1.0/typst.toml
ck3d: Thanks for your feedback. My new proposal is `inv-cmarker`.

packages/preview/to-commonmark/0.1.0/typst.toml
saecki: It fails with a warning, because most of the time the word `Typst` is just redundant. Here it would improve clarity of the description. While most warnings and errors from the package check are valid, there are some that are false positives, but hard to avoid.
The description isn't that bad without it though. I just wanted to tell you, because you initially included it and then decided against it because of the package check.

saecki:
Thanks

PR #5158, opened 2026-06-24
Fix joining of paths in bundler

saecki:
Bundling failed in #5126, because the normal `Path::join` function was used to join the `entrypoint` package path to the root directory. If the entrypoint starts with a root component (`/` or `\`) the whole path will be replaced.
Introduced in `0.15` `typst-syntax` has a `VirtualPath` type that allows to `realize` paths relative to a root. This is also used inside of the compiler.


saecki:
I've run the bundler on `main` and from this PR to validate they produce the same output.

PR #5162, opened 2026-06-25
examify:0.1.2

tarunjana:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Examify is a simple typst template to create question papers for exams. This version fixed an error due to removed symbols.


saecki:
Thanks!

PR #5163, opened 2026-06-25
lemming:0.3.0

fGolke:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- feat: Emit `math-environment` and `math-proof` tags when exporting to html
- feat: Add `end-marker` parameter to `environment`. This allows putting a symbol at the end of math environments, similar to the QED symbol.
- fix: Make proofs and environments full width, such that equation-only statements display correctly
- fix: Do not insert QED symbols to figure captions, place them afterwards instead.


saecki:
Thanks!

PR #5169, opened 2026-06-25
codepoint:0.2.0

ReeseHatfield:
I am submitting
- [ ] a new package
- [x] an update for a package

Added a tmTheme file the package depends on, that we forgot to include in the last version. Nothing like testing in production.


elegaanz:
Thank you for the fix. Sorry for not catching the issue during the previous review.

PR #5172, opened 2026-06-26
phonokit:0.5.11

guilhermegarcia:
I am submitting
- [ ] a new package
- [x] an update for a package

This is a bug-fix release (0.5.11). `#vowels()` and `#consonants()` previously required a positional vowel/consonant string, so language-only calls such as `vowels(lang: "spanish")` failed with `missing argument`. The positional argument is now optional (via an argument sink), so a language can be supplied through either the positional argument or `lang:`. Existing calls are unaffected and render identically.


saecki:
Thanks!

PR #5174, opened 2026-06-26
jurlstify:0.2.1

SchrodingerBlume:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
- `show-hyphen` now governs all break opportunities.
- Behavior change: `show-hyphen` now defaults to `false`, URLs break with zero-width spaces instead of soft hyphens, Set `show-hyphen: true` to restore the previous behavior.


saecki:
Thanks!

PR #5176, opened 2026-06-27
rcv-ballot-templates:0.1.0

reesericci:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Simple RCV Ballot templates

  - Explanation:
    

rcv-ballot-templates is specialized because its not just "rcv" as a general/canonical ranked choice voting package, but rather a specific set of templates for making ballot papers.


packages/preview/rcv-ballot-templates/0.1.0/lib.typ
```diff
@@ -0,0 +1,53 @@
...
+    === Example
+
+    #set text(13pt)
+
+
+    Johnny and his friends are voting on lunch options. Johnny would be satisfied with either hot dogs or burgers, but prefers hot dogs. 
+
+    Johnny ranks hot dogs number 1 and burgers number 2, but leaves pancakes blank because he would not be happy with them.
+
+    #image("media/example.png", width: 2in, height: 2in)
```
saecki, line 46: If you're using the `media/example.png` file in your package, it shouldn't be `exclude`d. Though I think it would make more sense to have an example in the readme or as a separate file that is linked from the readme. This example file should also be added to the `exclude` list. Since it's only relevant for documentation.
See: https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

After taking another look, it also seems like you could just get rid of the image entirely and just provide the Typst code to produce the example directly.

saecki (commented):
Thanks for this package!

If you did it manually, could you please not change the PR title? The change from the package-check is intentional.

packages/preview/rcv-ballot-templates/0.1.0/lib.typ
reesericci: The file is used in the package, is it excluded?

saecki: You've excluded it in the `typst.toml` manifest: `exclude = ["media/example.png"]`

reesericci: Oops, fixed

packages/preview/rcv-ballot-templates/0.1.0/typst.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "rcv-ballot-templates"
```
saecki, line 2: I'm sorry I didn't catch this during the first round of review, but this name is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

packages/preview/rcv-ballot-templates/0.1.0/lib.typ
saecki: This is ok for now, but I'd still urge you to present the example in the way I've outlined above at some point in the future. That way people will be able to look at the code and use it as a starting point.

saecki:
Hello. Can you please update your package submission to take into account the comments above? In case you don't have time to do so for now, I will close this PR to keep our review backlog organized. You can of course open a new PR once you have fixed the mentioned issues.

PR #5179, opened 2026-06-27
elegant-polimi-thesis:0.2.1

VictuarVi:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This new releases includes the added presentation template and some fixes.


VictuarVi:
I was wondering: could I link the `/examples` directory and `docs` file in the README to the Github repo and don't push them here instead?

elegaanz:
If you prefer to, you can. But the current setup is fine. What we want to avoid is unlinked examples or manuals, as they just bloat this (already too large) repository for no good reason.

If you decide to change it to link to your own repository, make sure to use a permalink to the files (with a Git tag or commit hash in the URL) so that they always point to the correct version. Otherwise, if you push updates to your repository while building the next version, the examples could get out of sync with what is on Universe.

elegaanz:
Thank you :)

PR #5193, opened 2026-06-29
athena-tu-darmstadt-exercise:0.3.0

FussballAndy:
I am submitting
- [ ] a new package
- [x] an update for a package

This update brings the ability to set a headline, add points to tasks, modify the task layout, add sub-logos and some minor style changes and refactors. 


FussballAndy:
Would it be possible, to get a review/approval, even though some checks are failing? I would fix those once I know everything, that will have to be done. Because with that, I could combine all changes into a single PR in the repository for the template.

w-lfchen:
quick update: we're fixing the errors, this pr will probably be updated tomorrow since @FussballAndy is now asleep ^^
warnings are in the works, we need to update our (frankly quite messy and outdated publishing scripts), i'll be working on it, those will probably be addressed as well.

i do have a different question though: this template (amongst two others) is now owned by a github organization. i was unable to find something in the docs about attributing authorship to an organziation, is it possible to do that?
we'd love to do that for our three packages (athena-tu-darmstadt-exercise, athena-tu-darmstadt-thesis, not-tudabeamer-2023).

elegaanz:
Sorry for not replying earlier! This new version looks good to me, thanks for fixing the warnings. Regarding the attribution, nothing prevents you from putting `"TU Darmstadt Typst contributors <@tuda-typst>"` in the authors fields. I don't think I recall anyone doing something similar in practice, but it should work fine. Otherwise, listing contributors individually works fine too :) I'll let you make that change if you wish, and then I'll merge.

FussballAndy:
I added "TU Darmstadt Typst members" as author. Thus everybody who is added to the org should be able to make PRs. Contributors would probably be a tad too broad, as it would allow anybody who has made a commit to make a PR. 🙂

elegaanz:
Great, thank you! For information, this field is only for metadata on the Typst Universe page, the CI bot may still ask you for confirmation every time the person who submits the PR is not the same as the previous one. We will take note to check if the person submitting the package is part of the org manually, so you can ignore the pings :) We are still discussing a good way to make the bot behavior smarter, but we are aware that it's annoying as it is and want to evolve it.

FussballAndy:
Great, thanks a lot ^^
I won't mind the pings. Especially with our current update frequency. And if you are anyways aware that other members are allowed as updaters that is even better :)

PR #5194, opened 2026-06-29
typed-smiles:0.4.2

GeronimoCastano:
I am submitting

- [ ] a new package
- [x] an update for a package

Changes:

- Fix acyclic branch-point layout so the longest chain continues outward instead of folding back and overlapping an earlier group.
- Draw symmetric branch hubs (e.g. tetraethylmethane, neopentane) mirror-symmetric about the vertical axis instead of as a lopsided pinwheel, independent of SMILES atom order.
- Increase molecule subscript and superscript label size so hydrogen counts and formal charges remain legible beside atom labels.
- Fold standalone bracket hydrogens (e.g. the `[H]` atoms in `C([H])([H])[H]`) into their neighbor's hydrogen count, matching how depiction tools treat explicit hydrogens, while keeping bracket hydrogen counts such as `[OH-]` and `[NH4+]` unchanged.
- Render bracket isotopes as a leading superscript mass number (e.g. `2H`, `13C`), keeping isotopic and charged hydrogens as drawn atoms.


saecki:
Thanks!

PR #5198, opened 2026-06-30
diffst:0.1.0

clvnkhr:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: this package gives the diff (backed by rust's crate `similar`) of two files, with user facing functions that format nicely into a table from which source can be easily copied out. Helpful if you need to share a list of changes with coauthors who do not use version control.

  - Explanation: diff + (typ)st. It is simple but I don't think it is canonical
    

packages/preview/diffst/0.1.0/lib.typ
```diff
@@ -0,0 +1,1017 @@
...
+  semantic-cleanup: true,
+  display: "collapsed",
+  collapse-threshold: 14,
+  context-lines: 3,
+  table-style: default-table-style,
+  table-layout: "split",
+  colors: (:),
+) = {
+  diffst-content(
+    read(old-path),
```
saecki, line 983: To read from a path outside of the package directory, so any path a user passes into this function, it must use the `path` type introduced in Typst `0.15`. See https://github.com/typst/packages/blob/main/docs/resources.md#paths-are-package-scoped

saecki (commented):
Thanks for this package!

packages/preview/diffst/0.1.0/lib.typ
clvnkhr: @saecki thanks for the comment - I had a look and I think I fixed this by passing the input into the path constructor, which accepts either strings or paths. I have tested this on my end, please confirm if this looks alright

clvnkhr: @saecki i see you added the waiting on author tag. I understand the issue to be solved so I will mark this resolved

clvnkhr: hey @saecki, sorry if I should not have resolved the thread, I'm a little new to this. I unresolved it. I think on reading the link you sent earlier again, I should not be expecting a string can be passed in, and I misread earlier expecting absolute path strings to also work. So I have pushed a new commit removing the string fallback entirely (passing through path constructor). Please let me know if there's anything else I need to do to get this moving, thanks

saecki: Hey, sorry I had a pending message that I meant to send, and I had a bit of time off. No worries about marking the comment as resolved.

> I think on reading the link you sent earlier again, I should not be expecting a string can be passed in, and I misread earlier expecting absolute path strings to also work. So I have pushed a new commit removing the string fallback entirely (passing through path constructor).

Yes, this looks correct now :)

PR #5200, opened 2026-06-30
typsium-iso-7010:0.1.1

Ants-Aare:
I am submitting
- [ ] a new package
- [x] an update for a package

Added new localised accessibility alt text and fixed wrong colours and inconsistent borders. Also some missing pictograms were added.


saecki:
Thanks

PR #5210, opened 2026-06-30
lemming:0.3.1

fGolke:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Actually append the end marker to math environments in the default show rule.


saecki:
Thanks

PR #5216, opened 2026-07-01
fix: formatting

F2011:


elegaanz:
This is intentional, `exclude` is a key in `typst.toml` manifests. Thank you still.

PR #5218, opened 2026-07-01
silky-letter-insa:0.2.3, silky-report-insa:0.5.3 and silky-slides-insa:0.2.0

SkytAsul:
I am submitting
- [ ] a new package
- [x] an update for a package

- Updated some assets
- Added multiple new options to the slides template
- Fixed issues


SkytAsul:
The 3 raised warnings will not be fixed since I want the link to point to the most up-to-date version of my fonts directory, which can change independently to the templates.

packages/preview/silky-letter-insa/0.2.3/README.md
```diff
@@ -0,0 +1,89 @@
...
+````
+
+## Fonts
+The graphic charter recommends the fonts **League Spartan** for headings and **Source Serif** for regular text. To have the best look, you should install those fonts.
+
+To behave correctly on computers lacking those specific fonts, this template will automatically fallback to similar ones:
+- Headings: [**League Spartan**](https://fonts.google.com/specimen/League+Spartan) -> **Arial** (approved by INSA's graphic charter, by default in Windows) -> **Liberation Sans** (by default in most Linux)
+- Body: **Source Serif** -> [**Source Serif 4**](https://fonts.google.com/specimen/Source+Serif+4) -> **Georgia** (approved by the graphic charter) -> _default Typst font_
+
+> You can download the fonts from [here](https://github.com/SkytAsul/INSA-Typst-Template/tree/main/fonts).
```
saecki: Could you use a link to a specific tag/release or a permalink to a commit instead. This will ensure that the linked resource always matches this version of the package.

Same for the other packages

saecki (commented):
Thanks for the updates!

packages/preview/silky-letter-insa/0.2.3/README.md
SkytAsul: Please see the comment above in this PR.

saecki: Sorry I didn't read that, my bad. Though this goes against the way we want packages to work. Their contents are supposed to be immutable and all resources that are linked should essentially be too. It's a bit unfortunate that there currently is no way to distribute fonts with the packages themselves, but we want to support that at some point in the future. Changing the linked fonts can change how documents using the package will be rendered and could be considered a change in the package. If you want to "update" the fonts, you can publish a new version of the package and update the links.
Another thing to consider is that if you changed the location of the font directory, for example because the repository is restructured, the instruction for using this version of the package would essentially break.

SkytAsul: I see. I fixed that

SkytAsul:
For some reason the PR got closed whereas I did not do it myself? I'll open another one.

saecki:
For future reference, I could've reopened this PR before you opened a new one, which is generally nice because the review history. It's no big deal though, just for next time :)

PR #5221, opened 2026-07-01
clean-ut:0.1.0

Lai-es:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a thesis template with the benefits of being able to do numbered subfigures, having a two-column bibliography and a caption function that allows to omit part of a figure description in the outline

  - Explanation:  "clean" reflects the template's minimal, uncluttered visual style, and "ut" stands for Universität Tübingen, where the thesis this template was built for was written. This is however a independent/unofficial template, not affiliated with the university itself.
    

packages/preview/clean-ut/0.1.0/typst.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "clean-ut"
+version = "0.1.0"
+entrypoint = "template/lib.typ"
```
elegaanz: `lib.typ` should be outside of the template directory. Most functions you define should also be part of the package and not the template: they should only define styling and apply it to the given content. If users really need to, they can chose to not use them and bring their own implementation instead, given how modular your package seems to be. If needed, you can also let them customize the defaults you provide further by letting them pass more arguments to your functions.

The template should mostly contain pure contents, and have sub-chapters `#include`-ed rather than imported as functions.

packages/preview/clean-ut/0.1.0/typst.toml
```diff
@@ -0,0 +1,18 @@
...
+authors = [
+	"Elias Riedling <https://github.com/Lai-es/>"
+]
+license = "MIT"
+description = "Thesis with two-column bibliography, subfigures, and custom captions"
+repository = "https://github.com/Lai-es/typst-thesis-template"
+keywords = ["Bachelorarbeit", "Masterarbeit", "aio", "thesis"]
+categories = ["thesis"]
+compiler = "0.14.2"
```
elegaanz: The `compiler` field is a minimum version, so unless you explicitly require this patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler.

elegaanz (commented):
Thank you for submitting this template. I have some remarks regarding its structure and advice to make it more idiomatic. If you need more guidance, feel free to ask me, or the community via the Forum or Discord.

packages/preview/clean-ut/0.1.0/typst.toml
Lai-es: thank you for noticing, i have updated the requirements

packages/preview/clean-ut/0.1.0/typst.toml
Lai-es: thank you, you are of course right. I changed the structure accordingly. The only thing troubling me is that `main.typ` compiled successful locally, but is a blank page. I couldn't pinpoint what the problem was, could you give me a hint maybe? Anyhow, i will look into this further in the next week

packages/preview/clean-ut/0.1.0/lib.typ
```diff
@@ -0,0 +1,201 @@
...
+      #context { strong(hydra(1, skip-starting: false, use-last: true)) }
+      #h(1fr)
+      #context { strong(counter(page).display()) }
+      #v(-0.9em) #line(length: 100%)
+    ],
+  )
+}
+
+// ===========================Template================================
+#let template = {
```
elegaanz, line 124: This should be a function that takes a `body` parameter and returns it after having configured the styles. Otherwise, `#show: template` will just replace the whole document with the value of this variable, which is just a bunch of styles and no content to apply it too.

PR #5226, opened 2026-07-01
dragonling:0.3.1

coljac:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Dragonling is a template for creating D&D 5E content (adventures, stat blocks, spell cards). This patch release (0.3.1) fixes the ability score modifier calculation in stat blocks: the modifier used `int()`, which truncates toward zero rather than flooring, so odd ability scores below 10 came out one too high (e.g. a score of 9 showed a modifier of 0 instead of −1). The modifier now uses `calc.floor`, matching the 5E rules for all scores. No other functional changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


saecki:
Thanks!

PR #5227, opened 2026-07-02
xwysyy:0.3.0

xwysyy:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: xwysyy helps create academic slides and notes with Typst and touying. This update adds a Universe template with thumbnail, expands built-in themes from 2 to 6, supports custom theme dictionaries, adds font and language parameters, adds xwysyy-doc dual output, adds handout and speaker-note support, splits internal modules, and sets the compiler to 0.14.2.

Checks:
- Fresh init passed locally with the submission copy: `typst init @preview/xwysyy:0.3.0 && typst compile main.typ`.
- Docker package-check passed locally: `ghcr.io/typst/package-check check @preview/xwysyy:0.3.0`.
- GitHub Lint and Test checks pass.


packages/preview/xwysyy/0.3.0/xwysyy.typ
```diff
@@ -0,0 +1,13 @@
+// xwysyy.typ — facade re-exporting every core submodule.
+// Users do `#import "xwysyy.typ": *` to pull in themes, slides, and note.
+// Optional cetz / fletcher / theorion integrations live in `xwysyy-extras.typ`.
```
elegaanz: As far as I can tell, this module is not actually exported, so it won't be possible for users to use these integrations. You probably want to expose it here.

packages/preview/xwysyy/0.3.0/typst.toml
```diff
@@ -0,0 +1,18 @@
...
+version = "0.3.0"
+entrypoint = "xwysyy.typ"
+authors = ["xwysyy"]
+license = "MIT"
+description = "Create academic slides and notes with touying."
+repository = "https://github.com/xwysyy/xwysyy-typst"
+keywords = ["slides", "presentation", "academic", "touying", "theme"]
+categories = ["presentation", "components"]
+disciplines = ["computer-science"]
+compiler = "0.14.2"
```
elegaanz: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.14.0` will allow using this package with a slightly outdated compiler. If you don't depend on even older features (and so do your dependencies) you can even lower that number.

packages/preview/xwysyy/0.3.0/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+- Universe template support: `typst init @preview/xwysyy:0.3.0` creates a ready-to-compile deck.
+- Six built-in themes: `sky`, `sunset`, `forest`, `midnight`, `violet`, and `graphite`.
+- Custom theme dictionaries can be passed directly to `theme`, so users can customize colors without forking the package.
+- Slide font parameters match note mode: `font`, `code-font`, and `lang`.
+- Touying handout mode, `#speaker-note`, and pdfpc export are documented and covered by examples.
+- `xwysyy-doc` compiles one source as a 16:9 deck by default and as A4 notes with `--input mode=note`.
+- CI scripts cover example compilation, visual regression, theme contrast, and README preview generation.
+
+## Preview
```
elegaanz, line 26: This section doesn't really make sense to display on Typst Universe, as these files are not available. I would either remove it (and keep it in the copy of the README in your own repo if you want to), or pre-compile them in your repository so that you can link to it directly.

elegaanz (commented):
Thank you for the update. I have a few minor comments that needs to be addressed before we can publish it.

packages/preview/xwysyy/0.3.0/xwysyy.typ
xwysyy: Thanks for catching this. I updated the main entrypoint to re-export `src/extras.typ`.

packages/preview/xwysyy/0.3.0/typst.toml
xwysyy: Thanks, that makes sense. I changed the package `compiler` field from `0.14.2` to `0.14.0` and updated the README badge and requirements text to match the lower minimum version.

packages/preview/xwysyy/0.3.0/README.md
xwysyy: I cleaned up the package README so it no longer assumes the repository-local `examples/` or `scripts/` directories are available on Typst Universe.

elegaanz:
Thank you!

PR #5228, opened 2026-07-02
larrow:1.2.0

Mambouna:
I am submitting
- [x] an update for a package

Description: Adds arbitrary point inclusion for `bend` parameter, arrow captions and passing elements directly as `from` or `to`.


saecki:
Thanks!

PR #5230, opened 2026-07-02
exercise-bank:0.5.2

nathan-ed:
I am submitting
- [x] an update for a package

#### Added
- **Optional exercise marker** — `optional: true` on `exo`/`exo-define`/`exo-show` shows a built-in star icon before the label; customize with `exo-setup(optional-symbol: ...)` or set to `none` to disable
- **Correction-given marker** — `corr-given: true` shows a dumbbell icon indicating the printed correction will be distributed to students; customize with `exo-setup(corr-given-symbol: ...)`
- **`margin` badge style** — side-label layout with a fixed-width left column and a horizontal rule; the 10th badge style
- **Per-display overrides in `exo-show`** — `optional`, `optional-symbol`, `corr-given`, `corr-given-symbol` can override bank metadata for a specific display call


saecki:
Thanks!

PR #5231, opened 2026-07-02
simple-plot:0.9.1

nathan-ed:
I am submitting
- [x] an update for a package
#### Added
- **Automatic function-label placement** — labels now find a clear spot along the curve automatically, avoiding overlaps with other curves, tick labels, axes, and previously placed labels; `label-anchor`/`label-side` override when explicit placement is needed
- **`min-tick-spacing`** — auto-widens tick step to the next "nice" value (1, 2, 5, 10…) when the axis range would crowd labels below the threshold; keeps large-range plots legible without manual `xtick-step`
- **`hide-crossed-tick-labels`** — hides individual tick labels that a plotted curve crosses, so digits stay readable even when curves pass through the label band
- **White label backgrounds** — tick labels and axis-name labels are drawn over a white background so curves behind them don't bleed through

#### Fixed
- `ylabel-offset` default corrected: was `(-0.08, -0.05)`, now `(0.08, -0.05)` — the $y$ axis label was shifted left instead of right


elegaanz:
Thank you!

PR #5232, opened 2026-07-02
taskize:0.2.6

nathan-ed:
I am submitting
- [x] an update for a package

#### Added
- **True baseline alignment** — with `label-baseline: "center"` (default), inline content (text, inline math, symbols) now shares the same paragraph line as the label via hanging-indent layout, so fractions and tall inline elements align their baselines with the label instead of sitting above it

#### Fixed
- `compiler` minimum lowered from `0.14.2` to `0.14.0` — patch releases add no new language features, so users on `0.14.0`/`0.14.1` were incorrectly excluded


saecki:
Thanks!

PR #5234, opened 2026-07-02
iridis:0.2.0

Robotechnic:
I am submitting
- [ ] a new package
- [x] an update for a package

This PR makes iridis compatible with typst 0.15.0


packages/preview/iridis/0.2.0/typst.toml
saecki: The `compiler` field is a minimum version. Does this update include changes that break the package on `0.14`? Otherwise this can just be left at `0.14`

saecki (commented):
Thanks for the update

packages/preview/iridis/0.2.0/typst.toml
emstruong: Just tagging @Robotechnic , in case they didn't see this. I really like this package :)

Robotechnic: Yes, I missed it, thanks. 
I only tried this on 0.15, this is why the minimum is 0.15. I can try 0.14 to see if it breaks

Robotechnic: Ok it works with 0.14

PR #5236, opened 2026-07-02
obelisk:0.2.0

atzlt:
I am submitting
- [ ] a new package
- [x] an update for a package

Added a toggle for line numbers; optimized inline math equation layout; implemented breakable theorem blocks with automatic stroke side.


saecki:
Thanks!

PR #5237, opened 2026-07-02
callisto:0.3.0

sijow:
I am submitting
- [ ] a new package
- [x] an update for a package

Update with support for code block execution, cell attachments, ANSI escape sequences and more.


sijow:
The remaining warnings are expected (three functions with capitalized names)

(Also in case it's confusing: I updated the `typst.toml` to use my new GitHub handle, it's not  "takeover" :)  You can see that https://github.com/knuesel/callisto redirects to https://github.com/sijow/callisto )

sijow:
Since cmarker 0.1.10 was just merged I just pushed a change to update the dependency in Callisto

elegaanz:
Thank you!

PR #5253, opened 2026-07-04
baler:0.1.0

yana-zhabura:
I am submitting
- [x] a new package
- [ ] an update for a package

Create tables and grids that automatically balance column widths in a way that reduces grid height. For every column, a minimum and maximum width may be set. 

- [ x ] selected [a name](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) that isn't the most obvious or canonical name for what the package does
  - Explanation:
    
A baler is a piece of farm machinery used to compress hay into compact bales, just like balancing table columns allows to fit it into a smaller space.
- [ x ] added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys
- [ x ] added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package
- [ x ] have chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) and added a `LICENSE` file or linked one in my `README.md`
- [ x ] tested my package locally on my system and it worked
- [ x ] [`exclude`d](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude) PDFs or README images, if any, but not the LICENSE


packages/preview/baler/0.1.0/README.md
```diff
@@ -0,0 +1,43 @@
...
+#let grid-content = (lorem(40), lorem(10), lorem(3), lorem(12), lorem(30), lorem(5))
+#let light-green = rgb(230, 255, 230)
+#let light-yellow = rgb(255, 255, 220)
+#let light-red = rgb(255, 230, 230)
+#set grid(stroke:1pt)
+```
+### Example 1: Table with `columns` Value
+In this example, only the number of columns is set.
+
+![Example 1](example1.svg)
```
elegaanz: More descriptive alternative texts would be nice. They don't have to be very long, but should convey the message behind the image to people who can't properly see it (because of a visual impairment for example, or because their Internet connection is too slow to load the image). For example here it could say: "The same table laid out by Typst and by baler. The former takes more vertical space than the later, which allocates larger columns when there is more content for better space distribution."

packages/preview/baler/0.1.0/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Example 2: Grid with `columns` Array
+
+In this example, a `columns` array is set. In the case of *Balanced Grid with Size Limits*, the first column's size is set to exactly `70pt`, and the third column's size it set to an automatically determined value between `30pt` and `100pt`, which overrides the default minimum of `50pt` for `auto`. 
+
+![Example 1](example2.svg)
```
elegaanz: Same here.

elegaanz (commented):
Thank you for this package, it looks very promising! I just have a small remark, once it is addressed, it should be good to merge.

packages/preview/baler/0.1.0/README.md
yana-zhabura: Both fixed now, thank you :)

elegaanz:
Great, thank you!

PR #5254, opened 2026-07-04
aero-dhbw:0.2.1

cocoveal:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: `aero-dhbw` is a template for writing theses/papers at DHBW Ravensburg (bilingual German/English, single or multiple authors).

This update (0.2.0 → 0.2.1) contains:
- List of Figures / List of Tables now render only when the document actually contains figures / tables, instead of emitting an empty heading.
- Removed a hardcoded text `region` so the document language alone drives hyphenation and quote style; users can set a specific region themselves via `#set text(region: ...)`.
- Added a clear assertion for `start-date` / `end-date` instead of failing with a cryptic error when they are omitted.
- Internal cleanups: de-duplicated an outline-style branch and fixed the acronym-list table column inset so the long-form column aligns to the right margin.


saecki:
Thanks!

PR #5262, opened 2026-07-05
presio:0.2.2

benedict-armstrong:
Bugfix release of [presio](https://github.com/benedict-armstrong/presio-typst-package).

- Fixed `unknown image format` error when passing a local video file (mp4/webm) to `media()`: the default `auto` placeholder used `image(source)` for every local path, which Typst can't decode for video. The auto-preview is now gated on image extensions; videos fall back to the dark placeholder block (an explicit `placeholder` can supply a poster frame). Reported by a user of the package.

I confirm that I am the author of the previous versions (0.1.0, 0.2.0, 0.2.1).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/preview/presio/0.2.2/LICENSE
+++ b/packages/preview/presio/0.2.2/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Benedict Armstrong
+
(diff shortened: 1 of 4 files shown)


saecki:
Thanks

PR #5271, opened 2026-07-06
simple-plot:1.0.0

nathan-ed:
- [x] an update for a package

### [1.0.0] - 2026-07-06

First stable release. The API documented in the manual is now considered stable.

#### Added
- **`width` / `height` accept lengths** — `plot`, `zoom`, `volume-of-revolution` (and the wrappers) now take real lengths (`width: 6cm`, `height: 30mm`) in addition to plain numbers, which keep their historical meaning of cm
- **Named `fn:` in series helpers** — `fill-area`, `riemann-sum`, `func-plot` accept `fn:` (and `area-between` accepts `fn1:` / `fn2:`) in addition to the positional function, matching the `(fn: ..., ...)` series dictionaries
- **`axis-x-pos: none` / `axis-y-pos: none` hide an axis** — the axis line, arrow, ticks and labels are skipped entirely (used by `limit-schema`, which previously drew an unintended y-axis)
- **Unknown named arguments to `plot()` now raise an error** — typos like `xtick-lables:` were previously silently ignored
- **Unified `hatch` argument** — `hatch` also accepts a dict `(style: "ne", spacing: 10pt, stroke: red + 1pt)` in `fill-area`, `area-between`, `riemann-sum` and `fill-closed`; the flat `hatch-spacing` / `hatch-stroke` arguments keep working as fallbacks
- **`volume-of-revolution` cleanup** — `radius-marker-x` is the new canonical name for the radius-marker position (`yaxis-x` and `show-yaxis` remain accepted as deprecated aliases); unknown named arguments now raise an error
- **`set-plot-defaults` validates option names** — a typo like `set-plot-defaults(widht: 10)` now raises an error instead of being silently stored and ignored
- **Zoom insets re-render every series type** — `fill-area`, `area-between`, `fill-closed`, `riemann-sum` rectangles, `vline`/`hline` and parametric curves now appear inside `zoom()` insets (previously only function curves and point series did), correctly clipped to both the rectangular and the circular lens

#### Fixed
- **Riemann rectangles cut by the plot window no longer draw false edges** — a bar extending past `xmax` (e.g. decimal `xmax: 2.5` with `domain: (0, 3)`) or above `ymax` was clamped and then stroked on all four sides, showing a wrong width and a flat top at the window boundary; clipped edges are now left open so the bar reads as continuing beyond the window, and bars fully outside are dropped
- **`show-xi` labels no longer collide with numeric tick labels** — every tick label *near* an `x_i` label (measured in cm, not only exact position matches) is hidden, and `x_0` takes the origin's `0` spot instead of being shifted into `x_1`
- **`show-dx` + `show-xi` keep all labels** — the `Δx` bracket now drops below the `x_i` row (dimension-line style) instead of silently skipping the two straddling `x_i` labels
- **`show-points` no longer sprays a label by default** — `point-label` defaults to `none` (dots only); when a label is requested, arrows point to the 3 nearest dots instead of criss-crossing the plot
- **Label backgrounds no longer punch white holes into area fills** — tick labels and axis-name labels suppress their white background when they sit on a Riemann bar or an area fill (the digits print directly on the fill)
- **`compiler` requirement corrected to `0.13.0`** — the hatch patterns use `tiling()`, which does not exist before Typst 0.13; the manifest previously claimed 0.11.0
- **Raw `(annotation: ...)` dicts default to 9pt** — aligned with the `note()` helper, which already used 9pt
- **Axis arrow extensions are no longer scale-dependent** — `axis-x-extend` / `axis-y-extend` default to absolute lengths (`0.3cm`), so a small data range (e.g. y from 0 to 0.5) no longer produces a huge axis overshoot; lengths are accepted for explicit values, bare numbers keep the legacy data-unit meaning
- **Thick hatch strokes no longer show holes** — hatch lines are extended past the tile boundary and repeated at the opposite corners, so `hatch-stroke` thicknesses comparable to `hatch-spacing` tile seamlessly
- `plot-fn` now forwards `samples` to the drawn curve (it was only used for the auto y-range)
- `zoom(box-fill: none)` now really makes the inset background transparent (the value was previously dropped and fell back to white)


elegaanz:
Thanks!

PR #5274, opened 2026-07-06
patstdlib:0.3.2

patrick-kidger:
I am submitting
- [ ] a new package
- [x] an update for a package

Extended `#show: appendix` with the ability to customise numbering and supplement.

### DIFF
--- a/packages/preview/patstdlib/0.3.2/README.md
+++ b/packages/preview/patstdlib/0.3.2/README.md
@@ -1,0 +1,54 @@
+<h1 align='center'>patstdlib</h1>
+
+Some standard pieces I find useful: fonts, subfigures, algorithms, labeled enums, etc.
+
+## Documentation
+
+Please see the inline documentation attached to each object. A full list of all available functionality can be found in `src/lib.typ`.
+
+## Functionality
+
+Here's a few highlights:
+
+**Labeled enums**
+
+```typst
+#show: enable-referable-enums
+#set enum(full: true)
+
+#referable-enum("step")[
+  + Foo
+  + Bar
+  + Baz <step:baz>
+]
+
+In @step:baz then ...
+// Renders as 'In step 3 then ...'
+```
+
+**Subfigures**
+
+```typst
+#show: enable-referable-subfigures
+
+#figure(
+  caption: [Main caption],
+  grid(
+    columns: 2,
+    subfigure(
+      caption: [Caption for subfigure '(a)'],
+      image(...)
+    ),
+    subfigure(
+      caption: [Caption for subfigure '(b)'],
+      image(...)
+    ),
+  )
+)
+```
+
+**Significant figures**
+
+```typst
+#sigfig(3.141592653, digits: 3)  // 3.14 as a `decimal`
+```
--- a/packages/preview/patstdlib/0.3.2/LICENSE
+++ b/packages/preview/patstdlib/0.3.2/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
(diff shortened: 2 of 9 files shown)


packages/preview/patstdlib/0.3.2/README.md
saecki: Links to external repositories should either be perma-links or links to a specific branch/tag. Here though, I would just link to the local file:
```suggestion
Please see the inline documentation attached to each object. A full list of all available functionality can be found in [`src/lib.typ`](src/lib.typ).
```

saecki (commented):
Thanks for the update!

packages/preview/patstdlib/0.3.2/README.md
patrick-kidger: In general that approach will only when viewing the README via GitHub. The full link is therefore intentional to avoid a broken link when accessed via any other means.

saecki: The local link will work on Typst Universe, GitHub, and most other places where you can host a git repository. But if you want to use an external link, that's also fine. Though, it should be more permanent than the current link to the `main` branch of the repository. The rationale is that all linked resources of the  published package should match the exact version, so the documentation will continue working as expected, when new versions are released.

(Sorry for the slow reply)

patrick-kidger: In the interests of maintainability, I have now just removed the URL from the README altogether.

(I question whether this is really so serious an issue as to impose a 1-week delay on releasing a package?)

saecki: > In the interests of maintainability, I have now just removed the URL from the README altogether.

As mentioned before, linking to the local file would not require any extra maintenance effort, and should work for most places this README would be hosted.

> (I question whether this is really so serious an issue as to impose a 1-week delay on releasing a package?)

We care about the quality of packages on Typst Universe and try to make sure there is little friction for users. Copying an example that doesn't work, because the upstream version broke some API, is a very frustrating and for new users maybe the last experience they will have with a package. We know that this results in more work for package authors/maintainers and also want to improve some of the pain points by building better tooling for package development. But we still think that being somewhat pedantic about things that improve the user experience is a tradeoff we're currently willing to make. That's why we built the `package-check` in the first place.
Package review is just one of my responsibilities at Typst, so I'm not solely focused on this, and sometimes there are just a lot of package submissions, and other things to do. I've apologized already for the slow reply, it's not like your package was exclusive held back. In addition to that, I've had some time off last week.

patrick-kidger: Okay, I understand your rationale. I would challenge that this approach minimises user frustration, however: the delay in package releases is frustrating for our (company-internal) users. I would also challenge that the current review improves matters: I have previously released packages with (accidentally) broken core functionality, and this hasn't been caught by review.

To be clear, I'm a huge fan of Typst. So much so that when I founded my company, we made the decision to use Typst ubiquitously for all documents internally, with this one tool replacing all of Word/Excel/PowerPoint/LaTeX, have automated wiring it up to DocuSign, etc. [The joys of running a geeky company!] But package distribution is currently a genuine pain point.

saecki: > Okay, I understand your rationale. I would challenge that this approach minimises user frustration, however: the delay in package releases is frustrating for our (company-internal) users.

I totally understand, and we'll try to reduce this delay in the future by building better developer, publishing, and review tooling.

> I would also challenge that the current review improves matters: I have previously released packages with (accidentally) broken core functionality, and this hasn't been caught by review.
 
This is also an area that we'd like to improve, and that we've discussed internally. Reviewing the core logic of packages is infeasible, and not something we strive to do. But we want to provide a standardized way of specifying tests for packages that can be checked upon publishing, and which will also be used to detect breaking changes in future Typst versions.

> To be clear, I'm a huge fan of Typst. So much so that when I founded my company, we made the decision to use Typst ubiquitously for all documents internally, with this one tool replacing all of Word/Excel/PowerPoint/LaTeX, have automated wiring it up to DocuSign, etc. [The joys of running a geeky company!] But package distribution is currently a genuine pain point.

Thank you for the kind words :)

For company internal use cases, private registries and importing packages from git repositories will hopefully remove some of this pain. Though there are unfortunately no concrete plans yet :/

patrick-kidger: You're welcome!

> importing packages from git repositories

That'd be awesome! Ideally this would be something that also supports monorepos, btw. I'm imagining syntax roughly of the sort `#let foo = remote-package("github.com/...", tag: "v2", subdir: "foo/bar")`. Food for thought! :-)

PR #5275, opened 2026-07-06
curriculo-acad:0.1.2

philkleer:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This update includes a new implemented area of *pós-doutorado* in the education area and also handles the different types now in a clearer way. Furthermore, package was successfully tested under typst 0.15.0.


elegaanz:
Thank you.

PR #5278, opened 2026-07-06
tschich:0.2.0

switchlex:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added functionality to display an estimate of the current optimal font size.


saecki:
Thanks!

PR #5279, opened 2026-07-06
theoretic:0.4.0

nleanba:
I am submitting
- [ ] a new package
- [x] an update for a package

This update adds parameters `numbering` and `numbering-depth` to customize the theorem numbering. In particular, the latter parameter customizes how many levels of the headings are inherited. This was requested in https://github.com/nleanba/typst-theoretic/issues/9 (and https://github.com/nleanba/typst-theoretic/pull/11).

Also, this adds a `#theoretic.configure()` function which allows setting these for all subsequent theorems, because these seem like things that would be set for all theorem-environments. I may add more options to `configure` in the future.

**Full Changelog**: https://github.com/nleanba/typst-theoretic/compare/v0.3.1...v0.4.0

### DIFF
--- a/packages/preview/theoretic/0.4.0/README.md
+++ b/packages/preview/theoretic/0.4.0/README.md
@@ -1,0 +1,69 @@
+# Theoretic
+
+> Opinionated tool to typeset theorems, lemmas and such
+
+Example Usage:
+```typ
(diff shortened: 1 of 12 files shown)


elegaanz:
Thank you!

PR #5280, opened 2026-07-06
ctz-euclide:0.2.0

nathan-ed:
I am submitting
- [x] an update for a package

#### Added

- **Automatic label placement (collision avoidance), on by default**: labels
  are placed as close as possible to the requested anchor without overlapping
  drawn segments, circles, point markers or other labels. Labels without an
  explicit position are fully automatic (`labels: true`). Guarantees: a label
  is always closer to its own point than to any other point by a clear margin
  (never reads as the neighbor's label); the font is shrunk (85%, then 70%)
  only when *no* full-size position exists — and then all labels of the
  figure share the same scale. Global switch `ctz-label-avoid(false)`,
  per-label `(pos: "above", avoid: false)`, tunable via
  `ctz-label-avoid((rings: ..., shrink: ..., assoc-factor: ...))`. Angle
  labels slide along the bisector to stay clear.
- **Exterior angle support** (#2): `ctz-draw-angle(..., angle-type: "reflex")`
  (explementary) and `angle-type: "supplementary"` (exterior).
- `ctz-def-golden(..., inside: false)` for the external golden section.
- `ctz-def-translate` accepts a pair of point names as the vector.
- `ctz-draw(points: ..., shape: ..., size: ..., color: ...)` style overrides.
- `ctz-style(angle: (...))` is now honored by `ctz-draw-angle`.
- Greek point names render as symbols in labels ("alpha" → α, "Omega1" → Ω1).

#### Fixed

- **Arc rendering** (#4): arcs, sectors, semicircles and compass marks were
  drawn offset by one radius (cetz anchors arcs at their start point, not
  their center) and with mirrored start/end angles.
- **Systemic `calc.atan2` argument-order bug** affecting slope calculations,
  tangent points from an external point, sector/semicircle angles, the
  german-style right-angle mark, sloped segment labels, ellipse
  projections/tangents and automatic label placement.
- **cetz version conflict** (#1): upgraded to CeTZ 0.5.2 — no more conflict
  when combining ctz-euclide with cetz-plot or your own cetz import, in any
  import order.
- `ctz-def-golden` crashed on every call.
- `ctz-draw(name, sketchy: true)` crashed on named lines and polygons.
- `ctz-def-excenter`/`ctz-def-excircle` vertex is now case-insensitive (the
  default silently computed the excenter opposite C).
- Feuerbach point returned the incenter instead of the tangency point.
- Tangent points from an external point used asin instead of acos.
- Duplicate labels for the same point (same content) are now drawn only once.
- Labels for point names longer than 5 characters were silently dropped.
- Grids and axes no longer drop their last line/tick to float rounding.
- Internal module names (`util`, `draw`, `circle`, ...) are no longer leaked
  into your document by the wildcard import (the `cetz` binding is still
  exported on purpose, for `import cetz.draw: *`).
- `ctz-version` is now read from typst.toml (was stuck at 0.1.0).

### 0.1.5

- Main-levée (hand-drawn/sketchy) style, including `sketchy: true` on `ctz-draw`
- Global style configuration variables for colors, point shapes, sizes, and mark widths
- Canvas-level clipping via `ctz-canvas(clip-canvas: ...)` with line/segment clipping
- Labels now render in math font by default
- Expanded docs and gallery examples for styling, main-levée, and clipping


saecki:
Thanks!

PR #5281, opened 2026-07-06
thwildau-telematics:0.1.1

C4mperpr0:
I am submitting
- [ ] a new package
- [x] an update for a package

**0.1.1**
- glossary entries, abbreviations and units can now be define multiple ways (see docs)
- bibliographic desciprtion: enhanced layout and bug fix (title language was always english)
- change title and chapter spacings


elegaanz:
Thank you for this update. Can you fix the warnings reported in CI please? The one about `hydra` can be ignored, it is more informative than an actual issue to be fixed.

C4mperpr0:
Thanks you for the reminder, I somehow forgot to check the pipeline :melting_face:

elegaanz:
Thank you!

PR #5282, opened 2026-07-06
haita:0.2.1

wensimehrp:
I am submitting
- [x] a new package
- [ ] an update for a package

Haita is a pure Typst documentation tool. See <https://wensimehrp.github.io/haita> for a working demo.

  - Explanation:
    
### DIFF
--- a/packages/preview/haita/0.2.1/README.md
+++ b/packages/preview/haita/0.2.1/README.md
@@ -1,0 +1,100 @@
+# Haita
+
+Here's Haita Docs. A simple tool that
+has a single requirement: [Typst](https://github.com/typst/typst). Here are some features:
+
+- Pure Typst workflow
+- Features inherited from Typst:
+  - Simple yet expressive Typst syntax helping you focussing on your content
+  - Native syntax highlighting
+  - Native MathML output
+  - Fast compliation
+  - Native support for `watch` and `serve`
+  - PDF and HTML generation from the same source (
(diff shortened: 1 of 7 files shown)


packages/preview/haita/0.2.0/README.md
saecki: Could you please add the missing alternative descriptions?

saecki: This file is missing. Make sure to commit all linked files and possibly add them to the `exclude` list.

saecki: This import is still using `0.1.0`

packages/preview/haita/0.2.0/typst.toml
saecki, line 13: The compiler field is a minimum version. If this package also works with `0.14`, could you lower it so older Typst versions aren't artificially blocked from using it. Otherwise, you can just leave it as is.

saecki (commented):
Thanks for the update

packages/preview/haita/0.2.1/README.md
wensimehrp: I am not very sure about this warning. This file here is only a demo, and I want the user to stick with the latest version. Perhaps it is better to keep it as-is?

saecki: The rationale is that the instructions of this published package keep working, even when you publish a new version of the package with breaking changes.

PR #5276, opened 2026-07-06
silky-letter-insa:0.2.3, silky-report-insa:0.5.3 and silky-slides-insa:0.2.0

SkytAsul:
I am submitting
- [ ] a new package
- [x] an update for a package

- Updated some assets
- Added multiple new options to the slides template
- Fixed issues

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use.

PR #5288, opened 2026-07-07
CI: sparse-checkout all touched packages

elegaanz:
The script would previously only load the first 30 files and ignore some packages in case there were more than one and the firsts accounted for more than 30 files.


saecki (approved):

saecki:
This seems to be the culprit, which caused the test CI to silently omit one of the submitted packages here: https://github.com/typst/packages/pull/5276

ndg6:
Interesting! My package (*staunton*) would definitely also have been hit by a 30 files limit (it has 50 files).

PR #5292, opened 2026-07-07
silky-slides-insa:0.2.0

SkytAsul:
Follow-up from https://github.com/typst/packages/pull/5287. Updated thumbnail to be 1080 in width.

I am submitting
- [ ] a new package
- [x] an update for a package

- Updated some assets
- Added multiple new options to the slides template
- Fixed issues


elegaanz:
Thank you for fixing it.

PR #5300, opened 2026-07-08
minerva-thesis:0.2.3

lvandevelde:
I am submitting
- [ ] a new package
- [x] an update for a package

Minor update: some bug fixes solving issues raised by users.


elegaanz:
Thank you.

PR #5307, opened 2026-07-09
jurlstify:0.3.0

SchrodingerBlume:
I am submitting

- [ ] a new package
- [x] an update for a package

Description:

feat: bump version to 0.3.0, **rename** API onto break-placement / hyphen-display axes (totally breaking change!):

- `hyphens`->`break-at-literal-hyphens`;
- `every`->`extra-break-every`;
- `breaks`/`big-breaks`/`no-breaks`->`break-chars`/`big-break-chars`/`no-break-chars`;
- split `show-hyphen` into `show-hyphens-after-delimiters` & `show-hyphens-at-extra-breaks`;
- `spaces`->`show-spaces-as`.

Add manual & committed PDF; slim bilingual README linking to it.

### DIFF
--- a/packages/preview/jurlstify/0.3.0/LICENSE
+++ b/packages/preview/jurlstify/0.3.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 SchrodingerBlume
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


SchrodingerBlume:
P.S.

I chose to make a breaking rename because I felt the original API name wasn't intuitive enough.

After the rename, it's much clearer.

I've already mentioned in the new README and manual that if anyone wants to use the old name, they can stay on version 0.2.x.

The fact that Typst itself supports version selection is what encouraged me to make this breaking change.

elegaanz:
Yes, that's what versioning is for :) Thanks for the update.

PR #5308, opened 2026-07-09
gbt9704-gongwen:0.1.0

songwupei:
I am submitting
  - [x] a new package
  - [ ] an update for a package

  Description: GB/T 9704-2012 党政机关公文格式 Typst 模板。

  A Typst template for Chinese government official document format (GB/T 9704-2012). Features
  red-header, GB-standard A4 page margins, 28pt line spacing, FangSong body text, and all standard
  document elements (attachments, signatures, dates, notes, 版记). CJK font fallback chain with
  open-source alternatives.

    - Explanation: "gbt9704" is the specific Chinese national standard number (GB/T 9704-2012). It
  is not a generic descriptive term like "government-document" or "official-letter". The name
  identifies the standard the template implements, making it a precise but non-canonical choice.
  - [x] ensured that my package is licensed such that users can use and distribute the contents of
  its template directory without restriction, after modifying them through normal use.


packages/preview/gbt9704/0.1.0/typst.toml
elegaanz: Because this is not an official package from the government, I will ask you to change the name [to conform to our naming guidelines](https://github.com/typst/packages/blob/0488b38bb7fdfbce81981803ebbcf5612f8919f8/docs/manifest.md#naming-rules). In short, it should be clear that this is only a possible implementation and not the canonical one just by reading the name. Adding a distinctive word to the current name can be a good strategy to achieve that. Even if "unofficial" technically fits the rule, I would prefer if you could refrain from using it, both because it is not that distinctive, already too common in package names but also because it could be confusing since this template is designed for formal and official documents.

elegaanz, line 17: Can you regenerate the thumbnail with a non-transparent background please? Otherwise, it won't be legible for people using dark theme.

elegaanz (commented):
Thank you for this template :)

songwupei:
Thanks for the review!

**Name change:** Renamed from `gbt9704` to `gbt9704-gongwen`. "Gongwen" (公文) is the pinyin for "official document" in Chinese — it is distinctive and clearly indicates this is a community implementation rather than an official government package.

**Thumbnail:** Regenerated with a solid white background (`set page(fill: white)`), so it should now be legible in both light and dark themes.

Please let me know if any further changes are needed. Thanks!

elegaanz:
Thanks!

PR #5330, opened 2026-07-13
taskize:0.2.7

nathan-ed:
I am submitting
- [x] an update for a package

Description: **taskize** adds auto-fit column selection: `columns: "auto-fit"` picks the largest column count that does not introduce a new item wrap or fixed-size content overflow, bounded by `max-columns`. Two modes are available: `auto-fit-mode: "fill"` (span items use their full rendered width) and `auto-fit-mode: "uniform"` (every item must fit one ordinary column). Useful for exercise sheets mixing short choices, longer text, math, spans, and small figures.

##### Added
- **Auto-fit columns** — `columns: "auto-fit"` selects the largest column count, bounded by `max-columns`, that does not introduce a new item wrap or fixed-size content overflow. `auto-fit-mode: "fill"` lets spans use their full rendered width, while `auto-fit-mode: "uniform"` requires every item to fit one ordinary column.


saecki:
Thanks!

PR #5333, opened 2026-07-13
sertyp:0.1.5

Uhrendoktor:
I am submitting
- [ ] a new package
- [x] an update for a package

- Sertyp now supports **arbitrary depth** deserialization:
The engines architecture was changed from recursive -> linear to avoid running into max function call depth issues.
- Minor README fixes

PR #5335, opened 2026-07-13
kalt:0.1.0

Uhrendoktor:
I am submitting
- [x] a new package
- [ ] an update for a package

Kalt evaluates arbitrary nested equations and removes the need to
transcribe your equations into the default `calc` system. You will no longer run
into unsupported operations, and it all works for scalars and matrices. Since
everything runs on normal content, you can store and adjust the results as
needed.

It is similar to `eqalc` but way more powerful.

It supports:
scalars including complex numbers,matrices, vectors and a lot of mathematical operations.
I can also do matrix element wise mapping, merging, etc.

Example:
```typst
#import "@preview/kalt:0.1.0": comp

$comp((vec(4 dot 3/sqrt(2), 22.3i) dot mat(3, 5))[::1,[1,1,-2]])$
```
image

For more information see README.

  - Explanation:
    
   `kalt` contains kal ~ cal for calculation. the `t` stands for typst (which is slightly against the naming convention, but very helpful, since the library has a rust backend package, that needs to be published on crates.io). Also, it is the german for cold, which is neat.
    

Uhrendoktor:
Fails check because:
- Requires merging #5333 first.

elegaanz:
Thank you!

PR #5336, opened 2026-07-13
quick-vertex:0.1.0

13Stokes31:
Adds **feasible-region 0.1.0**.

Plots feasible regions of two-variable linear programs with CeTZ: bounded and unbounded regions, unique/multiple optima (segment or ray), unbounded objective, empty regions, strict inequalities and any quadrant. Objective level lines, gradient, and a vertex table with the optimum highlighted. Labels are localizable (`lang`, English default, Spanish included).

Repository: https://github.com/13Stokes31/feasible-region

- [x] I read the submission guidelines.
- [x] The package is licensed under a valid SPDX license (MIT), and a `LICENSE` file is included.
- [x] `typst.toml` is complete (name, version, entrypoint, authors, license, compiler).
- [x] The entrypoint compiles with the declared compiler version and only depends on a published package (`@preview/cetz:0.5.2`).


packages/preview/quick-vertex/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "feasible-region"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Jesús Manuel González Meneses"]
+license = "MIT"
+description = "Feasible region plots for two-variable linear programming: bounded/unbounded regions, unique/multiple optima, level lines and gradient, with CeTZ."
+repository = "https://github.com/13Stokes31/feasible-region"
+keywords = ["linear-programming", "optimization", "plot", "math", "education"]
+categories = ["visualization"]
+compiler = "0.14.0"
```
saecki, line 11: Could you `exclude` the `gallery` directory here?
The images will still be available for display on Typst Universe, but won't be downloaded when a user imports the package.

packages/preview/feasible-region/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "feasible-region"
```
saecki: This name is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

saecki (commented):
Thanks for this package!

13Stokes31:
Thanks for the feedback! Renamed the package to `quick-vertex` and excluded the `gallery` directory from the package download.

saecki:
Thanks!

PR #5337, opened 2026-07-13
breather:0.1.0

nathan-ed:
I am submitting
- [x] a new package

Description: One show rule that fixes lines overlapping because of tall inline math (display-style fractions, matrices, big operators). It measures each inline equation with bounds text edges and expands only the lines that actually need it, so regular lines keep perfectly even leading.

- [x] added a typst.toml file with all required keys
- [x] added a README.md with documentation for my package
- [x] have chosen a license and added a LICENSE file or linked one in my README.md


elegaanz:
Thank you!

PR #5339, opened 2026-07-13
aero-dhbw:0.4.0

cocoveal:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Updates aero-dhbw to 0.4.0. Project-relative abstract and annex content is now evaluated by the calling document, and users can optionally provide a custom declaration of authorship while retaining the generated declaration as the default.


elegaanz:
Thank you.

PR #5342, opened 2026-07-13
stair-division:0.1.0

13Stokes31:
Adds **ruffini** 0.1.0 — a small, dependency-free package for typesetting
**synthetic division (Ruffini's rule)**: the classic three-row division box and
the stacked factorization staircase.

Unlike a sign/variation table, it actually **computes** — you pass the
coefficients and the root and it does the (exact, rational) arithmetic and draws
it, with a boxed remainder, exact fractions, a `variable` option, `en`/`es`
localization, and an optional teaching overlay (`trail`) showing the
bring-down / ×root / column-sum flow.

- Repository: https://github.com/13Stokes31/ruffini
- License: MIT
- No dependencies (native `table`), Typst `>= 0.14.0`.


packages/preview/ruffini/0.1.0/typst.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "ruffini"
```
saecki: This name is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

saecki (commented):
Thanks for this package!

13Stokes31:
Thanks for the feedback! Renamed the package to `stair-division`.

saecki:
Thanks!

PR #5360, opened 2026-07-16
vlna:0.2.0

iamanro:
Update of my own package **vlna** from 0.1.1 to **0.2.0**.

> Note for review: the previously published versions list the author as
> `ShinoYumi`, which is my former GitHub username — the account was renamed to
> `iamanro` (so `github.com/ShinoYumi/typst-vlna` now redirects to
> `github.com/iamanro/typst-vlna`). Same author, same package.

### What's new in 0.2.0
- **Jump-to-source preserved.** The old rule rebuilt text with
  `it.text.replace(" ", "~")`, which moved the source position of the affected
  glyphs onto the package. 0.2.0 wraps the **original** matched content in
  `#box(..)` instead, so clicking a glyph in the PDF still jumps to the user's
  document.
- Unified short-word + title runs → fixes chaining (`za tzv. X`,
  `Mgr. et Mgr. Ing. Novák`).
- Titles **after** a name (`Ph.D.`, `CSc.`, …); two-letter words; `debug` mode.
- Longest-first title alternation; `plk.` typo fix; word boundaries so `viz`
  doesn't match inside `vize` and `PhD` doesn't match `PhDr.`.

Public API is unchanged (`#show: apply-vlna`). License stays MIT.

### DIFF
--- a/packages/preview/vlna/0.2.0/CHANGELOG.md
+++ b/packages/preview/vlna/0.2.0/CHANGELOG.md
@@ -1,0 +1,40 @@
+# Changelog
+
+## [0.2.0] – 2026
+
+Major rewrite of the core.
+
+### Fixed
+- **Jump-to-source preserved.** Feature #1 no longer rebuilds text with
+  `it.text.replace(" ", "~")` (which moved the source position of the affected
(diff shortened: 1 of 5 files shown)


saecki:
Thanks!

PR #5362, opened 2026-07-16
gbt9704-redhead:0.2.0

songwupei:
I am submitting
- [x] an update for a package

Description: GB/T 9704-2012 official Chinese government document format template. This update adds spacing themes (compact/normal/relaxed), table themes (full-grid/three-line/single-line/plain), table alignment control, page number positioning (both/center/left/right), a new endmatter() function, faux bold table headers for CJK fonts, and a bilingual README (EN + ZH). The src/ layout has been refactored into spacing.typ + table.typ + lib.typ.

- [X] selected a name that is not the most obvious or canonical name for what the package does
  - Explanation: "gbt9704-gongwen" combines the standard number (GB/T 9704) with the pinyin for "official document" (gongwen), making it Googleable and descriptive without being too generic.
- [X] added a typst.toml file with all required keys
- [X] added a README.md with documentation for my package
- [X] have chosen a license and added a LICENSE file
- [X] excluded PDFs or README images, if any, but not the LICENSE
- [X] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction

● New features since v0.1.0:

  - Decoupled table styling — table themes are now applied via tbl.wrap() decorators ("full-grid",
  "three-line", "single-line", "plain") instead of a single global table-theme parameter
  - endmatter() — bottom-aligned CC + issuing info block with separator lines
  - Page numbering — - 1 - format with "both" / "center" / "left" / "right" positioning;
  first-page-number toggle
  - show-red-header — toggle entire red-header block
  - Red line at bottom of first page (when redline: true)
  - Faux bold table headers via text(stroke: ...) for CJK fonts
  - Signature — 3 blank lines before block, signatory right-aligned in header
  - Refactored src/ into spacing.typ + table.typ + lib.typ
  - Bilingual README (EN + ZH)


packages/preview/gbt9704-gongwen/0.2.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "gbt9704-gongwen"
```
saecki, line 2: Since this package mainly targets a chinese audience, and `gongwen` is the romanization of `official document`, this name is a little too descriptive. We try to avoid canonical names so multiple packages serving the same purpose can coexist without one having an unfair advantage. Could you change the name to something else, for example by adding a non-descriptive prefix or suffix, or ideally by coming up with something a little more creative? See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules)

saecki (commented):
Thanks for this package!

songwupei:
rename: gbt9704-gongwen → gbt9704-redhead

  Package name changed per Typst preview review feedback. The previous
  name "gongwen" (romanization of "official document") was too
  descriptive per Typst's package naming guidelines. "redhead" is a
  creative, non-descriptive reference to the red header (红头) that
  characterizes Chinese official documents, while still keeping the
  GB/T 9704 standard identifier.

saecki:
Thanks!

PR #5366, opened 2026-07-16
codepoint:0.2.2

ReeseHatfield:
I am submitting
- [ ] a new package
- [x] an update for a package

- Fixed bug where  `code-block` failed to render if `lang` property not passed. Default `lang` is now `bash`
- Fixed bug where content type was checked against `string` instead of `str`
- Fixed various assertion errors where type(value) == none was used instead of value == none,


saecki:
Thanks!

PR #5370, opened 2026-07-17
sang-math:1.0.2

sangnhc87:
I am submitting

- [ ] a new package
- [x] an update for a package

Description: Patch release fixing an inline-math baseline regression in 1.0.1. The package no longer rewrites every inline `math.equation` to display style or detects fractions through `repr()`. It now applies display style only to `math.frac`, preserving Typst's natural baseline and line height in prose, tables, and narrow columns. Public API and package contents are otherwise unchanged.

Validation:

- Compiles with Typst 0.14.2
- Regression test verifies plain inline math and a tall fraction remain on the surrounding text baseline
- Tall fractions remain larger than explicit `tfrac`
- Compiled directly from `packages/preview/sang-math/1.0.2`
- All 13 project package tests pass
- Exam and book template demos compile
- No generated PDFs or test artifacts are included in the package directory

- [x] I have read and followed the package submission guidelines
- [x] I have tested the package locally and it worked

### DIFF
--- a/packages/preview/sang-math/1.0.2/LICENSE
+++ b/packages/preview/sang-math/1.0.2/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Sang Nguyen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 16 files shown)


saecki:
Thanks!

PR #5371, opened 2026-07-17
exercise-bank:0.6.1

nathan-ed:
I am submitting
- [x] an update for a package

Description: Adds `qr-position: "tasks"`, an opt-in QR placement mode that overlays the QR code and lets a taskize `#tasks(...)` body flow around it via a shared state key (no import coupling), plus documentation fixes for previously-undocumented utility functions.


elegaanz:
Thank you.

PR #5372, opened 2026-07-17
taskize:0.2.8

nathan-ed:
I am submitting
- [x] an update for a package

Description: Adds a `wrap-zone` parameter (and shared `taskize-wrap-zone` state) so `#tasks` rows flow around a reserved top-right rectangle, letting overlay content like a QR code or stamp share the block without manual layout math. Also finalizes an adaptive-row-heights improvement carried over from the prior release cycle.


saecki:
Thanks!

PR #5373, opened 2026-07-17
econ-working-paper:0.6.0

statzhero:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update to v0.6.0 of the **econ-working-paper** template. Endfloat mode can now move tables and figures independently (`endfloat: "tables"` / `"figures"` / `"both"`) and flush them at a chosen position (`endfloat-position`: end, after references, or after the appendix), with floats grouped tables-first. Endfloat placeholders and captions now keep their original numbers, supplements, and caption positions (including user overrides). Adds a `math-fontsize` option for inline math sizing and a `status` parameter for draft banners (e.g. "WORK IN PROGRESS") that is hidden in anonymized mode, replacing the former `subtitle` parameter.


elegaanz:
Thank you.

PR #5378, opened 2026-07-18
ansi-render:0.9.2

### LINKED ISSUE #8: empty control code should default to 0
### Description

According to some `\u{1b}[m` without control code should default to code 0 like `\u{1b}[0m`.

For example, from https://man7.org/linux/man-pages/man4/console_codes.4.html

> # ECMA-48 CSI sequences
> 
> CSI (or ESC [) is followed by a sequence of parameters, at most
> NPAR (16), that are decimal numbers separated by semicolons.  An
> empty or absent parameter is taken to be 0. ...

The problem is, `ansi-render` seems to not act on `\u{1b}[m`.

Right now, user can adhoc fix this by:

```typst
str.replace("\u{1b}[m", "\u{1b}[0m")
```

### Steps to reproduce

```typst
#import "@preview/ansi-render:0.9.1": ansi-render
#let stdout = "\u{1b}[32m Green \u{1b}[m Should not be green"
#ansi-render(stdout, font: none)
```

8LWXpg:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:
* Same behavior as `0` for empty control code.

### DIFF
--- a/packages/preview/ansi-render/0.9.2/CHANGELOG.md
+++ b/packages/preview/ansi-render/0.9.2/CHANGELOG.md
@@ -1,0 +1,113 @@
+# Changelog
+
+## [0.9.2] - 2026-07-18
+
+* Same behavior as `0` for empty control code.
+
+## [0.9.1] - 2026-05-11
+
+### Fixed
+
+* Fixed `bold-is-bright` option rendering logic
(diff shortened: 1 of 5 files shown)


saecki:
Thanks!

PR #5384, opened 2026-07-19
datify-core:2.1.0 and datify:1.3.0

Jeomhps:
I am submitting
- [x] an update for a package

Both packages are updated in one PR because `datify:1.3.0` depends on `datify-core:2.1.0`, which is not published yet.

## datify-core: 1.0.0 → 2.1.0

- **Core rebuild.** Locale data is now generated from the pinned `cldr-dates-full` npm package by a reproducible Node pipeline. The old monolithic data file is gone: the package reads a small index eagerly and loads each locale file lazily on demand (memoized), so importing the package loads no locale data.
- **Fallback chain.** Locale lookups follow CLDR truncation (e.g. `fr-CA` → `fr`) down to a configurable default locale, with cell-level fallback for values omitted as root-equal. Unknown locales fall back instead of panicking.
- **Opt-in community overlay.** Accessors gain a `community` parameter (default `false`). When enabled, lookups try a hand-maintained overlay before CLDR, for locales where community conventions differ from CLDR data (seeded with pt-BR). Data-only; presentation choices stay with the caller.
- Public API signatures are unchanged. Tests now use hand-verified cross-locale fixtures plus structural invariant checks on the generated data.

## datify: 1.0.1 → 1.3.0

- **New `display-date` function.** Returns content and formats using the document's `text.lang` and `text.region` (e.g. `en` + `GB` → `en-GB`) via `context`, so dates follow the document language without an explicit `lang` argument. `custom-date-format` stays a pure string function.
- **Rewritten tokenizer.** Patterns are scanned as maximal same-letter runs with lazy symbol resolution, which adds support for stand-alone months (`L`), stand-alone weekdays (`c`), narrow widths (length-5 tokens like `MMMMM`), and `yy`; unhandled symbols pass through verbatim.
- **Community overlay support.** `custom-date-format` and `display-date` gain a `community` flag (default `false`) passed through to datify-core.
- README restructured into an API Reference and a Pattern Reference; test suite expanded with per-function unit tests and a visual regression test across Latin/Cyrillic/Greek scripts.


saecki:
Thanks for the update!

PR #5389, opened 2026-07-20
typsy:0.2.4

patrick-kidger:
I am submitting
- [ ] a new package
- [x] an update for a package

Two QoL improvements:
1. Unit classes (those with no fields) are now their own instance, exactly as in Rust. In particular this makes enumerations of unit classes [a very common use case] much more pleasant to use.
2. When using a class for pattern matching, and failing, then the printed result is nicer: you're told it doesn't match the class name, rather than a complicated-looking nested structure.


elegaanz:
Thank you :)

PR #5390, opened 2026-07-20
typed-smiles:0.7.0

GeronimoCastano:
I am submitting

- [ ] a new package
- [x] an update for a package

Changes:

- Add explicit subscripts and superscripts to custom SMILES labels using `_(...)` and `^(...)`, with paired script positioning matching `ce()` notation.
- Add `rxn-arrow(scale:)` for proportionally resizing individual reaction arrows and their condition labels.
- Render `mol()` items placed in `rxn-arrow(above:/below:)` label slots as molecules; in mechanism mode they become referenceable species of their own, so curly arrows can target molecules above or below a reaction arrow.
- Support per-molecule `mol(scale:)` in mechanism mode, multiplying the shared `reaction(scale:)` canvas scale, with curly arrows and highlights tracking the scaled coordinates.
- Add `arrow(heads: "end"/"both"/"none")` and `arrow(style: "solid"/"dashed"/"wavy")` for curly mechanism arrows, both combinable with `bend`.
- Fix ordinary reaction layout bounds for species moved with `mol(offset:)`, so wrappers such as `brackets()` correctly follow shifted content.


saecki:
Thanks!

PR #5391, opened 2026-07-21
quick-vertex:0.1.1

13Stokes31:
## Changes since 0.1.0

- **Equal aspect ratio** (new `equal-aspect: true` default): both axes now
  share one units-per-length scale (and one tick step), so angles on screen
  match true angles — most notably, `∇Z` is now genuinely perpendicular to
  the level lines (previously it wasn't, whenever the x/y data ranges
  differed, which is the common case). `equal-aspect: false` restores the
  old independent-per-axis stretching, with `∇Z` still correct in that mode
  via a general on-screen formula.
- `∇Z` is also now correctly flipped for `sense: "min"` so it points toward
  the optimum rather than always toward increasing `Z`.
- Vertex labels are placed along the exterior bisector of the polygon at
  each vertex instead of a fixed corner offset, reducing collisions with
  edges/other vertices.
- Minor cleanups: native `.any()` for vertex deduplication; the
  "unbounded objective" message is now overflow-safe on narrow pages.

No breaking API changes — `equal-aspect` is additive and defaults to the
behavior most users want (true 1:1 axes).


elegaanz:
Can you please fix the warnings and errors reported in CI? At least the one about the outdated import in the example, and the missing ROADMAP.md file.

Regarding the ones that ask you to not link to your `main` branch directly, it exists because doing so can cause confusion if you push changes to the said branch in the future: the example file would no longer show what the latest published version does. Since you commit the image files directly to this repository, you can link them using a relative path (e.g `alternative text here`), which will work both on GitHub and Typst Universe and guarantee that they stay in sync with the current version.

elegaanz:
Thank you!

PR #5402, opened 2026-07-22
unofficial-pku-mpl:0.1.0

xjsongphy:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A Typst template for Modern Physics Laboratory reports at Peking University, ported from the revtex4-2-based PKUMpLtX template. It provides RevTeX-faithful report layout, Chinese font presets, frontmatter, ruled tables, bibliography helpers, and an initialized report template.

- [x] selected a package name that is not the most obvious or canonical name for what the package does
- [x] added typst.toml with all required keys
- [x] added README.md documentation
- [x] chosen a license and added LICENSE
- [x] tested the package locally and it compiles
- [x] excluded PDFs, demo files, and local test assets
- [x] ensured the template directory is licensed for normal user modification and distribution


xjsongphy:
The package checker now reports only the license field:

> The `license` field should be OSI approved

This package is derived from the CC-BY-SA-licensed PKUMpLtX project, so changing the manifest to MIT/Apache without upstream authorization would be incorrect. The Typst package licensing documentation also explicitly allows CC-BY-SA versions. Could this be reviewed as an allowed CC-BY-SA exception, or should the package-check rule/documentation be updated?

saecki:
Thanks for this package!

> This package is derived from the CC-BY-SA-licensed PKUMpLtX project, so changing the manifest to MIT/Apache without upstream authorization would be incorrect. The Typst package licensing documentation also explicitly allows CC-BY-SA versions. Could this be reviewed as an allowed CC-BY-SA exception, or should the package-check rule/documentation be updated?

You're correct, it should be updated. There is an open issue, but it hasn't been implemented: https://github.com/typst/package-check/issues/21
It's a bit unfortunate, because the CC license isn't really supposed to be used with source code, but in this case I think it's ok.

PR #5405, opened 2026-07-23
sang-math:1.0.3

sangnhc87:
Vietnamese high school math macros: variation tables (BBT), sign tables, MCQ, geometry with CeTZ. New: node-pad, arr-shorten, tab-val, annotations, overlay.


packages/preview/sang-math/1.0.3/thumbnails/1.png
elegaanz, line 1: Maybe this is intended, but the table inner margings and foreground/background contrast look wrong to me. You may want to double-check and fix it.

elegaanz: Also this file doesn't seem to be used in your README nor any other file. If that's right, could you either remove it from this repository or include it in your documentation? We try to only keep useful files in this repository.

elegaanz (commented):
Just two little remarks about the thumbnail image :)

sangnhc87:
Thanks for the review! I removed the unused thumbnail in commit ad87f50, since it was not referenced in the README or any other file. This addresses both comments about the thumbnail.

sangnhc87:
Thanks again for the review. I removed the unused thumbnail, restored backward compatibility with version 1.0.2, cleaned up the README, and verified the corrected book and beamer examples. The latest changes are in commit 23642f8.

PR #5408, opened 2026-07-23
typed-dsa:0.4.0

GeronimoCastano:
I am submitting

- [ ] a new package
- [x] an update for a package

Changes:

- Add localization: `language:` ("en"/"de"/"es") and `messages:` overrides for every caption-producing builder — trees, heaps, linked structures, stacks, queues, hash tables, graph traces (BFS/DFS/Dijkstra), and sorting traces.
- Add a `messages(...)` helper for building partial caption overrides that layer over any language, with `step-label:` still winning as the final override.
- Persist the selected language and overrides through chained `.result` objects and through `transition(...)`.


saecki:
Thanks!

PR #5409, opened 2026-07-23
nexus-tools:0.2.0

mayconfmelo:
I am submitting
- [ ] a new package
- [x] an update for a package

This version adds the ability to determine the "amount of light" in a color—ranging from 0 (darkest) to 1 (lightest)—and to derive different colors based on relative luminance, which is ideal for determining an optimal text color that contrasts with the background. It is now also possible to generate a `#datetime` object from an array, dictionary, or string.

- Added: `#get.relative-luminance()` of a color
- Added: `#get.dynamic-color()` to get different colors based on the relative luminance of a color
- Updated: `#date` support for arrays, dictionaries, and strings

### DIFF
--- a/packages/preview/nexus-tools/0.2.0/LICENSE
+++ b/packages/preview/nexus-tools/0.2.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 Maycon F. Melo
+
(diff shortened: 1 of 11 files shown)


elegaanz:
Thank you!

PR #5412, opened 2026-07-24
basic-cv-with-photo:0.2.11

joelfischerr:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed incorrectly rendered en dashes


saecki:
Thanks!

PR #5420, opened 2026-07-26
minerva-thesis:0.2.4

lvandevelde:
I am submitting
- [ ] a new package
- [x] an update for a package

This updated version (0.2.4) of minerva-thesis is similar to v0.2.3, except for some modifications to make it compilable with Typst 0.15.


saecki:
Thanks!

PR #5421, opened 2026-07-26
tcdm:0.0.3

YDX-2147483647:
I am submitting
- [x] an update for a package

Description:
- In previous versions, an edge case cause _error: expected string, found none_. Now this case is handled correctly.
- Update dependencies.


YDX-2147483647:
As in v0.0.1 #3748 and v0.0.2 #3846, all lint warnings and errors reported by the bot are false positives.

packages/preview/tcdm/0.0.3/README.md
```diff
@@ -0,0 +1,66 @@
+# TCDM Generator
+
+HTML generator for best-of lists, especially [Best of Typst (TCDM)](https://ydx-2147483647.github.io/best-of-typst/).
+
+It loads metadata fetched by the canonical [best-of-generator][], and generates an HTML index page.
```
elegaanz, line 5: The link here is empty.

packages/preview/tcdm/0.0.3/example/history_to_json.py
elegaanz, line 1: We generally prefer templates to only contain the Typst document and its assets. Would it be possible to move that script outside of it and integrate it somewhere else in the build process you describe in the README? Alternatively, I think you could achieve this CSV to JSON conversion directly in Typst, using bundle export.

elegaanz (commented):
Thank you for the template. Two small remarks.

packages/preview/tcdm/0.0.3/README.md
YDX-2147483647: This link is defined below on line 44.

elegaanz: My bad, I didn't know about the empty brace syntax! Thank you for pointing it out.

packages/preview/tcdm/0.0.3/example/history_to_json.py
elegaanz: If it's too much effort or if it doesn't fit the workflow so well to have that file somewhere else, it can also stay here, since previous versions already did that. We should have caught it earlier.

YDX-2147483647: > Would it be possible to move that script outside of it and integrate it somewhere else in the build process you describe in the README? 

If that's a strict rule, I can do it; but I still hope to include the script.

The script is coupled with the tcdm version. If `*.py` cannot be included in a template bundle, then I'd like add the script's content to each version's README.
However, moving a standalone script into the README does not change the total file size; it merely increases the difficulty for users to use the template. It's kind of pointless…


> achieve this CSV to JSON conversion directly in Typst

I tried that before tcdm v0.0.1, but it turned out that the CSV is quite specific to python & pandoc.
The following is a typical row. It contains python lists and numbers. It's even more complicated that some numbers should not be parsed as numbers because they are version numbers.

```csv
math,"['runnable', 'web']",,True,,2.8,2026-01-13 15:41:30.000,3.1.0,2.8,Apache-2.0
```

Therefore, it's theoretically possible to parse in typst, but that would be complex and fragile.

elegaanz: I see, then let's keep it that way.

PR #5427, opened 2026-07-27
sentinel-cv:0.1.0

davidculemann:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A traditional CV template. Small-caps section headings, fine rules, justified text, and the name and location sharing the top line. It is deliberately unremarkable, because it is aimed at readers who treat novelty as a risk: professional bodies, public-sector appointment panels, chartered institutions.

What it adds over the existing CV templates is the long tail of a conventional career. Alongside ordinary job entries it has purpose-built rows for publications, professional qualifications and awards, each shaped for what it is rather than forced through one generic entry function, plus a `format-dates` helper that drops either end of a range when it is empty. The sample is a chartered civil engineer, which is the sort of CV that actually needs all four.

  - Explanation:
    A sentinel is a guard posted to watch and to hold a position. It suits a design whose whole purpose is to look conservative and give a cautious reader nothing to object to.

    Following the guidance for template packages, the name is a unique, non-descriptive part (`sentinel`) followed by a descriptive part (`cv`). It is not affiliated with any organisation, so no entity abbreviation is involved, and it is not the canonical name for anything: `classic-cv`, `traditional-cv` and similar were deliberately avoided even though they were free.

- `typst init @preview/sentinel-cv:0.1.0` into an empty directory, then `typst compile`, gives no diagnostics.
- `compiler = "0.12.0"` was found by compiling against real 0.11, 0.12, 0.13 and 0.14 binaries rather than guessed. It compiles on 0.11 too, but the rest of the family I intend to submit needs `par(spacing:)`, so I have set one floor I can keep testing across the set.
- The thumbnail is page one as initialized, recompressed from Typst's output to 292 KB.


davidculemann:
Closing this one, I have others to submit which are more original, this one is derivative of fantastic-cv so doesnt add much value. Apologies.

PR #5430, opened 2026-07-28
grid-cv:0.1.0

davidculemann:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A plain single-column CV whose one structural idea is that skills belong in a two-column grid rather than a comma-separated run of twenty tools. Everything above that stays conventional on purpose, because this is the layout most likely to be read by software as well as by a person: one typeface, one accent on the section labels, no panels or sidebars for a parser to misread. Useful to anyone who wants an ATS-safe CV without giving up a readable skills block.

  - Explanation:
    `grid` names the design's one distinguishing feature, the two-column skills grid, in the same way `spine-cv`, `emblem-cv` and `horizon-cv` name theirs. It is not the canonical name for the functionality: that would be `cv`, `resume` or `curriculum-vitae`, none of which I am claiming. It is also the name this design already carries on the hosted builder linked from `homepage`, so someone arriving from Universe does not land on a page called something else.

    Following the guidance for template packages, the name is a unique, non-descriptive part (`grid`) followed by a descriptive part (`cv`). It is not affiliated with any organisation, so no entity abbreviation applies.

- `typst init @preview/grid-cv:0.1.0` into an empty directory, then `typst compile`, gives no diagnostics.
- `compiler` was established by compiling against real 0.11, 0.12, 0.13 and 0.14 binaries rather than guessed.
- The thumbnail is page one as initialized, recompressed from Typst's own output.


davidculemann:
Renamed this package from `lattice-cv` to `grid-cv` before review, and force-pushed the branch. Nothing else changed.

`lattice` was a name I invented when the design is called Grid everywhere else, including on the page `homepage` points at, so anyone arriving from Universe would have landed on something called by a different name. `grid` names the design's one distinguishing feature, the two-column skills grid, the same way `spine-cv` and `emblem-cv` name theirs. It is not the canonical name for the functionality; that would be `cv` or `resume`, which I am not claiming.

Sorry for the churn on the branch. The PR title and description below are updated to match.

saecki:
Thanks for this package!

> Sorry for the churn on the branch. The PR title and description below are updated to match.

No worries :)

davidculemann:
@saecki Thank you very much for merging! I have more templates from my repo which I want to add: [https://github.com/davidculemann/typst-templates](https://github.com/davidculemann/typst-templates)

What is the preferred release cadence? I can drip these gradually or make a giga PR with ~40 templates? 

All high quality, but don't want to drown anyone in reviews.

for reference:
image

saecki:
These templates seems more like a family of templates, so I think the best approach would be to group them into one package, that either multiple different template functions, or one functions that allows configuring the layout/style etc.
Ideally some of the common code should be abstracted and reused between the templates, though that's more of a personal preference than a requirement :)

davidculemann:
> These templates seems more like a family of templates, so I think the best approach would be to group them into one package, that either multiple different template functions, or one functions that allows configuring the layout/style etc. Ideally some of the common code should be abstracted and reused between the templates, though that's more of a personal preference than a requirement :)

OK a single package can work, will see if its possible abstract out any function

PR #5436, opened 2026-07-28
typed-dsa:0.5.0

GeronimoCastano:
I am submitting

- [ ] a new package
- [x] an update for a package

Changes:

- Invalid arguments, malformed structures, impossible operations, and customizations that name a node, edge, cell, or index that does not exist now fail with an actionable compiler error naming the function, the offending value, and the accepted values, instead of being silently ignored or surfacing later as an unrelated internal error.
- Unsuccessful lookups are unchanged: searching for a key that is not present remains a normal result reported through the step's `found:` field.


saecki:
Thanks!

PR #5442, opened 2026-07-29
malos-presets:1.4.0

MDLC01:
I am submitting
- [ ] a new package
- [x] an update for a package

Add a show rule for `divider`.


packages/preview/malos-presets/1.4.0/typst.toml
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.15.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks!

PR #5447, opened 2026-07-30
xyzrender-rustyp:0.1.0

jonathandip:
I am submitting
- [x] a new package

Description: Render XYZ/PDB molecular structures as depth-cued, publication-quality SVG graphics directly inside Typst documents. Bond orders, aromatic rings, and connectivity are perceived automatically from atomic geometry via a Rust → WASM port of [xyzgraph](https://github.com/aligfellow/xyzgraph). It's a Typst port of the Python [xyzrender](https://github.com/aligfellow/xyzrender) project (11 built-in presets, styling/gradient/fog control, manual bond rules, TS/NCI overlays, vdW spheres).

  - Explanation: The package is named `xyzrender` after the upstream Python project it ports (`aligfellow/xyzrender`), rather than a generic term like "molecule-render" or "xyz-viewer". The README and LICENSE both make the port relationship explicit and unambiguous (upstream is MIT-licensed and credited throughout), so this isn't presented as an official/first-party package, and there's no other Typst package for this functionality that the name would create an unfair advantage over.

Note on `src/assets/xyzrender.wasm`: this is a compiled Rust plugin doing XYZ/PDB parsing and bond/aromaticity perception (a faithful port of `xyzgraph`). Source lives in the `plugin/` directory of the package's [source repository](https://github.com/Typsium/xyzrender), built reproducibly via `plugin/build.sh` with `Cargo.lock` committed there. It's included pre-compiled because Universe's sandbox can't run `cargo build`.


elegaanz:
Regarding the WASM file, don't worry, this is how it is supposed to work.

On the name you chose:

> rather than a generic term like "molecule-render" or "xyz-viewer".

I think we can agree that `xyz-render` (with a dash) would be such a generic term, and `xyzrender` is not that different, so I'm not convinced that it respects the package naming rules.

> there's no other Typst package for this functionality that the name would create an unfair advantage over.

The naming rule also exists for *potential future packages*, not only already existing one.

For these reasons, could you pick another name please? It could be argued that the current name is fine, but it's on the edge and I would rather have a name that clearly respects the rules. We can discuss it beforehand if you want, to make sure you pick a name that would work.

jonathandip:
I’d be more than happy to discuss this! This package is a port to Typst of xyzrender, an existing Python program (I’ve spoken to the original author, and he’s agreed to this port of part of his repo); that’s why I thought I’d keep the name. That said, would you still advise me to choose a different name, even though it’s a port of an existing program? If so, I’ll speak to the original author and give it some thought shortly.

Generally speaking, I’d like to keep the similarity because I didn’t build the package from scratch – it’s a port – so is there really a need to completely change the name?

elegaanz:
No need to completely change it, but adding an extra word or just slightly modifying it so that there is enough space for an alternative port to be made without conflicting names would be nice. This is the main reason we don't want too descriptive package names in the first place.

jonathandip:
Would something like ‘xyzrender-rustyp’ be okay?

elegaanz:
Sorry for the long response delay. That name would work!

jonathandip:
I updated the package name in the typst.toml file and in the readme file

elegaanz:
Thank you.

PR #5448, opened 2026-07-30
modernpro-cv:2.0.0

jxpeng98:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: `modernpro-cv` provides polished academic and professional CV templates in single- and double-column layouts. It is useful for producing readable, restrained CVs with consistent typography, section hierarchy, and page headers while keeping the starter files easy to customize.

Summary:
- Redesigns the academic visual system with clearer entry rhythm and aligned continuation headers.
- Adds density presets and keeps the 1.x API backward compatible.
- Makes both starter CVs self-contained single files, with no separate `profile.typ` dependency.
- Replaces personal details with fully fictional example content.

Validation:
- Both `template/cv-single.typ` and `template/cv-double.typ` compile with Typst 0.15.1.
- `typst init` produces independently compilable starters without a `profile.typ` dependency.
- The repository bundler completes successfully.
- `@local/modernpro-cv:2.0.0` imports successfully.


saecki:
Thanks!

PR #5455, opened 2026-07-31
jurlstify:0.3.1

SchrodingerBlume:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:

`jurlstify-text`: a show rule that gives URL-shaped text line breaks anywhere - `#bibliography()` output, templates, bare URLs - leaving raw/code alone (state-based detection) and staying idempotent with `jurlstify-links` via break-marker probing. `"schemes"|"pattern"|"match-in-raw"|"linkify"` options.

Manual: vendor min-manual (MIT, LICENSE kept) and render the Reference via tidy with a min-manual-flavoured `#arg` style.

### DIFF
--- a/packages/preview/jurlstify/0.3.1/LICENSE
+++ b/packages/preview/jurlstify/0.3.1/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 SchrodingerBlume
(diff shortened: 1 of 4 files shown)


saecki:
Thanks!

PR #5457, opened 2026-07-31
theoframe:0.2.0

hexiongwu1995:
I am submitting
- [x] an update for a package

Version:0.2.0
- Fix: Counter not resetting when encountering a level-1 heading (= heading).
- Root Cause Analysis: In Typst, #import only brings in variable bindings (functions and variables defined with #let) from a module. Meanwhile, #show rules are document-level directives, meaning their scope is strictly confined to the module where they are defined.


hexiongwu1995:
How long does the review process normally take?

saecki:
Thanks for the update!

> How long does the review process normally take?

It fluctuates a bit, depending on what else is going on. But we try to review everything in at least a couple of days. For this week I'm also doing package review alone, because Ana is on vacation :)

PR #5463, opened 2026-08-01
pku-thesis-pass:0.2.0

chuxinyuan:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Blind review mode is now handled inside the template itself: the `acknowledgements()` and `declaration()` closures auto-skip when `blind: true`, so users no longer need `#if not blind { ... }` wrappers (fully backward compatible). The `config()` return dict was reordered to match document flow, and docs/comments were cleaned up.

Changes:

- Blind review mode is now handled by the template itself. The `acknowledgements()` and `declaration()` closures now automatically skip themselves when blind review mode is enabled (`blind: true`). Users no longer need to wrap these calls in an `#if not blind { ... }` block in their `thesis.typ`. Fully backward compatible — existing documents keep working as-is.

- Returned dictionary order aligned with document flow. Keys of the dictionary returned by `config()` were reordered so that `appendix`, `acknowledgements` and `declaration` follow the order in which they appear in a thesis.

- According to the PKU thesis format guidelines, the footnote style has been corrected.

- Fix punctuation in copyright notice and update wording for thesis publication delay clause.

- Replace references in ref.bib with test entries.

- Documentation and comments improved.
  - README: logo/wordmark setup now points to the CTAN `pkuthss` package and uses `pkulogo.pdf` / `pkuword.pdf` filenames
  - Template files: clearer numbered comments for cover logo setup, unified section comments, and a clearer import-switch comment for local testing.

Checklist:

- [x] Version in `typst.toml` bumped to 0.2.0
- [x] All `@preview/pku-thesis-pass:0.1.0` references updated to 0.2.0
- [x] Template compiles in both normal and blind mode


saecki:
Thanks!

PR #5466, opened 2026-08-01
tue-thesis:0.1.0

alikaansun:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Thesis template for the Technical University of Eindhoven 

  - Explanation:
    

packages/preview/tue-thesis/0.1.0/src/thesis.typ
```diff
@@ -0,0 +1,136 @@
+// Style layer: page geometry, typography, headings, running headers/footers,
```
saecki, line 1: The library code should be separated from the template (example) code of this package.
The library code should live outside of the template directory, and should only be accessed through package specification imports (`@preview/<name>:<version>`).

See https://github.com/typst/packages/blob/main/docs/manifest.md#templates
and https://github.com/typst/templates/tree/main/charged-ieee for a simple example.

saecki (commented):
Thanks for this package!

packages/preview/tue-thesis/0.1.0/src/thesis.typ
alikaansun: Thank you for the review I believe I applied the changes as required now. And tested the compiled correctly. 

I made the library files in src then imported thru *src/...

saecki: Thanks, looks good!

saecki:
Sorry, I just noticed the name of this package doesn't fit the naming rules. I've reverted thise PR, since the package hasn't been deployed to Typst Universe yet. Could you open a new PR to continue the review?

If this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

alikaansun:
Ah I see. thank you for the explanation. I have created the pull request #5526 in return. Changed the name to definitely-not-tue-thesis :) Hope its ok

PR #5467, opened 2026-08-01
qooklet:0.7.0

ivaquero:
I am submitting
- [x] an update for a package

- fix several font errors
- update numbering on blank pages
- update header and footer on blank pages
- improve code readability


packages/local/qooklet
saecki, line 1: Could you remove these subprojects?

saecki (commented):
Thanks for the update!

ivaquero:
@saecki cleaned the unnecessary files

saecki:
Thanks!

PR #5468, opened 2026-08-01
touying-quick:0.5.0

ivaquero:
I am submitting
- [x] an update for a package

- fix font warnings
- fix eq numbering
- add new background images
- add new themes


packages/local/qooklet
saecki, line 1: Same as the `qooklet` PR, could you remove the subprojects?

saecki (commented):
Thanks for the update!

packages/local/qooklet
ivaquero: Sure, I will do this in the evening

ivaquero:
@saecki cleaned the unnecessary files

saecki:
Thanks!

PR #5473, opened 2026-08-02
conic-toan:0.3.2

Thutran1891:
Update of `conic-toan` from 0.3.1 to 0.3.2. No breaking API changes; the minimum compiler version stays at 0.13.0.

### New

- **Exam shuffling** — `de-toan(hoan-vi: true | "cau" | "pa")`. Only questions of the same type standing next to each other are shuffled, so a question never crosses a section heading, and the sub-items of true/false questions keep their order. The generator is deterministic and seeded from the exam code, so the exam PDF and the answer-key PDF always come out in the same order.
- **Two-column question layouts** — `#show: chia-2-cot` (equal columns, balanced by measuring the content, since Typst otherwise fills the first column to the full page height) and `#show: chia-2-cot-lech(rong-trai: 70%)` (questions left, ruled writing space right). `#thoi-cot()` ends the columns and is required before `#het()` / `#bang-dap-an()`, because a page break cannot happen inside columns.
- **`ke-het-trang`** — fill the rest of the current page with ruled lines without stating a line count; `them-trang: n` adds n more full pages.
- **Text wrapping around figures** (`om-hinh`) — with a long solution and a small figure, text now wraps around the figure instead of leaving half a column empty. Splitting happens at paragraph boundaries, never mid-line, so nested math cannot break.
- **New module `lib/mat-cong.typ`** — cones and cylinders that work out their own hidden lines. The old `hinh-non` / `hinh-tru` hard-coded "front half solid, back half dashed"; the new module casts a ray from each mesh segment towards the viewer, so it handles self-occlusion *and* one solid hiding another (a cone poking into a cylinder, a cylinder inscribed in a cone). Silhouette generators are solved analytically, and axes can be tilted (`nghieng:`, `huong:`, `truc:`) using an orthographic projection so the whole figure shares one viewing angle. Spheres are not supported yet.
- **Block equations in multiple-choice options are pulled back inline** — writing `$ cases(...) $` with spaces next to the dollars makes Typst treat it as a display equation, so every option took a tall block of its own. `tn` / `ds` now bring those inline, which usually lets all four options fit on one row.
- **Line spacing factor `gian-dong`** at four scopes: whole document, from a point onwards (`#gian-dong(1.4)`), a single question, or one block (`#voi-gian-dong(1.7)[...]`). Useful when nested fractions or radicals make consecutive lines collide; `1.0` reproduces the previous layout.
- **`ve-goc(A, O, B)` / `ve-goc-vuong(A, O, B)`** — TikZ-style argument order with the vertex in the middle, alongside the existing `goc(O, A, B)`.
- **Doc comments** (`///`) on the functions re-exported in `baigiang.typ` and in `ve.typ`, so editors can show parameter hints.

### Fixed

- **Variation and sign tables pad the `dau` array.** It must hold `2n - 1` entries with an empty string at each end; one or two missing entries are now inferred instead of raising `array index out of bounds`. `dau` also accepts bare numbers, and `x` / `gia-tri` accept `"-oo"` / `"+oo"` and bare numbers.
- **Variation tables line up.** Arrows now aim at the visual centre of each value — they were off by 4-6pt because a glyph box includes ascender and descender space — and the `||` mark for an undefined point is drawn across the full cell height.
- **README.** The `gian-dong` and `chia-2-cot` examples wrote a bare `...` inside a call, which parses as the spread operator; replaced with concrete arguments. A shell block is now tagged `sh` so it is not read as Typst.

### Opt-outs

The only behaviour change for existing documents is the figure wrapping, which is on by default. Switch it off per figure with `voi-hinh(..., om: false)` or document-wide with `#kieu-cau-hoi(om-hinh: false)`. The inline-equation change is opt-out too, per question with `trong-dong: false` or document-wide with `#kieu-cau-hoi(eq-trong-dong: false)`.


saecki:
Thanks!

Thutran1891:
Thank you for merging, and for reviewing all three releases so patiently.

Having `conic-toan` on the registry means Vietnamese high-school maths teachers can build their lecture slides and exam papers without installing anything locally — that makes a real difference for them. Much appreciated!

PR #5474, opened 2026-08-02
tidymind:0.2.0

pierryangelo:
Adds tidymind 0.2.0. The package draws horizontal, tidy mind maps on CeTZ, measuring every node before layout so long labels never overlap.

## What is new

- **`style: "outline"`** — an alternative that drops the boxes: the root becomes a heading over a baseline rule, first-level branches rest on a rule in their own color, and deeper nodes are plain text. Useful when the map sits inside a document and frames would fight with the surrounding text.
- **`branch` and `emphasis` on `node`** — a caller marks a node's role (`highlight`, `warning`, `definition`, `example`) or overrides its branch color *by name*, and the package resolves the color. `ink` and `emphasis-colors` on `mindmap` make those names configurable.
- **A measurement that matches the drawing.** Measuring and drawing were two hand-mirrored functions in two files, and an emphasized node ended up measured in `regular` but drawn in `semibold` — the label outgrew the box the layout had reserved for it and hyphenated inside it. Both passes now go through a single `node-body` in `src/style.typ`, and `examples/_assert_style.typ` asserts that what the layout reserves equals what the drawing emits, for every style, depth and emphasis.

## Compatibility

The default is unchanged: `style: "boxed"` renders exactly what 0.1.1 renders. I verified this by compiling the four examples published with 0.1.1 against both versions and comparing the rendered PNGs — they are byte-for-byte identical.

## Notes for review

- The `authors` field changes from "Fora da Curva" to "Elitus". This is a rebrand of the same entity, not a change of maintainer; `Pierry A. Pereira <@pierryangelo>`, the author of 0.1.1, is unchanged and is submitting this version.
- README images are served from the package repository over absolute URLs pinned to the `v0.2.0` tag, so they add nothing to the archive.
- `examples/` is committed but excluded from the archive. Files named `visual_*` produce the README images; files named `_assert_*` exercise the logic through `#assert`, so compiling them is the test suite. All 16 compile.

### DIFF
--- a/packages/preview/tidymind/0.2.0/LICENSE
+++ b/packages/preview/tidymind/0.2.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Elitus, Pierry A. Pereira
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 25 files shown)


saecki:
Thanks for the update!

PR #5482, opened 2026-08-04
meridian-cv:0.1.0

davidculemann:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A serif CV built on a single asymmetry: the masthead sets flush right and everything below it flush left. The eye lands on the name in the top corner, drops to the first section label, then follows one straight edge down the page, which gives the layout a focal point without spending a rule or a filled band on it. The entry function collapses any field you leave out, so the same call sets a job, a degree, a grant or a citation, which makes it practical for academic CVs that mix all four.

  - Explanation:
    A meridian is a single line of longitude. The design is built on one vertical axis: the masthead sets flush right and everything below it flush left.

    Following the guidance for template packages, the name is a unique, non-descriptive part (`meridian`) followed by a descriptive part (`cv`). It is not affiliated with any organisation, so no entity abbreviation applies.

- `typst init @preview/meridian-cv:0.1.0` into an empty directory, then `typst compile`, gives no diagnostics.
- `compiler` was established by compiling against real 0.11, 0.12, 0.13 and 0.14 binaries rather than guessed.
- The thumbnail is page one as initialized, recompressed from Typst's own output.


saecki:
Thanks for this package!

PR #5488, opened 2026-08-04
staunton:2.0.0

ndg6:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: staunton — chess publishing for Typst: boards, FEN/PGN diagrams,
tables, outlines. This is a breaking 2.0.0 release that redesigns the game /
PGN surface around two rules: there is one way to do each thing, and a move is
something you pass rather than something smuggled inside a value.
`parse-pgn` / `parse-fen` become `games` / `game` / `position`;
`diagram-after(g, "17w")` becomes `diagram(g, at: "17w")`; positions no longer
carry a hidden move payload, so move-quality badges and `%cal` / `%csl`
annotations require the game plus `at:`. Nearly every function is now
`f(subject, setting: ..)` — one positional argument, the rest named. Also new:
arrow tip and fade styling (the barbed `"hook"` tip is the new default), a
`last-move` marking, and figurine notation (♘f3) accepted as input alongside
localized SAN. No aliases were kept for the renames. Full changelog at:
https://github.com/ndg6/staunton/blob/v2.0.0/README.md#200

Note: 1.1.0 was a GitHub-only release and was never submitted here, so this
updates Universe from 1.0.0 straight to 2.0.0.


saecki:
Thanks!

PR #5493, opened 2026-08-05
genotypst:0.11.0, genotypst:0.6.0, genotypst:0.9.0 and sdu-touying-simpl:1.1.0

cskeeters:
This is a simple change to remove `.DS_Store` files that somehow got committed.


Dregen-Yor:
> @Dregen-Yor You released sdu-touying-simpl:1.0.1, so you probably want to have a look at this pull request. If you want this update to be merged, please leave a comment stating so. Without your permission, the pull request will not be merged.

OK

saecki:
Thank you, but we don't allow package edits: https://github.com/typst/packages/issues/2671

Fortunately the `.DS_Store` files won't be bundled into downloaded packages anyway because they're hidden, so this shouldn't be a real issue.

cskeeters:
I get the decision that was made, but I want to share my experience.

I run macOS.  One of the things I dislike about it is the `.DS_Store` files that Finder creates.  I wrote a program called [rmds](https://github.com/cskeeters/rmds) that uses macOS's File System Events API to delete the files as soon as they are created.  It's meant to remove files that Finder creates, and it's worked so well for me that I forgot about it.  Here, it deletes files that `git` creates when checking out `main`.  As a result, these 5 files are deleted locally, and `git status` shows that they were removed, and I have to work around them.  If I try to restore/discard the changes, `rmds` immediately removes the files again as if the `git restore` command didn't work.  This is why, for me and maybe a few others who run similar tools that auto-remove `.DS_Store` files, it's a bit annoying to have to work around these files.  Unlike other software, where the issue is resolved in a new version and then the problem no longer exists when working with the tip of `main`/`master`, this issue, with the current policy, will persist forever.

saecki:
Hey, we've discussed this, and we sort of agree that it would be nice if we could remove some of these artifact files. But it's hard to justify breaking our promise of immutability over 5 files.
The case regarding your tool is interesting and I hadn't considered it. Though the problem seems a little self-inflicted. Have you considered adding a configuration option that excludes specific directories on your machine from being manged by `rmds`?

cskeeters:
I understand; I think I'm in the vast minority of people who will have an issue.  I can just turn off the LaunchAgent while I'm working in `typst-packages`.  I don't work in it that much.  Or, yes, I can fairly easily modify the program to ignore a set of configured folders.

To be clear, `rmds` is not the only program that does this.  There is:
- [Asepsis](https://asepsis.binaryage.com/)
- [BlueHarvest](https://github.com/BlueHarvest-Mac-Cleanup-Tool)

And non-real-time tools such as:
- [DS_Store Cleaner](https://apps.apple.com/us/app/ds-store-cleaner/id6748859939?mt=12)
- [Onyx](https://www.titanium-software.fr/en/onyx.html)
- [CleanMyMac](https://macpaw.com/cleanmymac)

I only wrote my own because macOS's security features prevented the existing real-time tools from working.  This is just to say that if this is self-inflicted, there are many like me who installed software to get rid of these files.

cskeeters:
Turns out I wrote a TOML config file for `rmds`, and folders can be excluded.  I configred it and it works.

https://github.com/cskeeters/rmds#configure

PR #5523, opened 2026-08-09
iarticle:0.1.0

hideo67:
Hello. I am submitting
- [x] a new package
- [ ] an update for a package

Description: A LaTeX article/report inspired template with internationalization support. Keywords such as "Chapter"
may be localized by resource files.  Currently "en" and "ja" is supported. Other locales can be added with minimal
modification to the main lib.typ file(The lib.typ file contains a list of supported locales).

Japanese users typically need to use a fully japanese template because of the font names and label strings.
If i18n capable templates become standard, CJK(Chinese, Japanese and Korean)
users and western users can share the same templates and will improve overall productivity.
This simple template demonstrates one way to implement the i18n functionality in templates.
And it is useable for simple documents.

  - Explanation:
    - The name iarticle takes i from i18n and article to describe the target.
    - The most obvious name might be internationalized-article or i18n-article. I judged that the combined noun iarticle is an artificial (or creative) term that is not necessarily canonical.


elegaanz:
Thank you.

PR #5528, opened 2026-08-09
conic-toan:0.3.3

Thutran1891:
Update of `conic-toan` from 0.3.2 to 0.3.3. No breaking API changes; the minimum compiler version stays at 0.13.0.

### New

- **`cao-that` — inline formulas no longer collide with the line above.** Typst sizes an inline equation from the font metrics (cap height down to the baseline, ~6.83pt at 10pt), not from the ink it actually draws: `measure($1/2$)` and `measure($0,5$)` return the same height, while the fraction is drawn about twice as tall and overflows on both sides. Two consequences: inside a paragraph the numerator of one line touches the denominator of the line above, and inside a `grid` cell (the option grid of a multiple-choice question, the sub-items of a true/false question, `cot-item`) the cell inherits that same short height, so the fraction spills into the row above — which `gian-dong` cannot reach, since it only touches `par.leading`. The library now measures the real overflow and inserts a **zero-width** strut just before the equation, so both the line and the grid cell grow by exactly what is missing. Formulas that do not overflow past a 1pt threshold (`$x$`, `$0,5$`, `$x_(n+1)^2$`) get no strut at all, so existing layouts are unchanged except where fractions were actually colliding. The measurement is calibrated against a lowercase sample rather than against the font box, because ordinary text overflows its box too and `leading` already accounts for that. It also holds when the document enables `#show math.equation.where(block: false): it => math.display(it)`, a common way to get LaTeX-style `\displaystyle` fractions, which makes the problem noticeably worse.
- **`gian-dong` now reaches grids and the gap between questions.** Previously the factor only applied to text inside one content block, so setting `gian-dong: 3` tripled the leading within a question while the next question heading still sat flush against the last sub-item. The row gutters of the option grid, the true/false sub-items, `cot-item(cach-hang: auto)`, the prompt-to-options gap and the block spacing between questions now all scale with the same factor. A factor of `1.0` reproduces the previous numbers exactly.
- **`mui-ten-2-dau`** — a dimension line: arrowheads at both ends, the measurement centred on the shaft, the shaft split and padded around the label. `trong: false` puts the label outside, `vach:` adds end ticks, `le:` insets both ends, `ten-quay:` sets the label upright. Segments too short for an inline label move it outside automatically.
- **`khung-vua(..items, le:)`** — returns `(xmin:, xmax:, ymin:, ymax:)` fitted around the given objects, meant to be spread into `#hinh(..khung-vua(...))`. A circumscribed circle sticks out past its polygon and used to fall outside the drawing frame.
- **`tron-qua-diem(points)`** — least-squares circle through an array of points (Kåsa); with three points it agrees with `tam-ngoai-tiep`.
- **`co-net(content)`** — measures the real ink height of a piece of content, as opposed to `measure`, which returns the font box.
- **`phan-giac`** gains `vach: 0..3` (tick marks across the arc), `so-cung: 1..3`, `dai-vach:`, `ten-goc:` and `lech:`.

### Fixed

- **Labels clipped by their own ink.** `nhan`, `o-bo-tron` and the result cells of `so-do-cay` sized their boxes from the font metrics, so a fraction inside a label was cut off. They now take `max(font box, ink bounds)`. Plain text has zero overflow, so existing figures keep their layout.
- **`duong-tron-ngoai-tiep` accepts a polygon**, either `(A, B, C)` as before or an array of vertices. The centre label with `huong-tam: auto` picks the roomiest direction instead of landing on an edge, and `ban-kinh: true` picks a vertex whose radius does not lie along an edge (for a right triangle it uses the right-angle vertex rather than drawing over the hypotenuse). Adds `ten-r:`, `canh:`, `to:`.
- **Statistical charts no longer draw the horizontal axis through its own label.** The label was placed at the arrow tip (`above-left`); it now sits to the right, and `xmax` is widened by exactly the label width so the text stays inside the figure. Applies to `bieu-do-tan-so`, `da-giac-tan-so`, `bieu-do-cot` and `bieu-do-hop`.
- **Slides no longer reprint the prompt in the middle of a solution.** A slide reveals its solution step by step and each step is laid out as a full page, so as the steps grow the content eventually exceeds the slide body; Typst pushed the overflow onto a following page while the next step started again from the top, interleaving "prompt + first lines", "overflow", "prompt + first lines" again. On one real lecture a single slide produced five pages, two of them identical. The library now measures the rendered height of each step against the slide body and splits the material across continuation screens by binary-searching for the largest number of lines that still fits — the same thing an author would do by hand with `#sang-man`. If the prompt alone leaves no room, the first screen carries only the prompt and the next one is labelled with the solution heading rather than "(cont.)".
- **`#sang-man` no longer swallows the text sharing its line.** The splitter cut the material at `linebreak`/`parbreak` boundaries and treated any line *containing* the marker as a separator, dropping that whole line. Authors commonly end a sentence without a trailing `\` and write `#sang-man \` on the next source line, which puts the sentence and the marker on the same line — so the sentence vanished, from the slides and from the printed solution alike. The line is now cut in two at the marker, searching direct children only so that wrappers such as `strong` are not unwrapped; a marker nested deeper leaves the line untouched rather than risking the text. Documents without `#sang-man` are byte-identical to before.
- **`chia-2-cot` balances the last page of a multi-page question block.** Typst's `columns` fills column 1 down to the bottom before spilling into column 2, so a section of questions running over several pages ended with a full left column and an empty right one — half a sheet wasted. Earlier pages are still filled by Typst's own flow, which can split a long question across columns and beats any manual packing; only the material that lands on the last page is redistributed evenly with `#colbreak()`. A 25-question sample paper with figures went from 7 pages to 6. Bodies generated with `#for` are handled as well: Typst wraps such a loop in a single element, which previously left no seam to break at. Blocks that fit on one page, and documents that do not use `chia-2-cot`, are unchanged.

### Opt-outs

`cao-that` is on by default. Turn it off from a point onwards with `#cao-that(false)`, for a whole document with `#kieu-cau-hoi(cao-that: false)`, or widen it with `#cao-that(them: 1pt)`. `#cao-that(hien: true)` prints the missing height in points after each formula, which is handy for checking a document before changing anything.

The automatic screen splitting can be disabled with `#tu-ngat-man(false)`, and `chia-2-cot(can: false)` restores Typst's plain column filling.


saecki:
Thanks for the update!

Thutran1891:
Thank you for merging, and for reviewing all three releases so patiently.

Having conic-toan on the registry means Vietnamese high-school maths teachers can build their lecture slides and exam papers without installing anything locally — that makes a real difference for them. Much appreciated!

PR #5533, opened 2026-08-10
cekinar:0.1.0

hpcfzl:
I am submitting:
- [x] a new package
- [ ] an update for a package

Description:
* See `typst.toml`:
  * _Slovenian UPN QR payment slips for completing payment orders._
* More:
  * This template follows [the UPN QR standard](https://www.zbs-giz.si/wp-content/uploads/2021/10/EN_Tehnicni_standard_UPN_QR.pdf), issued by the Bank Association of Slovenia

I have read and followed the submission guidelines and, in particular, I:
- [x] Selected [a name](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) that isn't the most obvious or canonical name for what the package does
  - Explanation: the package name _Cekinar_ is a play on the word _cekin_ which stands for _a coin_
- [x] Added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys
- [x] Added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation for my package
- [x] Have chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) and added a `LICENSE` file or linked one in my `README.md`
- [x] Tested my package locally on my system and it worked
- [x] Ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use

Preview:
UPN QR SVG

Note to reviewers:
* There exists a package for this purpose, however I received no response to [my issue there](https://github.com/Olaii/olaii-upn-qr-typst-template/issues/1)

A couple of ambiguities:
* Should the README links to files within the repository be web links or would Typst Universe automatically resolve them as such?
* When README material is excluded, as recommended by [`tips.md`](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude), anyone who views its file offline won't have them available, hence why I used `alt`:
  * This is apparent even on the CeTZ _good example_ README

Please excuse the use of tabs everywhere but inside the licence, I'm pro-tabs.

Edits:
* I think that the licence is often `.txt`, or even `.md` as is mine, so I think there should be a notice that only `LICENSE` is allowed.
* About GFM task lists, how could I have anticipated their lack of support on Typst Universe?
* The couple of warnings (the GFM task lists and the exclusions' `./`) which are left, I'm okay with unless they're too strange.


packages/preview/cekinar/0.1.0/cekinar/moduli/qr/qrcodegen/iso_iec_8859_2.zig
elegaanz, line 1: In general, there is no need to include the source of WASM plugins on Typst Universe. Can you remove all that and only keep the WASM binary please?

elegaanz (commented):
Thank you for building this package :) Thank you also for clarifying the situation regarding the other package.

The way you linked files in your README will work on Typst Universe. Regarding the excluded assets, the way you did it seems correct. It is expected that the offline experience will be degraded: we prioritize lightweight packages, as most people download packages via the compiler to build a document, and reading the documentation locally is more of a last resort in case you can't access the online one.

No worries about the tabs, you decide of your code style (as long as the public API uses kebab-case).

Regarding the `.txt` or `.md` extensions for license files and the lack of support for GFM tasks lists, do you think adding this information in the docs would have helped you to avoid these mistakes? If so, feel free to make a PR to add that :) Otherwise, what would have helped?

packages/preview/cekinar/0.1.0/cekinar/moduli/qr/qrcodegen/iso_iec_8859_2.zig
hpcfzl: Sure, I suppose these are _**3. Other files (don't commit)**_.

I was more worried that the examples inside `./primeri/` might not need the images since the PDF is already there, however its source won't be able to compile, so should I only have the examples PDF there?

hpcfzl:
I have submitted a separate PR https://github.com/typst/packages/pull/5541 as suggested, to provide more information about the `LICENSE` file and GFM task lists in `typst/packages` docs.

I would otherwise recommend there be a file `typst-package-check.md` which lists all checks performed by the bot.

elegaanz:
Thank you for the other PR. Regarding the check list, this is something we could do, but I don't really see the benefit of having it since you can already run the tool locally before pushing.

PR #5537, opened 2026-08-10
primaviz:0.9.0

phiat:
## Summary

primaviz is a charting library with 50+ chart types, 7 themes, and zero dependencies — built entirely with native primitives (`rect`, `circle`, `line`, `polygon`, `curve`, `place`).

This release adds content labels, fixes a rendering defect in line-based charts, and implements three parameters that were accepted but silently ignored.

## Changes since 0.8.0

**Content labels** — labels and series names accept arbitrary content (math, markup) anywhere a string previously worked, across every chart type:

```typst
#stacked-bar-chart((
  labels: ($alpha$, $beta$),
  series: (
    (name: $x$, values: (90, 0)),
    (name: [*y*], values: (10, 100)),
  ),
))
```

**Joined line segments** — linear line series were drawn as one `line()` per segment, so adjacent segments met with butt caps, leaving notches at every vertex on thick strokes and restarting dash patterns. Series now emit a single `curve()`. The same fix applies to area, bump, dual-axis, metric sparklines, and sparkline.

**Three revived parameters** — `line-chart`'s `fill-area` (plus a new `fill-opacity`) and `multi-bubble-chart`'s `x-label` / `y-label` were documented but never read; they now work.

No breaking changes.

## Checklist

- [x] `typst.toml` version matches the directory name (`0.9.0`)
- [x] Package compiles against the declared compiler floor — verified against Typst 0.13.0 with zero warnings
- [x] Examples import `@preview/primaviz:0.9.0` and compile against this staged copy
- [x] `exclude` keeps `examples/`, `data/`, and `screenshots/` out of the download archive
- [x] MIT licensed, `LICENSE` included

## Links

- Repository: https://github.com/phiat/primaviz
- Release: https://github.com/phiat/primaviz/releases/tag/v0.9.0
- Full changelog: https://github.com/phiat/primaviz/compare/v0.8.0...v0.9.0

---

*Note: this PR was briefly and incorrectly retitled to `primaviz:0.9.1` a few minutes after it was merged — see the comment below. It shipped 0.9.0, as the title and body above describe. 0.9.1 follows in a separate PR.*

PR #5538, opened 2026-08-10
primaviz:0.9.1

phiat:
## Summary

primaviz is a charting library with 50+ chart types, 7 themes, and zero dependencies — built entirely with native primitives (`rect`, `circle`, `line`, `polygon`, `curve`, `place`).

A small follow-up to 0.9.0 (merged in #5537): two bug fixes and full parameter documentation.

## Changes since 0.9.0

**`pie-chart`'s `radius` parameter had no effect.** It was shadowed by the pie's own geometric radius (`let radius = size / 2`), so callers passing a corner radius got nothing, and the container was instead rounded by half the chart size. Pie charts on a theme with a background or border now render with square corners by default, matching every other chart type; pass `radius:` to round them.

**`radar-chart`'s docstring was silently truncated.** `axis-max-value` was documented with `//` instead of `///`. A non-doc comment terminates the docstring, so `data` and `size` — both correctly written above it — were dropped from the generated docs as well.

**Documentation** — all 58 previously undocumented parameters now appear in the generated docs. Mostly options that already worked but were invisible: `annotations`, `subtitle`, `radius`, `show-ticks`, `show-minor-grid`, the error-bar options on `bar-chart` and `line-chart`, `size-label` on both bubble charts, and the entire `ring-progress` signature.

No breaking API changes.

## Checklist

- [x] `typst.toml` version matches the directory name (`0.9.1`)
- [x] Package compiles against the declared compiler floor — full test suite, both examples, and all 21 per-chart demos verified against Typst 0.13.0 with zero warnings
- [x] Examples import `@preview/primaviz:0.9.1` and compile against this staged copy
- [x] `exclude` keeps `examples/`, `data/`, and `screenshots/` out of the download archive while leaving them available here for README rendering
- [x] File set is identical to the published 0.9.0
- [x] Adds only `packages/preview/primaviz/0.9.1/`; nothing else in the tree is touched
- [x] MIT licensed, `LICENSE` included

## Links

- Repository: https://github.com/phiat/primaviz
- Release: https://github.com/phiat/primaviz/releases/tag/v0.9.1
- Full changelog: https://github.com/phiat/primaviz/compare/v0.9.0...v0.9.1


saecki:
Thanks!

PR #5541, opened 2026-08-10
Clarify LICENSE and allowed Markdown

hpcfzl:
From https://github.com/typst/packages/pull/5533#pullrequestreview-4897810736:

> Regarding the `.txt` or `.md` extensions for license files and the lack of support for GFM tasks lists, do you think adding this information in the docs would have helped you to avoid these mistakes? If so, feel free to make a PR to add that :) Otherwise, what would have helped?


elegaanz:
Thank you!

PR #5543, opened 2026-08-10
exercise-bank:0.6.2

nathan-ed:
I am submitting
- [x] an update for a package

Description: Adds `badge-position: "above"`, which puts the exercise badge on its own header line so the statement can use the full measure instead of losing a left column on every line — useful in narrow columns and with enumerated statements. Also fixes a layout instability in the `link-style: "page"` solution reference, which could fail to converge when combined with the margin badge position.


elegaanz:
Thank you.

PR #5546, opened 2026-08-10
easy-hgb-thesis:0.2.0

TimerErTim:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added manual, documentation improvements, and most importantly a completely new overall thesis base style that more closesly matches inofficiell but established university templates.


saecki:
Thanks!

PR #5547, opened 2026-08-11
invaria:0.2.0

q-wertz:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A Typst package containing a lot of (physical) constants and their metadata (derived from the [NIST CODATA database](https://physics.nist.gov/cuu/Constants/index.html)). Can be useful when doing calculations in Typst that require constants.

  - Explanation: Derived from "invariant" showing the constant nature of the information that is provided
    

elegaanz:
Thank you for building this package. Regarding the license, do you know if the NIST data is open source and if they allow derivative works like this package? And regarding the package itself, it seems to use snake_case, while kebab-case is the preferred naming convention in the Typst ecosystem. Unless you have a strong preference for snake_case, can you switch to it please?

q-wertz:
Thanks for your review and the comments.

> Regarding the license, do you know if the NIST data is open source and if they allow derivative works like this package? 
- On the [database website itself](https://physics.nist.gov/cuu/Constants/index.html) I didn't find any information regarding licensing
- On the NIST website itself I found  
  - https://www.nist.gov/copyrights-disclaimers 
    >  With the exception of material marked as copyrighted, information presented on NIST sites are considered public information and may be distributed or copied. Use of appropriate byline/photo/image credits is requested.
  - https://www.nist.gov/open/copyright-fair-use-and-licensing-statements-srd-data-software-and-technical-series-publications
- Sidenote: This database is also referenced by e.g. [Wikipedia](https://de.wikipedia.org/wiki/Physikalische_Konstante)

To my understanding it should be fine to use this data. I already have a citation in the `README.md` but would also add an additional comment to the Typst files that contain the data, just to be sure.

> And regarding the package itself, it seems to use snake_case, while kebab-case is the preferred naming convention in the Typst ecosystem. Unless you have a strong preference for snake_case, can you switch to it please?

Sure.

---

I would mark the Pull request as draft until I fixed those points. Might take until the weekend.

q-wertz:
I updated the package accordingly and added some more things.

Should the files (e.g. `atomic_and_nuclear.typ`) also be converted to camelCase?

elegaanz:
Because the module names are derived from their path unless you `import … as …`, yes please. Otherwise the module will have snake_case names. If you want to keep the file names as is, renaming the modules with `as` in `src/CODATA2022/codata2022.typ` is also an option.

q-wertz:
I decided to change the file names :)

elegaanz:
The names are still not kebab case. For example `atomicAndNuclear.typ` should be `atomic-and-nuclear.typ`. Also, a lot of files seems to be duplicated under slightly different names.

And the syntax to start a comment is `// ` not `\\ `. It's not too bad because the comments will just be interpreted as content and won't be exported, but I think it's still worth using the right syntax.

q-wertz:
:man_facepalming: so sorry. I somehow picked up camelCase instead of kebab-case and consistently did it all wrong :face_exhaling: :sweat_smile: 

Redundant files should be gone now. 
And also the comments are now real comments. :sweat_smile: :man_facepalming: 

So sorry for your extra work due to my lack of concentration.

Hope everything is fine now.

elegaanz:
No worries, thank you for fixing that :)

PR #5549, opened 2026-08-11
zero:0.7.0

Mc-Zen:
I am submitting
- [x] an update for a package

## Changelog

Features:
- The new `quan` function provides a new way of defining quantities and units.
- Zero now adds alt descriptions for all numbers and units for supported languages. 

Rounding:
- ⚠️ Breaking change: the rounding API was changed to a more comprehensible model and to better reflect common applications. Numbers with an uncertainty are by default (`round.follow-uncertainty: true`) rounded to the precision of the uncertainty − we call this _automatic rounding_. If no uncertainty is present or `round.follow-uncertainty: false`, _manually rounding_ is applied according to the options `mode` and `precision`. The `mode: "uncertainty"` has been removed.
- ⚠️ Breaking change: zeros are now trimmed after rounding and padding is independent. This leads to the following effect of change:
  - Before: `num(round: (precision: 2, pad: false)[0.299]` -> 0.30
  - Now: `num(round: (precision: 2, pad: false)[0.299]` -> 0.3 
- Fixed a bug where rounding to zero were for negative precision led to multiple zeros being displayed in the integer part.
- Fixed rounding to negative (integer) places.
- Fixed rounding of 0 to a positive number of significant figures: now no trailing zeros are displayed anymore.

Input:
- Added support for capital E in exponent notation.

Formatting:
- New option `num.trim-zeros` for trimming trailing zeros.
- New option `num.omit-zero-exponent`.
- Fixed: Automatic exponents with scientific or engineering notation are now omitted when the mantissa is 0 and no explicit exponent is given.
- Fixed prevention of line breaks and add more fine-grained line break control: after the ×, after the ± and before the unit.
- ⚠️ Breaking change: moved the parameter `breakable` from `set-unit` to `set-num` since it now also affects numbers.
- Fixed `num.use-sqrt` with inline fraction mode.
- Fixed formatting of exponents with custom decimal separators.
- Fixed an error with scientific mode with asymmetric uncertainties.
- Fixed a tiny detail when formatting the decimal separator.

API:
- Added queryable metadata to numbers and units.
- Moved the internal `num-state` to the impl module.

### DIFF
--- a/packages/preview/zero/0.7.0/LICENSE
+++ b/packages/preview/zero/0.7.0/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024-2026 Mc-Zen
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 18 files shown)


Mc-Zen:
The warnings are expected and relate to the type annotations and simplified examples

elegaanz:
Thank you :)

PR #5557, opened 2026-08-11
pku-thesis-pass:0.3.0

chuxinyuan:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Complete architecture overhaul with SSOT style system, 9 theorem environments, subfigures, equation blocks, notation pages and CJK word counting.

Changes:
- Config return dict now uses key-based access instead of positional destructuring, and config.typ is split into cli.typ + resolve.typ + builder.typ.
- New format/utils/style.typ serves as the single source of truth for all style values across the template.
- Font system split into three platform-specific schemes (windows/macos/linux) with a font-aware faux-bold strategy.
- Added 9 theorem environments, subfigures, equation blocks, spine page, achievement page, notation page, and CJK word counting.
- Added LaTeX cross-reference compatibility and automatic PDF metadata generation.
- FAQ expanded to 12 entries with a 20-module import reference table and equation comparison guide.

Checklist:
- Version in typst.toml bumped to 0.3.0
- All @preview/pku-thesis-pass:0.2.0 references updated to 0.3.0
- Template compiles in both normal and blind mode


elegaanz:
Thank you for this update!

PR #5562, opened 2026-08-11
oasis-align:0.4.0

jdpieck:
I am submitting
- [ ] a new package
- [x] an update for a package

## `oasis-align` v0.4.0 - 2026/08/04
- added `padding` to all functions
    - helps when you don't want the content spanning the full page
- added `figure` variant of `oasis-align()`
    - for when you want to align the images but not the caption text
- added content ID system to automatically switch from `oasis-align()` to `oasis-align-images()` and `oasis-align-figures()` logic based on input content type
    - increased performance of image-image and figure-figure alignment 
    - added `override` parameter to override this behavior
- major refactoring of codebases
    - added help functions under `utils.typ`
    - separated `image` and `figure` variants of `oasis-align()` into separate files
- `README.md` now includes extra examples of use cases along with tips


jdpieck:
It seems that the `package-check` job panicked on `src/action.rs:23:10` due to a GitHub API authentication error. Let me know if I should make any modifications to the package itself!

saecki:
Thanks!

PR #5566, opened 2026-08-12
easy-wi-hwr:0.1.3

lultoni:
I am submitting
- [ ] a new package
- [x] an update for a package

**Changes in 0.1.3:**
- feat: `include-appendix-in-lists` parameter — optionally show appendix entries in ToC, LoF, and LoT
- fix: signature layout uses a two-row grid so the image appears above the line (not behind it)
- fix: 1.5-line spacing corrected to match Word behaviour (`leading: 0.8em`)
- fix: glossary entries with `long == short` no longer produce redundant "Foo (Foo)" on first use
- test: additional happy-path test scenarios for abbreviation expansion, glossary, and appendix-in-lists


saecki:
Thanks!

PR #5567, opened 2026-08-13
hanko-kitei:0.1.0

liquidcatmofu:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A template and library for Japanese workplace regulations (就業規則) and
similar internal rulebooks. Japanese regulations have a fixed hierarchy — 章 (chapter),
節 (section), 条 (article), 項 (paragraph), 号 (item) — with display conventions that are
tedious and error-prone to maintain by hand: full-width digits below ten, ASCII from ten
up, parenthesised item numbers, and inter-character spacing on two-character titles.

The package derives every one of those numbers from source order and renders them through
Typst's own constructs, so authors write standard headings, `+` enumerations, term lists
and labels, and never type a number. Cross-references use ordinary labels, so `@目的`
resolves to `第１条` and keeps working when an article is inserted above it. A
`line-spacing` preset keeps the vertical rhythm uniform, which matters because Japanese
regulations are read as dense running text.

  - Explanation:
    判子 (*hanko*) is the carved personal seal that Japanese companies and individuals
    stamp onto documents to approve them; 規程 (*kitei*) means "regulations". The name is
    a neologism pairing a non-descriptive term — a hanko says nothing about typesetting,
    numbering or documents being generated — with the descriptive part, following the
    "unique part plus descriptive part" guidance for templates. The canonical name for
    this functionality would have been something like `japanese-regulation`, which I
    deliberately avoided.

Tested against Typst 0.15.1 by placing the package in the local cache and running both
paths a user would take: `typst init @preview/hanko-kitei:0.1.0` scaffolds and compiles,
and `#import "@preview/hanko-kitei:0.1.0"` compiles as a library. The two README images
live in `docs/`, which `exclude` keeps out of the bundle; the thumbnail is excluded
automatically. MIT-0 is used so the template directory carries no attribution requirement
after a user edits it.


liquidcatmofu:
`compiler = "0.14.0"` is a floor I verified rather than guessed. I rendered the
same document with 0.14.0, 0.14.2, 0.15.0 and 0.15.1 and compared every glyph
box in the output: it is identical across all four. 0.13.1 compiles but spaces
table-of-contents entries differently, so the floor is not set lower.

elegaanz:
Thank you for taking time to verify the minimum version, I appreciate the effort. And thank you for the package in general!

PR #5583, opened 2026-08-14
polycv:0.1.1

bbinet:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: polycv 0.1.1 — a data-driven two-column CV package for Typst.

- [X] have chosen a license and added a `LICENSE` file


packages/preview/polycv/0.1.1/README.md
```diff
@@ -0,0 +1,332 @@
...
+```
+
+Prefer TOML? Edit `cv.toml` instead (same fields) and add `--input fmt=toml` when you compile. Edit `letter.yml` similarly for your cover letter. Run `make yaml-reference` to print every available field with its type and a one-line description.
+
+### 3. Build
+
+```sh
+# One command, incremental - rebuilds only what changed:
+make            # compiles every cv*.yml / letter*.yml to a PDF
+make watch      # live-preview while you edit (WATCH=<file> for a single one)
```
elegaanz: Users won't have access to the Makefile. I would advise removing that part in favor of the `typst` commands below.

packages/preview/polycv/0.1.1/template/validate.py
elegaanz, line 1: We prefer not to encourage the use of external tools in packages if possible. Shipping this script in every project created from this template will encourage people run arbitrary code on their machine, which Typst is designed to avoid. In the web app it will also be quite confusing, since there will be no way to run this file.

If you want to provide schema validation, I would advise doing it in Typst directly. You can try to use [`jsonschemeyst`](https://typst.app/universe/package/jsonschemeyst) to help you with that, since you already have a JSON Schema. Otherwise [`valkyrie`](https://typst.app/universe/package/valkyrie) or good old custom logic can work too.

In any case, can you please delete this file?

packages/preview/polycv/0.1.1/template/_schema.typ
```diff
@@ -0,0 +1,44 @@
...
+// document as metadata so the Makefile can pull it out as JSON:
+//   typst eval 'query(<polycv-data>).first().value' --in _schema.typ \
+//     --input data=<file> --input fmt=<yaml|toml> --format json
+// It mirrors the loading in cv.typ / letter.typ but does no layout, so it
+// never fails on data that merely violates the schema.
+#let fmt = sys.inputs.at("fmt", default: "yaml")
+#let data-file = sys.inputs.at("data")
+#let load-data(f) = if fmt == "yaml" { yaml(f) } else { toml(f) }
+
+#let deep-merge(base, over) = {
```
elegaanz: Some of these functions seems to be duplicated among different files. I would advise to import them all from here, or even to move that file to the `src/` folder, as I don't expect users will need to edit these functions.

packages/preview/polycv/0.1.1/template/cv.typ
```diff
@@ -0,0 +1,153 @@
...
+#let ats-split = input-bool("ats-split")
+#let entry-inline-meta = input-bool("entry-inline-meta")
+#let show-timeline = input-bool("show-timeline", default: true)
+#let locale = input-str("locale", default: "en")
+
+// Optional section ordering from meta (arrays); omitted keys use cv() defaults.
+// 0 = auto (one badge per line)
+#let keywords-lines = int(input-str("keywords-lines", default: "0"))
+
+#let locale-args = if locale == "fr" {
```
elegaanz: Same for the translation logic, I feel like it could be part of the core of the package and live in `src/` (while keeping the option to override it from the user side for other languages, of course).

packages/preview/polycv/0.1.1/template/gen-reference.py
elegaanz, line 1: Same for this file: I think you can easily remove the need for it by generating that text once and putting it in a `.txt` or `.md` (or `.pdf` if you want something more visual) file, since the schema won't change.

packages/preview/polycv/0.1.1/typst.toml
```diff
@@ -0,0 +1,34 @@
...
+name = "polycv"
+version = "0.1.1"
+entrypoint = "src/lib.typ"
+authors = ["Bruno Binet"]
+license = "MIT"
+description = "A data-driven, not-a-boring CV and cover-letter template"
+repository = "https://github.com/bbinet/polycv"
+keywords = ["cv", "resume", "cover-letter", "curriculum-vitae", "two-column", "letter"]
+categories = ["cv"]
+disciplines = []
```
elegaanz: You can remove that field if it is empty.

elegaanz (commented):
Thank you for building this template. I have some remarks regarding its structure.

bbinet:
Thanks for the thorough review! I've pushed changes addressing all points:

1. typst.toml: removed the empty disciplines field.
2. README: the Build section now leads with typst compile; the Makefile is presented as an optional local convenience only.
3. validate.py: deleted. Validation now happens at compile time in Typst via jsonschemeyst against a bundled src/schema.json, so an invalid field stops the build naming its path (e.g. /meta/locale) and it should work in the web app.
4. _schema.typ / duplicated functions: the load + inherit: merge logic now lives once in src/data.typ (load-cv-data); the entrypoints just call it.
5. Translation logic: moved into src/cv.typ, applied from locale and still overridable via section-titles / month-names.
6. gen-reference.py: removed from the template; replaced by a static FIELDS.md (regenerated from the schema on my side at release time).

Let me know if anything else needs adjusting!

elegaanz:
Looks good, thank you.

PR #5587, opened 2026-08-15
intl-math:0.1.0

13Stokes31:
Adds **intl-math** 0.1.0 — math operators that print in the language of the
document.

Mathematical abbreviations are not universal: in Spanish, sine is `sen`, tangent
`tg`, limit `lím` and the greatest common divisor `mcd`. Typst ships only the
English ones, so a Spanish document comes out in English notation unless the
operators are redefined by hand in every project.

The point of this package is that **the identifier does not change, only the
printed word does**. You keep writing `sin(x)` — the name Typst documents — and
`text.lang` decides what is typeset, so adopting it does not mean rewriting any
content. It ships English and Spanish, and any other language can be added
**without forking**, by passing it to `intl(extra: ...)`; the fallback cascade is
document language → English → the key itself, so an incomplete table or an
unknown language degrades to English instead of breaking the document.

It also covers a few symbols Typst does not ship and that teaching material needs
(`rank`, `dom`, `proj`, `mvt`, and the `if` of piecewise functions).

There is some overlap with
[`alterlang`](https://typst.app/universe/package/alterlang/), which translates the
standard operators by importing the translated symbol (`sen`) instead. This one
keeps the original identifiers and allows adding a language without forking. The
underlying gap is acknowledged upstream in typst/typst#3238 and typst/typst#7159.

- Repository: https://github.com/13Stokes31/intl-math
- License: MIT
- No dependencies. Typst `>= 0.14.0`.


elegaanz:
Thank you.

PR #5592, opened 2026-08-15
ficons:0.1.0

BalderHolst:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This package adds the open source feather icon set as a dictionary to be used in your documents.

  - Explanation:
    "Feather icons" is the name of the icon set. It includes "icons" so it should be self explainable. The name is simply the source of the icons.

The package is checked on v0.15.1. I am not using any wild features, just building a dictionary of icons, so it should be quite compatible (hopefully).


HackYardo:
I'm not a member of Typst team, but it is clear something wrong in ur pr.
There are **three** different repos for **one** Typst package or template:
1. host repo, such as github repo
2. minimum required script to download and run from `import`
3. the 2 + minimum readme document to show at Typst Universe

So, the .github, \_test\_, etc, **not only be excluded in typst.toml,
but also not to commit to typst/packages.**

BalderHolst:
Thank you. I will look into it:)

BalderHolst:
Sorry about that. I had comitted way too many files. The unneeded files are now removed from git. Let me know if there is anything else:)

packages/preview/feather-icons/0.1.0/typst.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "feather-icons"
```
elegaanz: This name is the canonical name for that package, which is [forbidden by our rules](https://github.com/typst/packages/blob/0488b38bb7fdfbce81981803ebbcf5612f8919f8/docs/manifest.md#naming-rules). The reasoning behind that rule is that two different implementation of the same package (or really similar variants of it) should be able to co-exist without any of the two having an unfair advantage over the other because of a more standard or more official-looking name. Can you please pick another name? Feel free to ask before renaming if you want to make sure the new name matches our guidelines.

elegaanz (commented):
Thank you for making this package. I have one remark to make regarding the name. Also, for icon sets, it is generally preferred to have the user install a font file and have the package only provide nice human-readable aliases to glyphs, instead of shipping all these SVG files. Could that approach work here?

BalderHolst:
Hello! Thank you for taking a look at my package:)

I would like to keep the icons in SVG format if possible, as I like the flexibility it gives; The `make-icons` function edits the SVG to create tweaked icon sets.

I do see that it would be best to keep the package size down, so I have packed all the icons into a [single file](https://github.com/BalderHolst/typst-packages/blob/main/packages/preview/feather-icons/0.1.0/feather-icons-data.typ) inspired by the [`booticons` package](https://github.com/typst/packages/blob/main/packages/preview/booticons/0.0.1/bs-icons.typ). This let me store the common SVG header separately, which saved quite a few bytes. The package is now only 72 Kilobytes.

Regarding the name, what do you think about these:
- `ficons`
- `fticons`
- `fthr-icons`

I like `ficons` the best as it is shortest, but maybe that is also too general? I see there is a package called "ccicons".

elegaanz:
Thank you for the size-reduction efforts :) I would say `ficons` is fine, it's some kind of neologism, not an actual word, so not too general.

BalderHolst:
I have renamed the package to "ficons":)

I also realized that it would be better to ship the [documentation](https://github.com/BalderHolst/typst-packages/blob/main/packages/preview/ficons/0.1.0/docs.pdf) and [example svg](https://github.com/BalderHolst/typst-packages/blob/main/packages/preview/ficons/0.1.0/example.svg) to pin them to this release.

These files are `exclude`d in "typst.toml".

elegaanz:
Perfect, thank you :)

PR #5595, opened 2026-08-16
clean-othaw:0.3.7

ruebe5w:
…declaration of authorship

I am submitting
- [ ] a new package
- [x] an update for a package

Description: add signature option in template, add KI-Statement in declaration of authorship


elegaanz:
Thank you.

PR #5596, opened 2026-08-16
datehog:0.1.1

zral0kh:
I am submitting
- [x] a new package
- [ ] an update for a package

**Datehog** enables parsing of datestrings and adds epoch arithmetics, adding a week or similar.
Dates are parsed as moments, dicts of various time fields, against the epoch root 1970-01-01 and can be extracted as milliseconds, iso strings and typst datetime objects and more.

It is basically on par with js v8, but deviates on a small number of instances where it follows pythons datetime utils. In other cases it follows v8 and not python's datetime.

For more details and the whole API, please see the [README.md](https://github.com/zral0kh/datehog/blob/v0.1.1/README.md).

The reason why this is not v0.1.0 but a patch is that i forgot the changelog which the readme links to and the repository field. could have updated inplace idk.

  - Explanation: The name just comes from that this package can feed on all kinds of date strings and still works fine.
    
### DIFF
--- a/packages/preview/datehog/0.1.1/CHANGELOG.md
+++ b/packages/preview/datehog/0.1.1/CHANGELOG.md
@@ -1,0 +1,38 @@
+# Changelog of all Changes to Datehog
+
+## v0.1.1
+
+No functional changes. The `v0.1.0` was prematurely created without this changelog and the release pipeline.
+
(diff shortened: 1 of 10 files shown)


elegaanz:
Thank you.

PR #5597, opened 2026-08-16
clean-dhbw:0.6.0

roland-KA:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:

- Template specific table formatting now available by using the new function dhbw-table; this avoids conflicts with e.g. the codelst-package
- Formatting of captions (especially multi-line captions) improved
- Handling of long chapter headlines within the header improved
- Some minor improvements within the README.md-docs


roland-KA:
As far as I can see, the warnings about links to the default branch can be ignored. And I don't understand the last warning about importing an older version of the template as it imports the new Version 0.6.0.

elegaanz:
Regarding the links, I agree that the ones linking to `.md` files can stay as is. For the ones linking to example PDF or Typst files, it would be better to pin them to a specific commit or git tag: otherwise they will point to an invalid version if you modify them while developing the next version of the package. If that makes sense, can you change the links please?

The outdated import warning is on the `hydra` import, which is not on the very last version, but it can also safely be ignored. It is more here as a reminder that updating can fix issues, in case there are any with that package.

roland-KA:
@elegaanz I've changed the links of the questionable files to local relative paths as recommended.

packages/preview/clean-dhbw/0.6.0/README.md
```diff
@@ -0,0 +1,281 @@
+# Clean DHBW
+
+A (modernized) [Typst](https://typst.app/) template for DHBW documents like Bachelor theses, "Studienarbeiten", project documentation etc. It is the official Typst template for Computer Science at DHBW Karlsruhe.
+
+You can see an example of how the template looks in this [PDF file](template/main.pdf).
```
elegaanz: This doesn't really work either, because `template/main.pdf` doesn't exist in this copy of the package. And it wouldn't make sense to ship it as part of the template. Can you copy it somewhere else (for example the root of the package) and update the link to point to that, please?

roland-KA: Oh, I've overlooked that. I've changed the link to an absolute URL with version number pointing to my repo.

To copy it somewhere else, as you suggest, doesn't really make sense. It is an inherent part of the documentation of the template. The problem is rather in the rules of Typst Universe not allowing such files (presumably to save storage space?).

elegaanz: We do try to save space by not committing to many files to the registry, but that's not really the issue here. The problem is more that the `template` directory is what gets copied to the user project when they start from your template, and we don't want to pre-populate it with a PDF file, especially in the Typst web app where the output is never written to a file of the project, but generated on demand when you download it.

In any case, the latest version works, thank you for fixing it!

PR #5601, opened 2026-08-17
iarticle:0.1.1

hideo67:
I am submitting
- [ ] a new package
- [x] an update for a package

Description:  A LaTeX article/report inspired template with internationalization support. Keywords such as "Chapter"
may be localized by resource files. Currently "en" and "ja" is supported. Other locales can be added with minimal
modification to the main lib.typ file(The lib.typ file contains a list of supported locales).

This update has the following changes:
- a bug is fixed for captions on tables. Captions for tables should appear above the table but it was shown below. 
- spaces before and after the caption and the figure block has been increased for readability.


elegaanz:
Thank you.

PR #5604, opened 2026-08-17
payqr-swiss:0.5.0

philippdrebes:
I am submitting
- [ ] a new package
- [x] an update for a package

### Added

- **Automatic language detection**: `language` now accepts `auto` and is the new default, resolving the bill language from the document's `text.lang`
- Alt text on the QR code and all layout icons, improving accessibility of the generated output, as well as PDF/A compliance

### Changed

- **Default language is no longer hard-coded to German.** With the new `auto` default, a document that sets `#set text(lang: "fr")` will render a French QR bill where it previously rendered German. Pass `language: "de"` explicitly to keep the old behavior.
- The fallback for an unrecognized language is now German instead of English


elegaanz:
Thank you.

PR #10, opened 2023-07-01
splash:0.3.0

kaarmu:
I am submitting
- [x] a new package
- [ ] an update for a package

I have read and followed the submission guidelines and, in particular, I
- [x] selected a name in conformance with the guidelines
- [x] added a `typst.toml` file with all required keys
- [x] added a `README.md` with documentation for my package
- [x] have chosen a license and added a `LICENSE` file or linked one in my `README.md`
- [x] ensured that my submission does not infringe upon the rights of a third party
- [x] tested my package locally on my system and it worked
- [x] named this PR as `name:version` of the submitted package
- [x] agree that my package will not be removed without good reason

Description: This library provides different color palettes with human-readable names in Typst dictionaries.

PR #5606, opened 2026-08-17
blockst:0.3.0

Loewe1000:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: `blockst` renders Scratch-style blocks for worksheets and teaching material. Typst passes the block text to a bundled WASM plugin, which parses it and returns SVG.

What changed in 0.3.0:

- **Right-to-left rendering** for Arabic, Hebrew and Persian. Each locale declares its own direction in its TOML, and the renderer mirrors the notch, the hat dome, the C-block mouth and its indented body, the loop arrow, the pen badge, the `define` hat and the order of labels. The turn-direction icons are never mirrored, since their direction is what the sprite must do.
- **A new Arabic locale**, built from the official scratch-l10n data. Short vowels are folded before matching, so `كرِّر` and `كرر` find the same block.
- **A `grayscale` theme** giving each Scratch category a distinct grey, for worksheets that get photocopied. The existing `print` theme is unchanged.
- **A `font:` parameter** on `scratch()` and `blockst()`.
- The French localisation is completed (`définir`, `appeler`, the `tourner à droite/gauche` aliases, French pen-effect values). The older aliases still parse, so documents written against 0.2.x keep working.
- Fixes: block labels no longer fall back to the surrounding font inside ```` ```scratch ```` blocks (they used to come out monospace with the labels overflowing); the spec hash mangled multi-byte characters, so every non-Latin locale fell back to the "unknown" block shape; C-block headers were drawn twice.

`compiler` is raised to `0.15.0`, which is the version 0.3.0 is tested against.

Upstream: https://github.com/Loewe1000/blockst — the work in this release closed issues #3, #6, #7 and #10 there.


elegaanz:
Thank you.

PR #5607, opened 2026-08-17
conic-toan:0.3.4

Thutran1891:
Update of `conic-toan` from 0.3.3 to 0.3.4. No breaking API changes; the minimum compiler version stays at 0.13.0. All existing figures and layouts are unchanged — every new parameter defaults to the previous behaviour.

### New

- **New circle helpers.**
  - `giao-duong-thang-duong-tron((A, B), (O, r))` returns the intersection of a line and a circle as an array ordered along `A -> B`: `()` when they miss, `(P,)` when tangent, `(P1, P2)` when secant. It works by projecting `O` onto the line, so it is exact rather than sampled.
  - `tiep-tuyen-tai-diem(ctx, O, r, M)` draws the tangent through `M` perpendicular to the radius `OM` (the direction is taken from the actual `OM`, so `M` need not sit precisely on the circle). `dai` is the half-length of the segment (auto = `1.4·r`, in math coordinates).
  - `dung-diem(A, B, goc, r)` returns the point `M` such that ray `AM` is ray `AB` rotated about `A` by the signed angle `goc` (degrees, positive counter-clockwise) with `AM = r`.
  - `nhan-goc(ctx, ..items)` labels several points at once, choosing each label's side by a signed angle instead of a named direction. Each item is a tuple `(P, text, angle)`, optionally extended with a radius and/or colour in either order.
- **`dut:` for `duong-tron-ngoai-tiep` and `duong-tron-noi-tiep`** — draws the circle with a dashed stroke. Defaults to `false`, so existing figures are unchanged.
- **`an` / `hien-khoang` — exit animations for slides.** The mirror image of `lo`: `#an(n)[…]` shows content from the start and hides it from step `n` onward (collapsing the following material by default, `giu-cho: true` to keep its space). `#hien-khoang(tu, den)[…]` shows content only during steps `[tu, den)`. Pairing `#an(n)[old] #lo(n)[new]` replaces an element in place on the same step. In the printed (non-beamer) profile everything is shown, as with `lo`/`chi`.
- **`fig-pos: "center"` — a figure centred between the prompt and the options.** The four existing placements are `"right"`/`"left"` (figure in a side column with text wrapping), `"top"` (above the whole prompt) and `"bottom"` (below all four options). None of these suit a wide figure such as a variation table, a statistical chart or a large graph. `"center"` puts the figure on its own centred line, between the prompt and the options / sub-questions. Available for `#tn`, `#ds`, `#tln` (`fig-pos:`), for figures attached to a solution (`fig-giai-pos:`) and for `#voi-hinh(..., vi-tri: "center")`. An invalid `fig-pos` now raises an error listing the valid values instead of silently falling back to `"right"`.
- **`fig-pos:` / `fig-width:` on all eight question types.** `#vd`, `#tl`, `#hd`, `#lt`, `#vdtt` (and `#cau-tl`/`#cau-hd`/`#cau-lt`/`#cau-vdtt`) previously accepted only `fig-giai-pos:` for a figure attached to the solution; a figure on the prompt was always forced into the right column, with no parameter to change it, even though the manual described it as "every question type". These five now take the same `fig:` / `fig-pos:` / `fig-width:` as `#tn`/`#ds`/`#tln`. Omitting them keeps the default `"right"`, so existing questions keep their layout.
- **SGK activity blocks: `kham-pha` / `trai-nghiem` / `thao-luan`.** The "Khám phá."/"Trải nghiệm."/"Thảo luận." activities of the Vietnamese textbooks are one-off, **not numbered** — unlike `hd`/`lt`/`vdtt` — so they carry no counter. Each has its own colour, plus an inline builder (`cau-kham-pha` / `cau-trai-nghiem` / `cau-thao-luan`) for use as a step inside a `#slide` (the top-level form creates its own slide and so cannot be nested).

### Changed

- **`bpt` labels rotate along the line.** `bpt(...)` gains `nghieng-ten: true` (the default), which rotates the label to follow the line and hugs it along the normal; `huong-ten` then only selects the side with `"above"`/`"below"`, and `cach-ten: 3pt` tunes the gap. `nghieng-ten: false` restores the previous horizontal label, which still accepts any direction. Each `bpt` also takes its own `goc-gach:` / `buoc-gach:` instead of sharing the values from `mien-nghiem`.
- **`dat-lai-cau` split per category.** `dat-lai-cau()` now resets only the "Câu" counters (`tn`/`ds`/`tln`/`tl`); the new `dat-lai-cau-vd` / `-hd` / `-lt` / `-vdtt` each reset a single counter, and `dat-lai-cau-tat-ca()` keeps the old behaviour of resetting all five groups. Exam files calling `#dat-lai-cau()` are unaffected, as they only use the "Câu" counters.

- **`so-do-cay` probability labels sit below descending branches.** An edge's probability label was always placed above the line; on a branch going *down* to a lower child this pushed the label into the fork. Labels on descending branches now go below their edge (ascending and level branches are unchanged), mirroring the top branches for a cleaner tree.

### Fixed

- **Implicit `ctx` no longer panics inside `measure`.** Drawing helpers may be called without an explicit `ctx` — the `_voi-ctx` wrapper picks up the current `#hinh` frame from a `state` stack. But content built inside `measure(...)` is never laid out, so that state reads back empty and the wrapper used to `panic`. This aborted the whole compile in the beamer profile (its automatic slide-breaking measures a trial slide) and in `cot: auto` for multiple-choice options containing a figure. An empty stack inside a measure is now skipped silently — the measured size is unchanged, since `#hinh` is a fixed-size box whose strokes are all `place`d — while a helper genuinely called outside any `#hinh` leaves a small red marker instead of aborting. Printed output is unchanged.


elegaanz:
Thank you.

Thutran1891:
Thank you very much.

PR #5613, opened 2026-08-18
pointless-size:0.1.3

YDX-2147483647:
I am submitting
- [x] an update for a package

Description:
This is a docs-only update.
Previously, the images in the README used github.com/user-attachments, which is slow in China (this package's main targeting area).
This update replaces them with github.io.


YDX-2147483647:
> Possibly inadequate alternative description for image: `Changelog`. Please add a short description to make this image more accessible.

Well, I think _Changelog_ is sufficient in this case.

elegaanz:
Yeah it makes sense here, this lint is a bit zealous.

Thank you for the fix :)

PR #5619, opened 2026-08-19
beautiframe:0.4.5

nathan-ed:
I am submitting
- [x] an update for a package

Description: adds trous (`#trou` / `#trou-inline`), fill-in space that carries its content — reserved, correctly sized blank space in the student build and the content itself in the instructor build, from one source. Also adds header layouts (`title-first`, `title-abbrev`, `title-only`, `prefix`) with label abbreviations, `env-colors` and `label-color` so every style can follow the per-environment colours, `preset-english()` and `beautiframe-reset-config()` to come back from a preset or a style, and perceptual background tints. Fixes the proof label and body sitting on different baselines.


elegaanz:
Thank you.

PR #5621, opened 2026-08-19
taskize:0.2.9

nathan-ed:
I am submitting
- [x] an update for a package

Description: fixes the label of a task item that opens inline and continues with a block (a formula, a line break, then a figure). Such an item fell back to the top-aligned rendering, so with a display fraction on the first line the number sat level with the numerator instead of on the formula's baseline. The item is now split at the end of its first line and the label shares that baseline.


elegaanz:
Thank you.

PR #5622, opened 2026-08-19
exercise-bank:0.6.3

nathan-ed:
I am submitting
- [x] an update for a package

Description: fixes the sentence an exercise places before its `#tasks` call running under the QR code when `qr-position: "tasks"` is used in a narrow column. That text is not part of the taskize grid, so nothing kept it clear of the overlay; it is now set at a reduced width for as long as the overlay lasts, and the wrap zone published for the grid shrinks accordingly.


elegaanz:
Thank you.

PR #5623, opened 2026-08-19
ctz-euclide:0.2.1

nathan-ed:
I am submitting
- [x] an update for a package

Description: label improvements from user reports. Positions accept numeric offsets ("above+1.2 left+0.5") for fine-tuning, `ctz-draw-labels` no longer needs the point list when positions are given (the short form used to draw nothing), and trailing digits in a point name are set as a subscript, so I1 renders as I₁.


nathan-ed:
Superseded by the 0.3.0 submission: the default point marker changed, which redraws existing figures, so this belongs in a minor release rather than a patch.

PR #5626, opened 2026-08-19
ctz-euclide:0.3.0

nathan-ed:
I am submitting
- [x] an update for a package

Description: the default point marker is now a small circle rather than a cross, which is what tkz-euclide and textbooks draw and stays readable where several lines meet at the point; every other shape remains one call away. Adds polar label placement (`"135deg"`, `"200deg+0.6"`, or `(angle: 135deg, dist: 0.6)`), a `ctz-def-triangle-center(kind, name, "A,B,C")` entry point over the eleven existing centre functions, `angle-type: "complementary"`, subscripts for trailing digits in point names, and fixes an arc named by two points sweeping the long way round when its end lay past the half turn.


elegaanz:
Thank you.

PR #5627, opened 2026-08-19
functable:0.2.0

nathan-ed:
I am submitting
- [x] an update for a package

Description: the vertical rule at a zero is dotted by default, since a solid one read like the double bar of a valeur interdite in a factor row that has no zero there (`zero-line: "solid"` restores it). Adds `x-label` so a sign table can be in `t` or `theta`, and `second-variation`, which draws the second block's bottom row as a variation row rather than a convexity row, so one table can carry the whole second-derivative reading: sign of f'' , variations of f', sign of f', variations of f.


elegaanz:
Thanks once again :)

PR #5628, opened 2026-08-19
transl:0.2.1

### LINKED ISSUE #11: `set-case` panics on translations starting with a multi-byte UTF-8 character
**Bug description:**
`transl` panics with `string index 1 is not a character boundary` when the translated value starts with a multi-byte UTF-8 character (e.g. `"í"`, `"é"`, `"ñ"`, `"à"`) and the requested key triggers case adjustment.

The root cause is in `src/utils.typ`, in `set-case`:

```typ
#let set-case(result, expr) = {
  let first = expr.first()

  if first == lower(first) {lower(result.first()) + result.slice(1)}
  else if expr == upper(expr) {upper(result)}
  else if first == upper(first) {upper(result.first()) + result.slice(1)}
  else {result}
}
```

`string.slice` in Typst is byte-indexed and requires the index to fall on a character boundary. When `result` starts with a multi-byte character (e.g. `"índice"`, where `í` occupies bytes 0–1), `result.slice(1)` panics. The first and third branches are both affected; only the all-uppercase branch (`upper(result)`) is safe.

**To reproduce:**

1. Create `lang.yaml`:
   ```yaml
   l10n: std
   en:
     contents: "contents"
   es:
     contents: "índice"
   ```
2. Create `main.typ`:
   ```typ
   #import "@preview/transl:0.2.0": transl
   #set text(lang: "es")
   #transl(data: yaml("lang.yaml"))

   #transl("Contents")
   ```
3. Compile with `typst compile main.typ`. Compilation fails with:
   ```
   error: string index 1 is not a character boundary
     ┌─ @preview/transl:0.2.0/src/utils.typ:7:57
     │
   7 │   else if first == upper(first) {upper(result.first()) + result.slice(1)}
     │                                                          ^^^^^^^^^^^^^^^
   ```

**Expected behavior:**
`#transl("Contents")` should return `"Índice"` (and `#transl("contents")` should return `"índice"`), regardless of whether the translated value begins with an ASCII or multi-byte UTF-8 character.

A safe replacement for the slicing logic uses the byte length of the first grapheme:

```typ
#let _head-tail(s, transform) = transform(s.first()) + s.slice(s.first().len())

#let set-case(result, expr) = {
  let first = expr.first()

  if first == lower(first) { _head-tail(result, lower) }
  else if expr == upper(expr) { upper(result) }
  else if first == upper(first) { _head-tail(result, upper) }
  else { result }
}
```

`s.first().len()` returns the byte length of the first character, so the slice always lands on a valid character boundary.

**Additional info:**
 - Platform: CLI
 - Typst version: 0.14.2 (b33de9de)
 - transl version: 0.2.0

mayconfmelo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: This patch internally renames the plugin responsible for fetching Fluent data and fixes an error that prevents `#transl` from retrieving translations starting with multi-byte characters.

- Internal re-design of Fluent WASM plugin
- Fixed [#11](https://github.com/mayconfmelo/transl/issues/11): panic when obtained translated text starts with multi-byte character ("é", "í", "ñ", etc.)


elegaanz:
Thank you.

PR #5630, opened 2026-08-19
theoframe:0.3.6

hexiongwu1995:
I am submitting
- [x] an update for a package

Update to version 0.3.6
downgrade the typst compiler to version 0.15.0 for compatibility


hexiongwu1995:
The tinymist version 0.15.2 in VS Code  does not support importing theoframe 0.3.5. It hints that what I installed is typst version 0.15.0 instead of 0.15.1 which is indeed what I installed in my computer.

So I downgrade the typst compiler to 0.15.0 in typst.toml for compatibility.

elegaanz:
Thank you.

PR #5633, opened 2026-08-19
decasify:0.11.4

alerque:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Added a naive Unicode style that ignores all grammar features and just brute forces a title cased string, but that follows the contextual language (so Turkish is handled properly) and alllows specific word overrides configured the same way as for other style guides.


elegaanz:
Thank you!

PR #5635, opened 2026-08-20
conic-toan:0.3.5

Thutran1891:
Update `conic-toan` to 0.3.5.

This release adds sphere and polyhedron support to the ray-casting solid
renderer, and fixes figure handling on multi-screen slides.

### Added

- `khoi-cau` — a sphere solid for the `mat-cong` renderer, with optional
  equator and meridians. Hidden-line removal comes from the same ray-casting
  pass used by cones and cylinders, so a sphere correctly occludes — and is
  occluded by — the other solids in the same call.
- `khoi-da-dien` — polyhedra can now be fed to the ray-casting renderer, which
  makes cross-solid hidden-line removal work between polyhedra and curved
  solids (a cone piercing a box, a sphere inside a cube). The existing
  `da-dien` back-face renderer is untouched and still available; for a
  standalone polyhedron the two agree edge for edge.
- `hinh-moi-man(bat)` and the per-question `fig-giai-moi-man:` switch. When a
  solution runs longer than one slide, the accompanying figure is now repeated
  on every continuation slide instead of appearing only on the first one.
  Slide-splitting measurement accounts for the figure, so continuation slides
  no longer overflow.
- `fig-giai:` also accepts an array of figures. Figure *i* belongs to logical
  screen *i* — the screens separated by `#sang-man` — so a question with parts
  a) and b) can show a different figure for each part. Screens produced by
  automatic splitting inherit the figure of the part they belong to. On the
  paper profiles, where there are no screens, the solution is split at the same
  boundaries and each part is paired with its own figure, so a figure always
  sits next to the text it illustrates.
- `khoi-tu-dien-deu` takes a `pha:` argument — the rotation of the base
  triangle.

### Changed

- The default shape of `khoi-tu-dien-deu` changed from `pha: 210deg` to
  `pha: 270deg`. At 210° an oblique camera with `k · cos(goc)` near
  `1/sqrt(3)` projects two base vertices onto the same vertical line,
  collapsing one face into a single stroke; the new default keeps all four
  faces open. `pha: 210deg` (or `pha: auto`) restores the previous shape.

Documents that inserted a figure by hand into a `#sang-man` screen will now
show it twice; remove the manual copy, or set `fig-giai-moi-man: false` on that
question. Questions without `fig-giai:` are unaffected, and the two paper
profiles keep their previous layout.
Regular tetrahedra drawn with `khoi-tu-dien-deu` change orientation and, with
`h: auto`, become slightly taller; pass `pha: 210deg` to keep the old figure.


elegaanz:
Thank you.

Thutran1891:
Thank you very much.

PR #5638, opened 2026-08-20
classic-thesis-uct:0.1.2

james-hepworth:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Updates `classic-thesis-uct` from 0.1.1 to 0.1.2. Adds configurable font support: documents can now select between the default EB Garamond 12 and Adobe Garamond Pro via `meta.font-choice`. Small-caps and heading text now respect the chosen typeface through a shared state. Also introduces a `page-body-height` constant and updates `.gitignore`.


elegaanz:
Thank you.

PR #5649, opened 2026-08-21
mastermind:0.2.0

manuel3053:
I am submitting
- [ ] a new package
- [x] an update for a package

- BREAKING CHANGE: the creation of classes and relationships requires less boilerplate but it's not compatible with older versions
- fix: association and dependancy now uses straight instead of triangle as mark
- Introduce column, row and group functions to simplify the arrangement of classes
- the sides where the relationships attaches to can be overridden
- overhaul of built in themes


elegaanz:
Is this PR ready for review?

packages/preview/mastermind/0.2.0/README.md
```diff
@@ -0,0 +1,205 @@
...
+*You can also build your own custom theme:*
+
+<p align="center">
+  <img src="examples/custom-theme.png" alt="UML diagram with a custom theme"/>
+</p>
+<details>
+  <summary>See the source code</summary>
+  
+```typ
+#import "@local/mastermind:0.2.0": *
```
elegaanz: Unless specifically demonstrating how to import from another namespace, examples should use `@preview`, so that they can be easily copied by users. Can you update this import please?

elegaanz (commented):
There is one small thing to change in an example, but otherwise the update looks good to me.

elegaanz:
Thank you!

PR #5652, opened 2026-08-22
letterloom:3.0.1

nandac:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
### Fixed

- Fixed single-signature alignment so that when only one signature is given, it aligns across the full page width instead of using its previously measured column width. ([#6](https://github.com/nandac/letterloom/pull/6))

### Documentation

- Added a note on embedding annotated PDFs as enclosures: print the PDF to a new file before embedding to flatten annotation layers into the page content, since Typst's `image()` function only rasterizes the underlying content stream. ([#6](https://github.com/nandac/letterloom/pull/6))

  - Explanation: A highly customizable and user-friendly template designed to streamline the creation of professional correspondence.
   

saecki:
Thanks!

PR #5658, opened 2026-08-23
ntnu-report-ethical-hacking:1.0.0

Agryos:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a nice little template for ntnu student to give them a template for the iik3100 course. It will be easy for them to pick up typst for them after that !

  - Explanation:
    

saecki:
Thanks for this package! Unfortunately we don't want to publish templates that don't have a library part. One problem with such template packages, is that there is no real upgrade path when a new a new version of the package is published. Which somewhat defeats putting them into a package manager. Feel free to ask, if you have any questions :)

Agryos:
Well I don't plan to add a library part sooooo.... I guess it is too bad for me 🙃

Agryos:
I think it would be cool that template where just template... And that the library part is not mandatory, but if it is... I guess we can close it 😓

saecki:
> I think it would be cool that template where just template... And that the library part is not mandatory, but if it is... I guess we can close it 😓

Yes, the problem is that it would be less of a template and more of an example.
And if we allowed arbitrary examples to be published as packages that would lead to an influx of packages submissions, which would be very hard to judge.

PR #5659, opened 2026-08-23
typsium-ghs:0.1.2

Ants-Aare:
I am submitting
- [ ] a new package
- [x] an update for a package

better readme, better parsing of statements, added show rule. better standard output when using no parameters.


elegaanz:
Thank you!

PR #5660, opened 2026-08-23
patstdlib:0.3.3

patrick-kidger:
I am submitting
- [ ] a new package
- [x] an update for a package

Added:
- Support for setting up multiple fonts (in particular useful when working with CJK languages);
- Support for loading a dynamic number of show rules;
- Extended our referable `enum` support to allow for different styling between where it is displayed inline, and where it is `@referred`.


saecki:
Thanks!

PR #5663, opened 2026-08-24
min-writing:0.1.0

mayconfmelo:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Create quick notes intuitively and rapidly, using syntactic sugar that extends the markup supported by Typst. This package was designed to make creating documents as easy as possible, without requiring extensive initial
configuration—in fact, you simply need to import the package and apply the `#show` rule to access its
features—yet it still offers various options for fine-tuning.

- Quick-note mode (no page breaks)
- Extended syntax sugar
  - Unnumbered headings
  - Highlighted texts
  - Boxed texts
  - Inline quotes
  - Block quotes
  - Tables
  - Paragraph breaks
  - Page breaks
  - Mermaid diagrams
  - Dividers
  - Check lists
- Additional commands
  - `#boxed` for boxed texts
  - `#mermaid` for Mermaid diagrams
  - Custom `#figure` with support for source
- Catppuccin colors
- Embedded help page

_**PS:**_ The thumbnail differs from the template because, by default, the package creates a [single-page document](https://raw.githubusercontent.com/mayconfmelo/min-writing/35f80f13e506ddbd3e18e2d2b75950291c8650e8/docs/example.pdf) with an automatic height, and the generated template does not match standard paper proportions; consequently, the resulting thumbnail does not fit well within the Typst Universe site's thumbnail display.

  - Explanation: It is the shorthand for "minimal-writing," part of the "min-" family of templates I developed. The name "minimal" reflects the minimal effort required to configure and use this package.
    

packages/preview/min-writing/0.1.0/src/cmd.typ
```diff
@@ -0,0 +1,72 @@
+/// = Additional commands
+
+#let original = (figure: figure, align: align)
```
elegaanz: For info, you can use the `std` module instead of making aliases manually: `std.figure` will always refer to the "original" figure function (unless you create a variable called `std` too, of course). You don't have to change it for your package to be accepted, but I wanted to let you know in case you need it again in the future.

packages/preview/min-writing/0.1.0/src/cmd.typ
```diff
@@ -0,0 +1,72 @@
...
+:boxed:
+Inserts inline text boxes.
+**/
+#let boxed(
+  stroke: auto, /// <- stroke
+    /// Set border stroke. |
+  body /// <- content | string
+    /// Text to be boxed. |
+) = context {
+  import "@preview/catppuccin:1.1.0": get-flavor
```
elegaanz: This import seems unused. If that's right, can you remove please?

elegaanz (commented):
Two small things. The second comment in particular needs to be addressed, so that users don't download a package they don't need.

packages/preview/min-writing/0.1.0/src/cmd.typ
mayconfmelo: I completely forgot about `std`! Thanks for reminding me; it's much easier than manually keeping track of all the shadowed functions.

packages/preview/min-writing/0.1.0/src/cmd.typ
mayconfmelo: I've removed it. That import was a leftover from the previous implementation, which was intended to fetch the color theme from within the command. When I changed the approach, I forgot to remove it. Sorry!

mayconfmelo:
The issues pointed out were resolved in the last commit 🙂

elegaanz:
Perfect, thank you!

PR #5670, opened 2026-08-24
typsy:0.2.5

patrick-kidger:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:
- Added support for classmethods. This is particularly useful for custom class constructors, of course.
    I decided against Rust-style associated functions as Typst's scoping rules (i.e. lack of hoisting) means that the class constructor isn't yet available from the global scope.
- Added support for using classes as variadic patterns. This means that `Array(..SomeClass)` and `Dictionary(..SomeClass)` are now valid patterns inside `match` statements / in class type hints.
- Added the `Optional` pattern, allowing for e.g. `Arguments(Int, Str, foo: Optional(Int))` to indicate an optional member of a `Dictionary` or `Arguments` pattern.
- runtime `typecheck`ing now uses more concise error messages.

### DIFF
--- a/packages/preview/typsy/0.2.5/README.md
+++ b/packages/preview/typsy/0.2.5/README.md
@@ -1,0 +1,222 @@
+<h1 align='center'>typsy</h1>
+<h3 align='center'>Classes/structs, pattern matching, safe counters... and more!<br>Your one-stop library for programming tools not already in core Typst.</h3>
+
+---
+
+_Pronounced 'tipsy', because I think that's funny and it's a nice pun on 'Typst'._ 😄
+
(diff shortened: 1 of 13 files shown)


packages/preview/typsy/0.2.5/README.md
saecki: Ideally the `add_integers` functions should be kebab-case and this example doens't compile, because `Int` isn't found.

saecki (commented):
Thanks for the update!

packages/preview/typsy/0.2.5/README.md
patrick-kidger: done.

PR #5674, opened 2026-08-24
primaviz:0.10.0

phiat:
## Summary

primaviz 0.10.0 — a charting library with 50+ chart types, 7 themes and no dependencies, built entirely from native primitives.

## Changes since 0.9.1

**New:** `sunburst-chart` gains `total-angle` and `start-angle`, so sunbursts can sweep a partial arc and start anywhere on the dial instead of always being a full circle. Both default to the previous behaviour (`360deg` / `0deg`).

**Fixes:** labels in `sunburst-chart`, `pie-chart`, `heatmap` and `correlation-matrix` were being painted over by shapes drawn after them, because each chart emitted a shape and its label in the same pass. Those charts now draw all shapes first and all labels second. `pie-chart`'s donut percentages also sat at a fixed fraction of the radius that fell inside the hole at high `donut-ratio`; they now follow the ring band. Heatmap values are sized to fit their cell rather than overflowing into the neighbouring one.

No breaking changes.

## Checklist

- [x] Version directory added, nothing else touched
- [x] `typst.toml` version matches the directory name
- [x] Examples import `@preview/primaviz:0.10.0`, not a local path
- [x] All 23 examples compile clean with zero warnings
- [x] Verified against the declared compiler floor using the actual Typst 0.13.0 binary, not inference
- [x] File list identical to 0.9.1 — no stray dev files, no missing sources
- [x] LICENSE (MIT) included

## Repository

https://github.com/phiat/primaviz

## Release

https://github.com/phiat/primaviz/releases/tag/v0.10.0


saecki:
Thanks!

PR #5680, opened 2026-08-25
sapians-paper:1.0.0, sapians-report:1.0.0, sapians-slides:1.0.0 and sapians:1.0.0

wbendinelli:
Official packages of **SAPIANS** — a token-driven design system for presentations and documents:

- **`sapians`** (core library): a 16:9/16:10 slide engine with 10 standardized slide families, incremental reveal (`pause`/`uncover`/`only`), handout mode, pdfpc speaker-notes export, plus two-column scientific-article and technical-report layouts. MIT.
- **`sapians-slides`**, **`sapians-paper`**, **`sapians-report`** (templates): ready-to-use scaffolds on top of the core. Template content is MIT-0 so user documents carry no attribution obligation.

**This is an official submission by the SAPIANS organization** — please let us know the steps to authenticate entity affiliation, and we're happy to adjust the template package names if the naming policy requires it.

Source, visual-regression suite (tytanic), and a tidy-generated reference manual: https://github.com/wbendinelli/sapians-design · gallery with compiled PDFs: https://wbendinelli.github.io/sapians-design/

All four bundles pass `typst-package-check` locally with no diagnostics. Fonts (Inter/JetBrains Mono, SIL OFL) are documented, not bundled, and the font stacks declare system fallbacks so everything compiles out of the box.


packages/preview/sapians/1.0.0/src/layouts/paper.typ
saecki, line 1: We don't want to have template packages without library code on typst universe. Because that could easily lead to an influx of example packages with questionable benefit. Here I can see a point of having separate `paper`, `report`, and `slides` packages.
For that you could move the respective file from `layouts` into the corresponding packages, while still having shared code in the `sapians` package.

packages/preview/sapians/1.0.0/typst.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "sapians"
```
saecki, line 2: For officially associated packages we require email confirmation. Could you send an email from a company email address to `hello@typst.app` that states you officially endorse all four packages.

saecki (commented):
Thanks for these packages!

PR #5681, opened 2026-08-25
fretwork:0.3.0

snaggen:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: fretwork sets guitar tablature — rhythm stems and beams under the
staff, technique symbols, bend arrows, lyrics, and ASCII tab import. This
updates it from 0.2.0 to 0.3.0.

**Added.** A verse may now be given as timed syllables —
`lyric-at(measure, position, text)` — instead of a run spent over the notes,
which is what lets a singer's line be printed under another part's staff, where
the two share bars but not notes. Verse numbers on each lyric lane, so one
stanza's row can be told from another's under a repeated passage whose music is
written once. And optional extender rules for melisma, behind
`lyric-extender` on the theme.

**Changed.** The far end of a tie now prints no fret number and is carried by
the arc alone: set in parentheses it was the same picture as a ghost note, which
means the opposite thing. And a run of legato links is drawn as one slur over
the whole gesture rather than an arc per pair.

**Fixed.** Two marks of one kind — two `let ring` labels, say — no longer print
on top of each other where no single height holds both. A word the system break
cuts in two keeps its hyphen. A free instruction is boxed so it stays inside the
system. The system closing a passage is spaced at the density of the one above
it rather than being left at natural width. And a short bar closing a passage is
no longer reported as wrong, since it pairs with the pick-up that opened it.

Full changelog: https://github.com/snaggen/fretwork/blob/v0.3.0/CHANGELOG.md


elegaanz:
Thank you.

PR #5683, opened 2026-08-25
merman:0.2.0

Latias94:
## Summary

Updates `@preview/merman` from 0.1.0 to 0.2.0. The release provides a Typst 0.15.0 package surface backed by Merman 0.8.0-alpha.6 and Typst plugin ABI 2.

## What changed

- Adds the complete SVG and analysis surface used by the publish profile.
- Includes Mermaid's Cytoscape and ELK layout backends. ASCII, PNG, JPEG, PDF, and RaTeX are intentionally outside this package profile.
- Adds documented examples for direct rendering, raw blocks, document context, profiles, figures, options, SVG export, print, presentation, and ELK.
- Keeps `examples/**` in the submitted source for package review while excluding it from the runtime package download.
- Documents the 0.1.0 to 0.2.0 API and ABI migration in the package README.

The optimized WASM runtime is 10,166,376 bytes (about 9.70 MiB). The complete package directory is about 10.31 MB because it also carries the wrapper, documentation, and legal materials required for redistribution.

## Licensing

The wrapper and Merman-authored code use `MIT OR Apache-2.0`. The embedded ELK-derived component remains under `EPL-2.0` as a component-level boundary. `THIRD_PARTY_NOTICES.md` and the applicable files under `THIRD_PARTY_LICENSES/` are included so downstream redistributors can identify and preserve each obligation.

## Validation

- `cargo nextest run -p xtask --locked`: 592 passed.
- `cargo nextest run -p merman-typst-plugin --locked`: 13 passed.
- Typst package smoke: 20 positive fixtures passed and 1 expected compile failure was verified with Typst 0.15.0.
- `cargo fmt --all -- --check`, legal-material checks, and artifact dependency-closure checks passed.

The corresponding Merman source and legal-material changes are public on `Latias94/merman` `main` at `9cea78d1f`.


elegaanz:
Thank you for this update. I noticed that the WASM file is much larger now, is there a reason for that? Did you add a new dependency? I would say it's fine for now, but if it grows too much in the future we may reject package updates, as we don't want the registry to be too large and we want package to be as fast to download as possible. If you have not looked into it yet, optimizing the WASM bundle may be a good starting point.

PR #5685, opened 2026-08-25
conic-toan:0.3.6

Thutran1891:
Update `conic-toan` to 0.3.6.

Automatic plot windows, two new content blocks, and a fix for multi-column exam
layout on landscape pages.

### Added

- `#hinh` now infers its coordinate window. `xmin`/`xmax`/`ymin`/`ymax` all
  default to `auto`; when any of them is left out, the package lays the drawing
  out once inside a zero-sized probe box, intercepts every `place` call the
  drawing primitives emit, and measures the real extent on the page. Because
  the measurement happens at the `place` layer it also covers what a purely
  geometric bounding box cannot see: the *text width* of point labels, arrow
  heads, dot radii, and a circumscribed circle poking out past its polygon.
  Sides that are given explicitly are kept as written. Drawings whose strokes
  fill the whole window by construction (`truc`, `luoi`, `he-truc`,
  `gach-vung`) are detected and fall back to the previous default window.
- `#hdkp` — an opening "discovery" question type for introducing a new lesson.
  Unlike the other question types it does not require a solution; when one is
  given, the block is titled "Gợi ý" (hint) rather than "Hướng dẫn giải"
  (worked solution). The tag carries no number and its text can be replaced
  with `nhan:`. An inline builder `cau-hdkp` is included for use inside a
  manually built slide.
- `#phuong-phap` — a "solution method" block in the same family as
  `#dinh-nghia`, `#dinh-ly`, `#tinh-chat` and `#cong-thuc`, for stating the
  recipe for a class of exercises. It is numbered from its own counter, so its
  numbering runs independently of worked examples and practice problems;
  `so: false` drops the number and `#dat-lai-cau-pp()` restarts the count.

### Fixed

- `chia-2-cot` balanced its columns against portrait dimensions on pages set
  with `#set page(flipped: true)`. `flipped` does not change the `width` and
  `height` fields of `page` — the swap happens when the page is laid out — so
  reading `page.height` in a context still returned the portrait height. On A4
  landscape the library therefore believed a column was 259mm tall instead of
  172mm, placed the "last page" mark a whole page too early, and inserted a
  `colbreak()` in the middle of the document. The column being filled at that
  point was cut short, leaving a blank half-page in the middle of the exam and
  often pushing a final question onto a page of its own. Column heights are now
  taken from the region size reported by `layout`, which is always the real
  paper, and the remaining space on the current page is derived from the same
  figure.
- `ke-het-trang` had the same defect in three places and drew far too many
  ruled lines on landscape pages, overflowing onto the next page.
- Drawing helpers called without an explicit context now also work inside
  `measure`. The implicit context used to be carried in a `state`, which is
  inert for content built inside `measure`; it is now passed as a marker
  resolved by a show rule. This fixes figures silently disappearing from the
  continuation slides of a long solution.

Portrait documents are unaffected by the column fix: with `flipped` unset the
new code resolves to exactly the previous numbers, and page counts do not
change. Landscape documents that use `chia-2-cot` will re-flow, usually onto
fewer pages — that is the point of the fix.

Figures that already declare all four window bounds take exactly the same path
as before and cost nothing extra. A figure that declared none of them used to
fall back to a fixed `-5..5 / -4..4` window; it is now fitted to its content
instead. Probing costs a few extra layout passes per figure, so documents with
very many auto-fitted figures compile more slowly; declaring the window keeps
the previous speed.


elegaanz:
Thank you!

Thutran1891:
Thanks you very much.

PR #5689, opened 2026-08-26
tcdm:0.0.4

YDX-2147483647:
I am submitting
- [x] an update for a package

Description:
Previously, if no new items have been added since the last update, then building against a new update will trigger _error: dictionary does not contain key "new_addition"_. This is now fixed.


YDX-2147483647:
All lint warnings and errors reported by the bot are false positives as usual (v0.0.1 #3748, v0.0.2 #3846, v0.0.3 #5421).

elegaanz:
Thank you.

PR #5690, opened 2026-08-26
typsium-iso-7010:0.1.2

Ants-Aare:
I am submitting
- [ ] a new package
- [x] an update for a package
removed borders from svg
some signs had stray dots increasing the size
fixed colors
added iso 3864 colours


saecki:
Thanks!

PR #5692, opened 2026-08-26
structured-uib:0.2.0

AugustinWinther:
I am submitting
- [ ] a new package
- [x] an update for a package

Changed the faculty name to the new one, and added information about the usage of the emblem of UiB. Also removed the use of the codedis package in the example to keep the example as simple as possible. Also fixed link to STIX Two fonts.


saecki:
Thanks!

PR #5698, opened 2026-08-27
modernpro-coverletter:1.0.1

jxpeng98:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: modernpro-coverletter provides coordinated academic and professional cover-letter and statement templates.

Summary:
- Fixes configuration fallback resolution so explicit theme, layout, and legacy argument values override defaults.
- Removes the obsolete profile template asset from the release workflow.
- Keeps both starter templates self-contained.

Validation:
- Both template/coverletter.typ and template/statement.typ compile with Typst 0.15.1 through @preview/modernpro-coverletter:1.0.1.
- The source cover-letter and statement examples compile successfully.
- The repository bundler completes successfully.


saecki:
Thanks!

PR #5705, opened 2026-08-28
jxl-loader:0.3.0

frabera:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:
jxl-loader is a package that adds support to JPEG XL (https://jpeg.org/jpegxl/) images in Typst.

https://github.com/frabera/jxltypst

It provides an image wrapper function `image-jxl` to load the image file as path (for Typst >= v0.15.0) or as binary data, and it accepts all the arguments of the native image function.

It collects the image binary data and then decode it using jxl-rs (https://github.com/libjxl/jxl-rs), the new pure rust jxl implementation.

The wasm plugin returns the decoded data to typst as raw `rgb8`, `rgba8`, `luma8` or `lumaa8` (depending if it's grayscale and there is an alpha channel in the source image) to the native image function.

  - Explanation:
    It is quite quite descriptive and short.


frabera:
Do i need to remove typst from the name? I think it's quite related, otherwise it should be named something like `jxlloader`, let me know if you have any feedback.

elegaanz:
Having typst in the name is generally frowned upon because it is redundant for what is obviously a Typst package to be named like that. Here it is a bit less the case because it is not `jxl-typst` but some kind of mix of the two words, but I still think it is not great. I would say that `jxlloader` is better. In case you want other ideas to brainstorm, I thought of `jixel` (which is how I would pronounce JXL) or `virtual-jxl` ("virtual" because it is not a native implementation in the compiler), which would both be good distinctive and easy-to-remember names in my opinion. I'll let you change the name to whatever you prefer, `jxlloader` being a good option if that's what you like the most. If you think of another name, you can ask first if it fits our guidelines before actually making the change.

frabera:
@elegaanz Thank you, I modified the typst package name to jxl-loader, now all the checks are passing.

elegaanz:
Great, thank you :)

PR #5706, opened 2026-08-28
typograph:0.3.0

BenCichos:
I am submitting
- [ ] a new package
- [x] an update for a package

Adding the option to place nodes relative to other nodes, before this was only possible for drawing edges.


packages/preview/typograph/0.3.0/typst.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "typograph"
+version = "0.3.0"
+entrypoint = "src/lib.typ"
+authors = ["Benjamin Cichos <https://github.com/BenCichos>"]
+license = "MIT"
+description = "Draw unopinionated diagrams: nodes, edges, curves, ports."
+repository = "https://github.com/BenCichos/typograph"
+keywords = ["diagram", "graph", "node", "edge", "drawing"]
+categories = ["visualization"]
+compiler = "0.15.1"
```
saecki: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.15.0` will allow using this package with a slightly outdated compiler.

saecki (commented):
Thanks for the update!

packages/preview/typograph/0.3.0/typst.toml
BenCichos: Thank you. I updated this.

PR #5708, opened 2026-08-28
monofolio:0.1.0

harshkaso:
I am submitting
- [x] a new package
- [ ] an update for a package

  - Explanation:
    

packages/preview/resumetyp/0.1.0/typst.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "resumetyp"
```
saecki: While in theory this name is barely ok in regards to the [package naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) I would encourage you to find something a little more creative, because I just seems like you've appended `typ` to an unrelated word.
An alternative to a combination/word-play would be to use a non-descriptive prefix, that is possibly related to the style of this template.

packages/preview/resumetyp/0.1.0/example-resume.pdf
saecki, line 1: This file will be very hard to find on Typst Universe. Could you link to it from the readme or remove it?

saecki (commented):
Thanks for this package!

packages/preview/resumetyp/0.1.0/example-resume.pdf
harshkaso: I removed the file

packages/preview/resumetyp/0.1.0/typst.toml
harshkaso: Should I just change the name in this one place, or do I also need to change the directory name and the `#import` in the template file?

harshkaso: The name was chosen to follow the same convention as a similar LaTeX template I created, `ResumeTex`, which solves the same set of challenges. That said, I understand the concern, and I’m open to choosing a more creative name that better fits the Typst ecosystem and its naming conventions.

saecki: > Should I just change the name in this one place, or do I also need to change the directory name and the `#import` in the template file?

If you change the name, you also need to update the directory and all imports in the template and examples.

> The name was chosen to follow the same convention as a similar LaTeX template I created, ResumeTex, which solves the same set of challenges. That said, I understand the concern, and I’m open to choosing a more creative name that better fits the Typst ecosystem and its naming conventions.

That would be appreciated, though not strictly necessary :)

harshkaso: Hope this name, "Monofolio" is a better fit

saecki: Thanks, yes that's better :)

PR #5709, opened 2026-08-28
bme-vik-azslab-thesis:1.0.0

afranko:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: This is a template for thesis works at Budapest University of Technology and Economics, Faculty of EE and Informatics. It aims to provide students an easy way to choose Typst over WYSIWYG editors and improve the quality of sumbitted works.

  - Explanation: BME-VIK is the abbr. of the university and the faculty in hungarian. It contains thesis, but since university name is included it is not too generic, easy to find, not confusing at all.


packages/preview/bme-vik-thesis/1.0.0/typst.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "bme-vik-thesis"
```
elegaanz: The current name is not the most obvious or generic one for thesis in general, but just adding the University name doesn't make it fit our guidelines: we consider that this would be the most obvious name for a thesis template at this specific University.

Since this isn't an official package of the university, could you pick another name please? It should be distinctive enough so that another package with the same purpose can be published without one of the two having an unfair advantage, and it should be clear from the name that it is community-built and not officially endorsed by the university. See [our naming rules](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules).

elegaanz (commented):
Thank you for this package. I would just like you to pick another name, otherwise it looks good :)

packages/preview/bme-vik-thesis/1.0.0/typst.toml
afranko: Okay, I see.
I've added "azslab" to the title as the name of our lab where I'm a prof. So any other labs or even students can sumbit their own versions as well.

elegaanz: Thanks.

PR #5713, opened 2026-08-29
typstage:0.1.0

Loewe1000:
Animated HTML presentations from a single Typst file, and the slide set and the handout as PDF from that same file.

Every slide is typeset by Typst and embedded as SVG, so the arrangement in the browser is the same as in the PDF, down to the point. Only what asks to move moves: reveals, magic move, slide transitions, video, embedded documents. A small runtime does the moving; the HTML file is self-contained and needs no server.

- Manual: https://loewe1000.github.io/typstage/
- Seventeen example decks, running in the browser: https://loewe1000.github.io/typstage/beispiele/en.html
- Repository: https://github.com/Loewe1000/typstage

A note on the name: the guidelines ask that names not contain "typst". `typstage` is a portmanteau of *typst* and *stage* rather than the word plus a suffix, and there is precedent in the collection (`diatypst`, `genotypst`). Happy to rename if you would rather it did not.

Checked before submitting: the package imports and compiles on Typst 0.15 in both targets; every example in the README compiles and carries the current version; the licence in the manifest matches `LICENSE`; `examples/`, `docs/` and the build scripts are excluded, and the README's four images are excluded from the bundle but kept in the repository so the README can show them. Bundle after `exclude`: 23 files, 0.69 MiB.


elegaanz:
The name is fine, thank you for this package!

PR #5715, opened 2026-08-29
synkit:0.1.0

guilhermegarcia:
Adds `synkit` 0.1.0.

New `#spheres()` function for drawing Lewis-style systems of spheres.

Traces now use their own node label (`t1`, `t2`, ...) instead of a separate `trace1`, `trace2`, ... namespace, so every anchor in the package follows the same naming rule. `show-refs` displays that label, including for traces inside triangles, which previously had no reachable name.

The README has been rewritten so each example figure is accompanied by the code that produces it.


elegaanz:
Thank you!

PR #5720, opened 2026-08-30
nonodraw:1.0.1

dayala1:
… well as placement of guide numbers

I am submitting
- [ ] a new package
- [x] an update for a package

Description: Fixed the sizing of nonogram cells when nonogram is smaller than container, avoiding its squeeking. Also, fixed the placing of guide numbers in such case.


saecki:
Thanks!

PR #5723, opened 2026-08-31
touying-simpl-nwpu:0.1.0

RamanaVais:
I want to submit this package to the official package index so that it can be installed via "typst init" or "@preview". Thank you for reviewing!


elegaanz:
The `nwpu-vi-logo.png` image is broken: it is actually a text file that says "400: bad request". Can you fix it please?

PR #5724, opened 2026-08-31
conic-toan:0.3.7

Thutran1891:
Update `conic-toan` to 0.3.7.

### New internal helpers

- `_tach-tu-om(content)` flattens a content sequence and splits only plain
  text nodes at whitespace boundaries. Math, emphasized text, links, and all
  other content nodes remain atomic, so wrapping never cuts through their
  internal structure.
- `_cat-doan-om(content, narrow-width, figure-height, truoc: none)` finds the
  longest safe prefix that fits beside a figure. It includes any question
  blocks already placed before the paragraph and uses binary search, requiring
  only `O(log n)` measurements for a paragraph containing `n` atoms.

These helpers are private implementation details. Version 0.3.7 adds no new
public names and does not change any public function signature.

### Fixed

- Long question statements placed beside a figure now wrap below it even when
  the statement is a single paragraph. Previously the wrapping algorithm only
  split at paragraph, line-break, or block boundaries, so a long uninterrupted
  paragraph remained confined to the narrow text column and left unused space
  below the figure.
- Figure wrapping now also works when the statement consists of exactly one
  block, and when a long paragraph follows a separately laid-out question
  prefix. The fitting prefix stays beside the figure; the suffix and all later
  blocks resume at full width below it. If no safe split exists, the existing
  two-column fallback is preserved.
- Prebuilt variation tables now measure the horizontal space provided by their
  current layout region and scale down proportionally when necessary. This
  prevents their fixed-width interval columns from overflowing a `chia-2-cot`
  column and overlapping the questions in the adjacent column. Tables placed
  in a region that is already wide enough keep their previous size unchanged.

The public API is unchanged. Documents can still disable wrapping globally
with `#kieu-cau-hoi(om-hinh: false)` or per call with `om: false`. Multi-paragraph
questions and solutions keep the existing block-boundary behavior; only the
previously unsupported single-paragraph case reflows.

Variation tables require no new option. Their arrows, labels, rules, and text
are scaled together, so the table remains proportional inside narrow regions.

### Validation

- Compiled with Typst 0.13.7.
- `thu-om-hinh.typ`: 3 pages, no errors; the single-paragraph regression case
  wraps below the figure while the existing long-solution case still wraps.
- `thu-chia-cot-dai.typ`: 6 pages, no errors or convergence warnings.


saecki:
Thanks!

Thutran1891:
Thanks!

PR #5727, opened 2026-08-31
meshpad:0.1.1

13Stokes31:
Adds **meshpad** 0.1.1 — a patch release that validates the numeric arguments of
`grid-paper`. No change to what it draws.

Until now only `kind` and `lines` were checked. `step: 0` failed with Typst's
"cannot divide by zero" pointing inside the package, `major-every: 0` with
"divisor must not be zero", and a negative `step`, a negative or fractional
`major-every`, and a zero or negative `size` all compiled in silence and drew
nonsense. Each of those now raises a `meshpad:` message that says what was
expected and what it got, before anything is drawn.

Two details worth noting, since `size` and `step` accept both plain numbers and
Typst lengths. Positivity is tested against the zero of the value's own type, as
a length and an integer cannot be compared. And the type check is "all lengths
or all plain numbers" rather than "all the same type", so the documented
`size: (6, 4)` with `step: 0.5` — an int/float mix — keeps working; what is
rejected is mixing a length with a plain number, which used to fail deep inside
CeTZ with "cannot add length and float".

The must-fail cases are now an actual battery (`tests/errors.sh`, 10 cases each
asserting its own message); they previously sat commented out and never ran.

- Repository: https://github.com/13Stokes31/meshpad
- License: MIT
- Depends on CeTZ 0.5.2. Typst `>= 0.14.0`.

0.1.0 is untouched; this adds `packages/preview/meshpad/0.1.1/`.


elegaanz:
Thank you.

PR #5731, opened 2026-08-31
fretwork:0.4.0

snaggen:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: fretwork typesets publication-quality guitar tablature — rhythm stems
below the staff, technique symbols, lyrics, and import of plain ASCII tab.

0.4.0 adds the Drop C tuning and rewrites the user guide as Markdown chapters
instead of thirteen page images, so the syntax can be searched, linked to and
quoted rather than only looked at; only the staves are images now. The fixes are
all in the marks above the staff: the three kinds of articulation — length,
attack and stroke direction — now stack against the note instead of only the
first one written surviving; a staccato dot sits on the staff rather than
floating a third of a space above it; an ornate repeat's serifs no longer lift
every mark in the system; and a mark over a note on the top string no longer
prints on its fret number.


elegaanz:
Thank you!

PR #5734, opened 2026-09-01
taskize:0.2.10

nathan-ed:
I am submitting
- [x] an update for a package

Description: Fixes `columns: "auto-fit"` overflowing content that cannot reflow — a table with fixed columns, a fixed-size box, an image given an explicit width all keep the same height at any width, so the height-based fit test accepted every column count and laid them on top of each other. Such items are now measured by their natural width as well and get the columns they need, while content that does adapt (wrapping text, scaling images) is unaffected.


saecki:
Thanks!

PR #5735, opened 2026-09-01
exercise-bank:0.6.4

nathan-ed:
I am submitting
- [x] an update for a package

Description: 0.6.4 adds two-column layouts — `exo-page-columns` for a whole document, `exo-columns` for a single block, and `corr-columns` for the collected end-of-chapter corrections — each with an optional vertical rule and badges moved above the statements so they cost no column width. It also makes badge geometry configurable (`badge-scale`, `badge-pad-x/y`, `badge-radius`) and fixes several layout-convergence bugs, notably a label column that measured the badge it was about to draw.


elegaanz:
Thank you.

PR #5743, opened 2026-09-02
kinetic-kit:0.1.1

ll-nick:
I am submitting
- [ ] a new package
- [x] an update for a package

Just a small change to the package description for compliance reasons.
KSP (the KIT in-house publisher) endorses this dissertation template but has not relations to the Bachelor/Master thesis template that's provided as a convenience. This change makes sure that won't be misunderstood.


elegaanz:
Thank you for the clarification. :)

PR #5745, opened 2026-09-02
vlna:0.4.0

iamanro:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: `vlna` inserts Czech non-breaking spaces — short prepositions and conjunctions, titles before and after a name, numbers with units, dates, ratios, abbreviations, dashes and URL break points — following the ÚJČ line-breaking guideline and the luavlna command set. It wraps the matched content in a `#box` instead of rewriting the text, so jump-to-source keeps working.

0.4.0 is a performance release. The typography is unchanged; only the compile time moves.

- The three module-level states are merged into one dictionary, so a match costs one state read rather than three.
- A document that uses neither a mid-document toggle nor a `raw` block now needs no state read at all: `apply-vlna` walks the content before layout and installs a `context`-free glue in that case. A document-level query stays as a safety net for `raw` created during layout.
- Measured on a 501-page all-rules-firing Czech document, 25 interleaved runs pinned to 8 cores: 12.83 s → 8.66 s at p50 and 12.98 s → 8.79 s at p95, i.e. −32%.
- 24 configurations were rendered with 0.3.0 and with 0.4.0 and compared two ways: every word position from `pdftotext -bbox`, rounded to 1/1000 pt, is identical, and the page rasters differ only by antialiasing rounding. The harness, the chart and the raw dataset are in the repository.

No public API change.

### DIFF
--- a/packages/preview/vlna/0.4.0/CHANGELOG.md
+++ b/packages/preview/vlna/0.4.0/CHANGELOG.md
@@ -1,0 +1,169 @@
+# Changelog
+
+## [0.4.0] – 2026-09-02
+
+Performance release: identical typography, roughly a third off the compile
+time. Verified on 24 configurations (every rule group, per-language sets,
+links, `raw`, mid-document toggles, nested scopes, and columns from 12 mm to
+120 mm) — every word lands on exactly the same spot as in 0.3.0, debug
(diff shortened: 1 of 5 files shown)


elegaanz:
Thank you!

PR #5746, opened 2026-09-02
mastermind:0.2.1

manuel3053:
I am submitting
- [ ] a new package
- [x] an update for a package

feat: when overriding the side of a relationship, you are no more forced to specify both of them
fix: restore end tip of relationships
fix: the algorithm that draws relationships now draws correctly every combination of sides


packages/preview/mastermind/0.2.1/typst.toml
```diff
@@ -0,0 +1,12 @@
...
+name = "mastermind"
+version = "0.2.1"
+entrypoint = "src/lib.typ"
+authors = ["Manuel Zani <https://github.com/manu3053>"]
+license = "GPL-3.0-or-later"
+description = "Draw UML class diagrams."
+keywords = ["class diagrams", "uml diagrams"]
+categories = ["visualization", "scripting"]
+disciplines = ["engineering", "computer-science"]
+exclude = ["docs/*", "thumbnails/*", "*.pdf", "*.sh"]
```
saecki, line 11: Could you exclude the `examples` directory here? I think you may have renamed to `thumbnails` directory and forgot to update it here 🤔

saecki (commented):
Thanks for the update!

packages/preview/mastermind/0.2.1/typst.toml
manuel3053: the examples directory is saved for the readme

saecki: Please have a look at https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

The `exclude` list will affect which files are bundled and downloaded when importing a package. The files will still be available for display on Typst Universe.

PR #5749, opened 2026-09-03
conic-toan:0.3.8

Thutran1891:
Update `conic-toan` to 0.3.8.

### Added

- Add `duong-path`, a general open or closed drawing path that combines
  straight segments, circular or elliptic arcs, cubic Bézier segments, and
  sampled function or parametric graphs before applying one fill and stroke.
  It supports multiple subpaths via `bat-dau`, tangent-aligned arrowheads at
  the overall endpoints, and optional closing of every subpath. `duong-kin`
  remains the concise always-closed compatibility wrapper. Unlike
  overlapping independently filled shapes, semi-transparent fills do not
  darken at internal overlaps and no artificial seams are produced.
- Add the path segment constructors `noi-thang`, `noi-cung`,
  `noi-cung-elip`, `noi-bezier`, `noi-do-thi`, `noi-tham-so`, and `noi-cuc`.
  Ordinary points are shorthand for straight segments.
- `noi-do-thi(f, a, b, n: 80)` supports both increasing and decreasing
  bounds, allowing a graph boundary to be traversed in either direction.
  Functions used as path boundaries must be finite and continuous over the
  sampled interval.
- Add optional `mui-ten-dau` and `mui-ten-cuoi` parameters to `cung` and
  `cung-elip`. Each accepts an arrowhead size (for example `6pt`), follows the
  actual tangent at that endpoint (including reversed and rotated arcs), and
  defaults to `none`.
- Add `ve-tham-so` for parametric curves and `ve-cuc` for polar curves. Both
  use the existing drawing context, split samples outside the viewport, and
  support endpoint arrowheads. `lay-mau-tham-so` exposes the sampled points.
- Add composable affine drawing contexts: `ctx-affine`, `ctx-ti-le`,
  `ctx-nghieng`, and `ctx-doi-xung`. Reflection supports the coordinate axes,
  a point, or an arbitrary line through two points. Circles now follow the
  full affine transform instead of transforming only their centers.
- Add `giao-duong-cong`, which approximates all intersections of two
  parametric curves over supplied parameter intervals.
- Add lightweight free-drawing nodes with `nut-hinh` and `ve-nut`, plus
  `noi-nut` connectors that automatically meet rectangular, rounded,
  circular, or elliptic node boundaries. Compass anchors can be selected
  explicitly when automatic boundary intersection is not desired.

### Fixed

- Draw `truc-so` with a single arrowhead at the right end, matching the
  positive direction of the number line, instead of arrowheads at both ends.
- Move number-line labels slightly farther below their ticks so endpoint
  brackets no longer obscure the labels.
- Place coordinate labels for extrema intelligently in the built-in cubic,
  biquadratic, and quadratic-over-linear graphs. The x-coordinate label is
  placed below the axis for a positive y-value and above it for a negative
  y-value; the y-coordinate label is placed to the left for a positive
  x-value and to the right for a negative x-value. At extrema on the y-axis,
  the y-coordinate label uses `above-left` or `below-left` according to its
  sign. Explicit per-point label directions continue to override these
  defaults.
- Avoid duplicate y-coordinate labels for the two symmetric outer extrema of
  a biquadratic graph when automatic labeling is used.

### Implementation notes

- The implementation uses Typst's native `curve.move`, `curve.line`,
  `curve.cubic`, and `curve.close` primitives and requires no external
  package.
- Circular and elliptic arcs are converted to cubic Bézier pieces spanning
  at most 90 degrees by default. Rotated ellipses are supported, and callers
  can override the number of pieces with `chia:`.
- Arc arrowheads use the analytic endpoint derivative of the circle or
  ellipse, then apply rotation and the drawing-context transform. Their
  direction is therefore independent of the sampling count `n`.
- All endpoints and control points pass through the drawing context's
  coordinate transform. The resulting curve is normalized inside one
  positioned box, so `hinh` automatic viewport detection measures it
  correctly.
- When an arc or graph starts away from the current point, a straight joining
  segment is inserted automatically. A Bézier segment requires a preceding
  current point.
- Parametric intersections use sampled segment pairs with bounding-box
  rejection and geometric deduplication. Increasing `n` improves accuracy
  for strongly curved paths or nearby intersections.
- Node sizes are measured from their content in the active context;
  connectors calculate boundary intersections in page coordinates so text
  padding remains exact regardless of graph scale.

### Validation

- Compiled with Typst 0.13.7.
- Added `thu-duong-kin.typ` covering a capsule made from lines and arcs, a
  region bounded jointly by a parabola and a semicircle, a rotated elliptic
  arc combined with a cubic Bézier segment, a reversed function-graph
  boundary, automatic viewport detection, and elliptic arcs with integrated
  tangent-aligned arrowheads.
- Added `thu-giong-cuc-tri.typ` covering extrema on both sides of each axis,
  extrema lying on the y-axis with positive and negative ordinates, cubic,
  biquadratic, and quadratic-over-linear graphs, and explicit direction
  overrides.
- Added `thu-tikz-0.3.8.typ`, a visual integration test covering open and
  multi-subpath paths, endpoint arrowheads, parametric and polar plots,
  composed affine transforms, parametric intersections, and automatic or
  explicit node anchors/connectors.


elegaanz:
Thank you.

Thutran1891:
Thank you very much!

Thutran1891:
The PR was merged successfully, but all Deploy workflow runs after the merge were cancelled, and https://packages.typst.org/preview/conic-toan-0.3.8.tar.gz still returns 404. Could a maintainer please rerun the latest Deploy workflow?

elegaanz:
They were intentionally cancelled at the time, but were re-run since then, the new version is available.

PR #5752, opened 2026-09-03
measured-jair:0.1.0

veryviolet:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: Article template for the Journal of Artificial Intelligence Research (JAIR). JAIR moved to an `acmart`-based LaTeX class (`jair.cls`, acmlarge configuration) in 2025 and ships a LaTeX-only author kit; this template reproduces that layout in Typst so authors can draft and submit to JAIR (which accepts PDF) without porting a LaTeX class. Every dimension was measured on PDFs compiled with the real `jair.cls` (type sizes, page geometry, running head and foot, review-mode line numbers, title block, first-page notices, spacing around displays, floats and headings); the README documents the measurements and the known deviations.

  - Explanation: `measured-jair` follows the template naming rule (non-descriptive prefix + official venue abbreviation). "Measured" refers to how the template was built: by measuring the compiled reference documents rather than copying values from the class. It is a community port, not an official JAIR package, so the bare `jair` name is not claimed.

- [x] ensured that my package is licensed such that users can use and distribute the contents of its template directory without restriction, after modifying them through normal use. (`template/` is MIT-0; the manifest says `MIT AND MIT-0` and `LICENSE` carries both texts.)


packages/preview/measured-jair/0.1.0/typst.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "measured-jair"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Kirill Groshenkov <kir.groshenkov@gmail.com>"]
+license = "MIT AND MIT-0"
```
saecki, line 6: Out of curiosity, is there any reason you're dual licensing under `MIT` and `MIT-0`?

saecki (commented):
Thanks for this package!

PR #5756, opened 2026-09-04
presio:0.2.3

benedict-armstrong:
Bugfix release of [presio](https://github.com/benedict-armstrong/presio-typst-package).

- Fixed media placement under alignment: `media()` recorded `here().position()`, which is where the call sits in the flow, not where the block is finally laid out. An enclosing `align(center)` moved the rendered placeholder but not the recorded coordinates, so the viewer drew the video at the left margin under a centred placeholder. The sidecar coordinates now come from an out-of-flow anchor placed at the block's own top-left corner, which also fixes `align(right)`, vertical alignment, `place`, `move`, and nested containers. Reported by a user of the package ([#2](https://github.com/benedict-armstrong/presio-typst-package/issues/2)).
- Fixed reusing one local file across slides. The bytes are attached under `media-<id>.<ext>`, which carries no page number, so a second slide referencing the same file raised `attempted to attach file ... twice`. This also affected any local media inside a touying overlay group, since `#pause` repeats the call across subslide pages. The bytes are now attached once per document.
- YouTube `/live/` URLs (streams and premieres) are now recognised, and both the YouTube and Vimeo patterns are anchored at the scheme so a video link appearing inside another site's query string is no longer treated as the real host.

I confirm that I am the author of the previous versions (0.1.0, 0.2.0, 0.2.1, 0.2.2).

### DIFF
--- a/packages/preview/presio/0.2.3/LICENSE
+++ b/packages/preview/presio/0.2.3/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Benedict Armstrong
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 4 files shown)


saecki:
Thanks!

PR #5757, opened 2026-09-04
ctyp:0.3.2

HPDell:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

- Fix: default fontsets support `title` element
- Fix: package default options
- Feature: enable set another font family as a variant of font shape.


elegaanz:
Thank you.

PR #5764, opened 2026-09-05
ribbony:0.1.1

solstice23:
I am submitting
- [x] an update for a package

Description:

- Added "io-auto" option for snap parameter in label-drawer.default-linear-label-drawer as the new
default, which automatically chooses the best snap position based on the node’s incoming and outgoing edges.
- Added the #void node as a “hidden node” that can be used to override/customize the sizes of nodes.


saecki:
Thanks!

PR #5766, opened 2026-09-05
sanity:0.2.0

techgustavo:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Update for [`sanity`](https://github.com/techgustavo/sanity) to 0.2.0.                                                                 
                                                                                                                                                                                         
- `missing-label` now also flags numbered equations and footnotes with no label.
- `uncited-entry` no longer counts a citation written only inside a figure or table caption as citing that entry.
- `duplicate-caption` no longer false-positives on two captions that differ only by their citation.
- Added the "model" and "scripting" categories (`typst.toml`).


saecki:
Thanks!

PR #5768, opened 2026-09-05
codigo-matematico:0.1.0

ctafur80:
I am submitting
- [x] a new package
- [ ] an update for a package

Description:

Código Matemático provides mathematical notes with PDF and experimental HTML
output from the same source. It includes Spanish-labelled mathematical
environments, cross-references, globally numbered axioms, and native collapsible
proofs in HTML. PDF output shows proofs in full. It requires Typst 0.15.0 or later.

The starter is an original example about distance on the real line. The library
is MIT-licensed; the editable starter is MIT-0. HTML loads text and math fonts
from external font services with fallbacks, as documented in the README. No
fonts, personal notes, or third-party example texts are included in the package.

  - Explanation: “Código Matemático” is the distinctive name of these notes' layout, rather than a canonical category such as “math-notes” or “theorems”.
- [x] have chosen a license and added the license files
  - The submission includes no documentation PDFs. The thumbnail is automatically excluded from the downloadable bundle.
- [x] ensured that users can use and distribute the contents of the template directory without restriction after modifying them through normal use

Validation: initialized an isolated `@preview/codigo-matematico:0.1.0` project
with `typst init`, then compiled it to PDF and HTML with Typst 0.15.0 and 0.15.1. The
thumbnail is an unmodified rendering of page 3 of that starter at 250 PPI.
The existing five regression examples also compile to both targets.

The minimum compiler version was verified against 0.15.0 after the initial
check reported that the registry validator uses that version. Both the starter
and the five regression examples compile to PDF and HTML with 0.15.0. The
existing underscore-separated parameter names are retained for compatibility.


packages/preview/codigo-matematico/0.1.0/src/template.typ
```diff
@@ -0,0 +1,224 @@
...
+// ----------------------------------------------------------------------------
+#let templ(
+  // dark_theme: false,
+  sheet: "a4",
+  lang: "en",
+  title: none,
+  authors: (),
+  abstract: [],
+  parts: false,
+  outline_depth: 4,
```
saecki, line 27: You seem to be using snake_case for some parameters. The Typst ecosystem usually prefers kebab-case, which is also consistent with the standard library. If you strongly prefer snake_case you can leave it as is, otherwise it would if you could change these parameters to kebab-case.

saecki (commented):
Thanks for this package!

PR #5769, opened 2026-09-05
typstage:0.1.1

Loewe1000:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: typstage typesets every slide with Typst and writes it into the HTML as SVG, so what stands in the browser is what Typst set; a small runtime then animates it. The same source also gives a PDF slide set and a handout.

What changed in 0.1.1:

**Added**
- **Decks that read from the right.** `#set text(lang: "fa")` or `dir: rtl` turns the whole slide around — body, title band, callout bar, footer, progress bar, and the moving parts carry the direction into their own frames. Reported on the forum from a Persian deck; decks that read from the left come out byte for byte as before.
- **`build(at: …)`** — a drawing whose stages do not come one click after another. On a slide with three diagrams: ten sprites instead of 22, and the file 2.98 MB instead of 3.45 MB.
- **`contents()`** — a table of contents whose entries link to their section slides, with a link back. Contributed as a pull request and rebuilt so the entries take the deck's own palette. A deck with more than one structure level indents the deeper entries (`indent:` for a length of your own, `none` for flush), and `highlight: true` says where the talk stands: the running part and chapter keep the full ink, everything before and after steps back. Every entry also carries `when` — `"past"`, `"running"`, `"coming"` — so a highlight of your own needs no arithmetic against `info().levels`.
- **A step bar under every tile in the overview**, and a pinned clock that can be resized at its edges.
- **`pages: "step"`** — one PDF page per step instead of one per slide, so the paper turns the way the talk does. Every reveal takes part; a scene gets a page per stop rather than per tween frame, and a camera move gets none of its own. Asked for on the forum.

**Fixed**
- Seven reveal helpers got their steps wrong in the same way: `stagger`, `cue`, `alternatives`, `tiles` and `build` handed a computed `at` into a tracked element, and `scene` and `camera` computed their step advance from a read. Either cost the document its convergence once the revealed body carried something out of the flow. All seven now let `track` hand out the steps, or compute the advance from the cursor alone.
- A `cue()` group claimed the forward arrow before it was due and swallowed the steps in between.
- A click on a tile in the overview also paged on.
- A centred `anim` was centred in the browser and flush left in the PDF.
- The handout sheet fits its page, and there is room under a 4:3 slide.

The full list is in the [changelog](https://github.com/Loewe1000/typstage/blob/main/CHANGELOG.md). [navigator](https://typst.app/universe/package/navigator) builds on `info()` and `deck-outline()` and is waiting on this release.


elegaanz:
Thank you :)

PR #5776, opened 2026-09-06
roomy-math:0.1.0

5mKh04:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: roomy-math provides a `room` function that fixes line
overflow when math expressions are enlarged to display size (via
`display()` or naturally-spaced `$ ... $`) while written inline. It
measures the display-style content's natural height and wraps it in a
box with that height, so the surrounding line expands correctly instead
of overlapping the text above and below.

  - Explanation:
    The package is named "roomy-math" rather than something literal like
    "inline-display" or "display-fix". "Roomy" evokes the idea of giving
    an expression enough room/space to render without colliding with
    surrounding text, which is a creative rather than purely descriptive
    term for the problem it solves.


packages/preview/roomy-math/0.1.0/typst.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "roomy-math"
+version = "0.1.0"
+entrypoint = "lib.typ"
+authors = ["Koha <khongchiuchepbai@gmail.com>"]
+license = "MIT"
+description = "Fixes line overflow when math expressions are sized as display while written inline."
+repository = "https://github.com/5mKh04/roomy-math"
+keywords = ["math", "inline", "display", "equation", "typography"]
+compiler = "0.15.1"
```
saecki, line 10: The `compiler` field is a minimum version, so unless you explicitly require the latest patch release, changing it to `0.15.0` will allow using this package with a slightly outdated compiler.

saecki, line 10: Could you add an `exclude` list here with the `examples` directory? This will ensure the example images won't be downloaded when importing this package. The images will still be available when displaying the readme on Typst Universe. See: https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude

saecki (commented):
Thanks for this package!

PR #5792, opened 2026-09-08
erna:0.2.0

tad-lispy:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 

This version addresses the issue where passing a relative path to an image results in wrong absolute path resolution. It's discussed here:

https://forum.typst.app/t/erna-a-typst-library-to-scale-and-align-rows-of-images/8375/20?u=tad-lispy

The solution is to pass content, not paths to images. This opened another issue: what if user passes content that is not an image. After some tinkering, this is now possible and seems to work correctly. So now Erna can lay out any kind of content, or at least many different kinds, like images, shapes, text, etc.

What started as an annoying bug, resulted in a substantial expansion of useful features. I'm quite pleased with the result.


saecki:
Thanks!

PR #5795, opened 2026-09-08
moustaches:0.1.1

Akilon27:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: The new release brings proportionality tables (tables of nodes with each cell associated with a fletcher label for better positioning of edges and equal-sized columns and heights) and equation solving, both with the math-grid function, and two functions help-en and help-fr may be used to get help about a function or a parameter.


elegaanz:
Thank you!

PR #5816, opened 2026-09-10
abbrev:0.2.0

girasole123:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Abbrev v0.2.0 introduces several new features while maintaining backward compatibility with v0.1.7:

- Support for additional categories of abbreviations and terminology.
  - abbreviations and initialisms
  - glossary terms
  - symbols
  - acronyms
- The ability to create custom categories.
- Category-specific aliases for defining entries, referencing entries, and generating outlines.
- Support importing definitions from CSV and JSON files.


packages/preview/abbrev/0.2.0/README.md
```diff
@@ -0,0 +1,611 @@
...
+
+Then run:
+
+```bash
+typst compile example.typ
+```
+
+## Examples and compiled PDFs
+
+The repository contains example documents demonstrating both the backward-compatible API (file `example.typ`) and the new v0.2.0 functionality (file `example2.typ`).
```
saecki: These files will be quite hard to find on Typst Universe. Consider making these links:

```suggestion
The repository contains example documents demonstrating both the backward-compatible API (file [`example.typ`](example.typ)) and the new v0.2.0 functionality (file `example2.typ`).
```

saecki (commented):
Thanks for the update!

packages/preview/abbrev/0.2.0/README.md
girasole123: Hello,
Yes, this makes sense. But I improved your suggestion a little bit. I hope now it is much simpler to find those files.
Thank you!

saecki: The URLs you've added isn't valid, they return a 404. Since the local files are already present in this repository, you should link to them (in the way I suggested above), instead of linking to the files in you own repository. If you still want to link to the files in your own repository, you should either use a perma-link or a link to a specific branch/tag to ensure the files match the version of the released package. Additionally you should remove the example files from the submitted package, since they aren't used anymore.

girasole123: Sorry, I forgot to include the `blob/main` part of the path.

I didn't realize I could link to the files excluded in `typst.toml`, so I'll do as you suggested... it’s much simpler.

I didn't understand the last part of your message:
> Additionally you should remove the example files from the submitted package, since they aren't used anymore.

If I remove them, I won't be able to link to them as you suggested, will I?

Thank you!

saecki: Thanks, no worries :)

> I didn't understand the last part of your message:
> If I remove them, I won't be able to link to them as you suggested, will I?

Yes, you're right. That was only meant in the context of using a link to an external repository.

PR #5819, opened 2026-09-10
non-boring-notes:0.1.1

martincanga:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Adds the term 'unit' (and its corresponding translation) to the h1_heading so class notes can be easily organized by unit.


aruzdh:
Hi, @martincanga! Thank you for contributing—just a small suggestion. 

We usually map "Tema" to "Topic", whereas "Unidad" is used for "Unit". We could easily add both: Unit -> Unidad; Topic -> Tema (if you want to use "Tema").

Feel free to update the PR if that works for you, or let me know what you think :]

martincanga:
Hi, @aruzdh ! That makes complete sense.

I'll update the PR to follow that mapping (Unit -> Unidad and Topic -> Tema). Thanks for the feedback!

elegaanz:
@aruzdh Sorry for the repeated pings, the bot is a bit dumb here. Still, I would like you to double-check that the changes corresponds to what you asked for. Also, in case you didn't already do it, it would be best to backport this new version to your own repository, so that future contributors can keep building on it.

PR #5821, opened 2026-09-11
conic-toan:0.3.9

Thutran1891:
Update `conic-toan` to 0.3.9.

### Added

- Add a `tieng-anh: false` option to both `de-toan` and `bai-giang`. Setting
  it to `true` switches all library-generated default labels to English,
  including Definition, Theorem, Property, Formula, Example, Practice,
  Method, Note, Remark, Activity, Application, Question, Solution, Answer,
  question-type slide titles, test metadata, and answer-key headings.
- Switch true/false markers from `Đ`/`S` to `T`/`F` in English mode and set
  the document language to English for appropriate shaping and hyphenation.
- Preserve every explicitly supplied title or label (`tieu-de`, `prefix`,
  `ten`, `nhan`, and `nhan-giai`) without translating user-authored content.

### Compatibility

- The option defaults to `false`, so existing documents retain their
  Vietnamese labels and layout without source changes.
- The implementation is pure Typst and adds no external dependency.

### Fixed

- Make `bang-xet-dau` sign charts detect the available container width and
  scale down uniformly with reflow when needed, matching variation tables in
  `chia-2-cot` and other narrow layouts while preserving their original size
  in wider contexts.
- Make `giong` place axis labels automatically according to the projected
  point: the ordinate label switches across the y-axis based on the sign of
  x, while the abscissa label switches across the x-axis based on the sign of
  y. Explicit `huong-x` and `huong-y` values still override this behavior.
- Place vertex labels in the widest free angular gap between incident edges
  for the built-in box, cube, triangular-prism, regular hexagonal-prism, and
  regular truncated-pyramid figures. This keeps labels away from solid and
  hidden edges, including at interior-looking vertices in the projection.

### Validation

- Compiled `thu-tieng-anh.typ` with Typst Python 0.13.7 in both `beamer` and
  A4 `loigiai` profiles.
- Verified the English PDF text contains the complete English label set and
  no Vietnamese framework labels; also compiled the same test with
  `tieng-anh: false` to confirm the Vietnamese fallback.
- Compiled a focused sign-chart regression in both full-width and two-column
  layouts and verified that the narrow chart stays inside its column without
  overlapping adjacent content.
- Compiled a focused `giong` regression covering all four quadrants, points
  on either axis, and explicit direction overrides.
- Compiled the affected built-in spatial figures at the compact 3.2--3.4 cm
  widths used in the guide and visually verified that vertex labels no longer
  sit on incident edges.


elegaanz:
Thank you.

Thutran1891:
Thank you very much!

Thutran1891:
It would be great if the Typst app incorporated AI to write Typst code
based on user requests. Thanks.

PR #5825, opened 2026-09-11
polycv:0.2.0

bbinet:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: polycv 0.2.0 - a data-driven two-column CV package for Typst.

- [X] have chosen a license and added a `LICENSE` file


packages/preview/polycv/0.2.0/template/Makefile
elegaanz, line 1: It looks like I didn't really made the remark during the previous review, but I'm a bit uncomfortable shipping a Makefile as part of the template: for web app users and more generally users who are not familiar with Make, this file will be just be confusing. I still understand that it can be useful for people who know how to use it. I think it would make sense to move it out of the template directory, and link to it from your README: this way people who need it can easily download it and add it to their project, but by default you don't get a file that you don't know how to use or can't run.

PR #5827, opened 2026-09-12
typed-scores:0.5.1

GeronimoCastano:
I am submitting
- [ ] a new package
- [x] an update for a package

Changes:

- Corrected key signature accidental positions in every clef. Flats were engraved a third too high in treble, bass, and alto, and a step too high in tenor.
- Tenor clef sharps no longer reuse the treble layout; the first sharp now sits on the low F line so the signature stays inside the staff.
- Key cancellation naturals follow the same corrected positions, so flat keys are cancelled where the replaced accidentals stood.
- Added `cue { ... }` groups, which draw notes, chords, and rests at cue size while keeping their rhythmic value.
- Tuplets accept `number=never` to hide the numeral independently of the bracket.
- Tuplet numerals placed on the notehead side no longer touch the noteheads.
- Tuplets and tremolos written after an automatic rest `_` now attach to the correct notes.


elegaanz:
Thank you.

PR #5828, opened 2026-09-12
phonokit:0.5.16

guilhermegarcia:
Patch release of `phonokit`.

- `vowels()` and `consonants()` now validate inventories instead of silently dropping unsupported symbols. Unknown preset names and unsupported IPA symbols produce an error listing the available presets; empty or whitespace-only input renders nothing.
- Whitespace between symbols in custom inventories is normalised, so `vowels("a e i o u")` and `consonants("p t k")` behave as expected.
- No API changes.


saecki:
Thanks!

PR #5839, opened 2026-09-14
chengsi:0.1.2

zhaozigu:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: 
- **Lettered appendices.** A new `appendix` environment (`#let appendix = env.appendix`) numbers
  its chapters A, B, C (sections A.1, …) through the new `appendix-numbering` option, labels the
  chapter 附录 / APPENDIX, and leaves the numbering of any chapter that follows untouched.
- **One shared caption layout for images and tables.** Captions are now rebuilt by hand: a
  centered figure, an accent-coloured label, and small muted caption text below the figure,
  configured through `figure-caption-size` and `figure-caption-color`. A figure is never split
  from its caption across pages. References read 图 / Figure 1 and 表 / Table 1 in bilingual
  documents.
- **Template and docs.** The bundled template gained an appendix on figures and captions plus a
  sample plot in `template/assets/`; the README documents the new options and the new defaults.
- **Fix: inline code size.** Inline code now uses the body font size instead of a fixed `0.85em`,
  so code spans no longer look smaller than the surrounding text. The new `code-inline-size`
  option (default `1em`) adjusts it per project.


elegaanz:
Thank you for the update.

PR #5841, opened 2026-09-14
Remove tud-corporate-design-slides:0.1.0

jakoblistabarth:
Remove the template package for slides in the (old) corporate design of the Technische Universität Dresden, due to a request by the university: the design changed and a new template is in the making.

I am submitting
- [ ] a new package
- [x] an update for a package


elegaanz:
We generally consider the package index to be append-only, to guarantee that old documents will always keep compiling. If this version of the slides design is outdated and should no longer be used, a better solution would be to publish a minor version that adds a warning section in the README, explaining the current status. Once the new design is ready, you can publish it as a new major version of this same package.

Let me know if that doesn't fit your needs or if you need clarifications on this decision.

jakoblistabarth:
> We generally consider the package index to be append-only, to guarantee that old documents will always keep compiling. If this version of the slides design is outdated and should no longer be used, a better solution would be to publish a minor version that adds a warning section in the README, explaining the current status. Once the new design is ready, you can publish it as a new major version of this same package.
> 
> Let me know if that doesn't fit your needs or if you need clarifications on this decision.

Thank you for the explanation. I’ll discuss this with the responsible team at the university and will follow up to see if we can publish the new design as an updated version of this package.

PR #5842, opened 2026-09-14
qrypst:0.1.1

kljensen:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: qrypst draws QR codes from a small WebAssembly plugin (Nayuki's qrcodegen, 42 KB) and sizes the printed symbol from the exact module count the encoder reports, so a `module: 0.635mm` argument means 0.635 mm on paper. It encodes in about 7 ms per code, which matters for documents that print many distinct codes (we use it for 100-copy paper quizzes whose answer sheets each carry a code). Three functions: `qr` draws a symbol, `encode` returns the module count and an SVG, `matrix` returns the raw module matrix for callers who want to draw modules themselves.

  - Explanation: "qrypst" is a portmanteau of QR and Typst, not the canonical "qrcode" or "qr-code"; it names this particular plugin rather than the feature.

Provenance of the binary: the wasm is built from https://github.com/kljensen/qrypst at tag v0.1.1 by a release workflow that builds it twice from a clean tree and refuses to publish unless the two checksums agree; the release attaches `SHA256SUMS`. The package's `qrypst.wasm` matches that release asset. The embedded encoder is qrcodegen (MIT) and the plugin protocol is wasm-minimal-protocol (Unlicense); both are named in the README's licence section.


saecki:
Thanks for this package!

PR #5843, opened 2026-09-14
zhaji:0.1.0

Boshugege:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A clean, publication-grade Typst template tailored for mathematics, STEM, and lecture notes. Supports frictionless single-lesson note taking and seamless full-book compilation with visual-anchor heading hierarchy, dual-track compilation, and refined Chinese/Latin typography.

  - Explanation: "Zhaji" (札记, zhájì) is a traditional Chinese scholarly term referring to reading notes, study logs, and lecture commentaries compiled by scholars. It is distinctive, non-canonical, and culturally meaningful for lecture notes and study compendiums.
    

packages/preview/zhaji/0.1.0/README.md
```diff
@@ -0,0 +1,355 @@
...
+- **Dual-Track Compilation**: Compile single lesson notes independently during class with instantaneous preview, or compile the whole book into a unified volume with an auto-generated Table of Contents and clean page numbers.
+- **Visual Anchor Hierarchy**: Headings are distinguished through vertical color accents, subtle geometric glyphs, and calibrated whitespace rather than noisy digits.
+- **Restrained Callouts**: Only 3 intuitive block macros (`#thm`, `#def`, `#hint`) that eliminate verbose parameter boilerplate.
+- **Dynamic Headers & Footers**: Centered footer page numbers, cover pages with zero header/footer noise, and running headers that dynamically track the current section.
+
+---
+
+### Project Structure Recommendation
+
+It is strongly recommended to decouple the template from specific course repositories:
```
elegaanz: This part is confusing to me: Typst templates/packages are invisible to the user, they just import them using `#import "@preview…"` and don't have to worry about where they are stored.

packages/preview/zhaji/0.1.0/README.md
```diff
@@ -0,0 +1,355 @@
...
+my-zhajis/
+├── book.typ               # Full book entry point (Cover + Outline + Chapters)
+├── notes/                 # Lesson notes (01.typ, 02.typ, ...)
+├── figures/               # Vector plots, phase portraits, diagrams
+└── out/                   # Exported PDFs (git-ignored)
+```
+
+---
+
+### Installation via Typst Local Packages (`@local`)
```
elegaanz: Same for this part, it can be kept as an alternative solution to `@preview` imports, but it should not be presented as the default way of installing and importing the package.

elegaanz (commented):
I just have a remark about your README: it should present the `@preview` import as the default way of installing the package, as it works out of the box in all contexts (web app or local editor). Other options can be presented too if you feel like it is important to have them, but only as alternatives in case the standard way doesn't work (even if I can't really think of such a situation).

packages/preview/zhaji/0.1.0/README.md
Boshugege: Sorry, I just uploaded the same README.md from my personal repo, and it is meant to be used locally. I'll fix it immediately.

elegaanz:
Thank you :)

PR #5856, opened 2026-09-15
aero-dhbw:0.4.2

cocoveal:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Three changes in this release.

- Adds an `outline-depth` parameter to `aero-dhbw()` that sets the maximum heading depth shown in the table of contents (default `3`, matching the previous hardcoded behaviour). It is scoped to the table of contents only, so the figure/table lists are unaffected.
- Makes table styling consistent across the document. Table styling configured before `#show: aero-dhbw.with(...)` now applies to the main body, the annex, and the AI-usage table, so users define it once instead of repeating it. The glossary is locked to the template's own fixed styling so user rules cannot leak into it.
- Fixes the starter template's code-block setup so it actually uses codly as documented, by also importing `codly-languages` and passing `languages: codly-languages` to `#codly()`.

All are documented in the README.


saecki:
Thanks!

PR #5857, opened 2026-09-15
symbolica:0.1.0

benruijl:
I am submitting

- [x] a new package
- [ ] an update for a package

## Description

Symbolica is the official Symbolica plugin for Typst. It provides exact
algebra, differentiation, symbolic integration with steps, numerical evaluation,
equation solving, and matrix operations in one WebAssembly engine.

Authors can compute symbolic results and numerical values directly in their
papers, avoiding precomputation in another tool and error-prone copy-pasting.
Results update when equations or parameters change.

All use of Symbolica through this plugin within Typst is free, including
commercial and hosted use. No subscription, registration, activation, or
license key is required.

## Review considerations

- **Licensing:** The manifest declares `MIT` for the original Typst interface and Rust adapter code. The bundled `symbolica/symbolica.wasm` is built from components under multiple licenses and is not covered solely by MIT. Its dependencies are open-source except for the [symbolica](https://crates.io/crates/symbolica) crate, which retains its source-available license and the accompanying Symbolica Typst permission for free use and redistribution within Typst. The README explicitly documents this scope, and all applicable license files and third-party notices remain included.
- **Size:** Symbolica is a large code base that can perform many non-trivial calculations. As a result, the uncompressed wasm is 23.36MiB and the compressed package is 6.19Mib. I have put in significant work to reduce the size, including metadata compression of integration rules and flags that prevent monomorphization. It will be hard to make significant reductions without cutting features. If symbolic integration is removed, the uncompressed Wasm build is under 10MiB, but I reckon there will be quite some researchers and students that want to do symbolic integration from within Typst.
- **Usefulness:** I believe that a plugin like `symbolica` will find tremendous use among students and researchers who want to write papers in a less error-prone way. In fact, researchers from Bern University are writing open source extensions that interact with the primitives defined in this plugin to allow for tensor calculations inside Typst.

## Documentation

The development repository can be found [here](https://github.com/symbolica-dev/symbolica-typst-plugin). It contains a 50-page user manual, examples, complete API reference, and dependency license notices.

## Submission checklist

- [x] Selected [a name](https://github.com/typst/packages/blob/main/docs/manifest.md#naming-rules) that isn't the most obvious or canonical name for what the package does.
  - Explanation: Symbolica is the existing name of the computer algebra system,
    and this is its official Typst integration, rather than a generic name such
    as “algebra” or “integrate”.
- [x] Added a [`typst.toml`](https://github.com/typst/packages/blob/main/docs/manifest.md#package-metadata) file with all required keys.
- [x] Added a [`README.md`](https://github.com/typst/packages/blob/main/docs/documentation.md) with documentation.
- [x] Chosen [a license](https://github.com/typst/packages/blob/main/docs/licensing.md) meeting the submission requirements and added the license files.
  - License files are included, including a custom license that needs a green light from the Typst Universe maintainers.
- [x] Tested the package locally and it worked.
- [x] [`Excluded`](https://github.com/typst/packages/blob/main/docs/tips.md#what-to-commit-what-to-exclude) documentation PDFs from the runtime bundle, while retaining license files.


elegaanz:
I'm afraid the plugin is too large to be accepted. We already have a few files that are larger (or around that size) in the package index, but it was a mistake to accept them at the time, and it was decided to be more careful about file size. If removing symbolic integration makes the WASM file more reasonably sized, I think it is a good middle ground for now. We can reconsider this choice if a lot of people end up using your package and requesting support symbolic integration. Does that sound good?

benruijl:
Thank you for the quick reply! There is another size reduction option that I want to run by you: it is possible to compress `symbolica.wasm` to 6MB and include a very small WASM unzipper. The library is first decompressed in memory, then loaded through Typst's `plugin` constructor. This brings the size of the whole Typst Universe package down to 8.88MB. Is this something you would consider? 

@elegaanz I have updated the PR so you can see what it looks like.

elegaanz:
Looks like a good solution, thank you :)

PR #5861, opened 2026-09-15
navigator:0.1.8

eusebe:
I am submitting
- [ ] a new package
- [x] an update for a package

Description: Add a typstage example (https://typst.app/universe/search/?q=typstage)


packages/preview/navigator/0.1.8/examples/typstage_miniframes.html
elegaanz: The example HTML files are not used anywhere, and will be very hard to discover. You should either link to them from your README or remove them. If they also exist in your own repository, you can also take the same approach as for PDF files and link there directly instead.

elegaanz:
Thank you.

PR #5863, opened 2026-09-16
qooklet:0.7.1

ivaquero:
I am submitting
- [x] an update for a package

- fix bugs in font-selection


elegaanz:
Thank you.

PR #5871, opened 2026-09-16
phonokit:0.5.17

guilhermegarcia:
Patch release of `phonokit`.

- Portuguese consonant inventory in `consonants()` revised: the strong rhotic is now the trill /r/ (contrasting with the tap /ɾ/) instead of /x/, and affricates were removed (phonetic [tʃ dʒ] are allophones of /t d/).
- `consonants(..., affricates: true)` no longer adds an empty affricate row for languages without affricates (e.g., Portuguese, French).
- No API changes.


elegaanz:
Thank you! :)

PR #5872, opened 2026-09-16
timble:1.4.0

Mambouna:
I am submitting
- [x] an update for a package

Description: Adds a parameter to hide the specific times of individual events.


packages/preview/timble/1.4.0/example/example.png
saecki, line 1: Could you remove this file? It seems to be a compilation artifact.
In the readme you could link the `template.png` file, since it seem to be exactly the same.

saecki (commented):
Thanks for the update!

PR #5878, opened 2026-09-17
classic-aau-report:0.3.2

Tinggaard:
I am submitting
- [ ] a new package
- [x] an update for a package

I've updated the dependency `hydra` to `v0.6.3` to prevent a compiler panic introduced in Typst 0.15.0.


saecki:
Thanks!

PR #5886, opened 2026-09-18
approximate-acmsmall:0.2.1

gasche:
This is a bugfix release of my template `approximate-acmsmall` which provides a Typst version of the `acmart` LaTeX class used by some ACM publications, with `acmsmall` options.

The rendering of first-level list bullets was previously as intended, but changed in recent Typst versions to display the bullet well below the box. The change in behavior may come from https://github.com/typst/typst/pull/8150 which was merged in Typst 0.15.1 (I have not tried to go back to be sure). In any case, the new definition should work equally with old and new Typst versions.


elegaanz:
Can you please fix the version numbers, as reported in CI?

gasche:
I updated the package to fix the version field in the manifest, thanks!

gasche:
A warning remains from `typst-package-check`: my template file is named `template/example.typ` (as the template is just a simple example of using the template), and it wonders if it should be excluded in typst.toml. It should not, but I could silence the warning by renaming the file. For now I propose to leave it as-is.

The warning seems to be a regression in the checker introduced by https://github.com/typst/package-check/commit/eff71383d2093b840c9dce05fa3d5d8d14673cda last May, as the code was previously careful not to suggest to exclude template files. I reported this upstream as https://github.com/typst/package-check/issues/88

elegaanz:
Yes, it is a false positive, sorry about that. Thank you for opening an issue.

gasche:
Thanks for taking care of the repository!

PR #5887, opened 2026-09-18
latex-doc:0.0.1

maobin-xu:
I am submitting
- [x] a new package
- [ ] an update for a package

Description: A LaTeX-style book template for Typst. Provides part/chapter/section structure with proper numbering, frontmatter/mainmatter/appendix/backmatter book parts, nine theorem-like environments (theorem, definition, lemma, corollary, proposition, assumption, exercise, example, notice), subfigures, citations, and bibliography. Useful for anyone writing longer documents like books or theses who wants LaTeX-style structure without leaving Typst.

  - Explanation:
    - The name "latex-doc" reflects that it emulates LaTeX's book-class document structure within Typst, without being as generic as "book" or "template".
- [x] added a typst.toml manifest file with all required keys
- [x] added a README.md with documentation for my package
- [x] have chosen a license and added a LICENSE file (MIT-0)


packages/preview/latexy-book/0.0.1/typst.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "latex-doc"
```
elegaanz: The name feels like it is a port of the LaTeX look made by the LaTeX project itself: it sounds too canonical. [Our guidelines](https://github.com/typst/packages/blob/0488b38bb7fdfbce81981803ebbcf5612f8919f8/docs/manifest.md#naming-rules) forbid such names, as we don't want to give an unfair advantage to one package over similar ones because of its name. The `-doc` suffix also doesn't bring much to the name in my opinion, as it is a very generic term and doesn't really reflect the fact that this is a book template.

Can you please pick another name? For example `booktex` or `latexy-book` would work better (but feel free to choose anything else). If you are unsure about a potential new name, please ask me.

maobin-xu: Thank you for reviewing! I followed your advice to change the package name from `latex-doc` to `latexy-book`.

PR #5890, opened 2026-09-19
sang-math:1.0.6

sangnhc87:
Release sang-math version 1.0.6 with pure Typst QR code generation for OMR grading.


elegaanz:
Thank you.

